PR #3, opened 2018-10-06
added requirements.txt

girisagar46:


torrmal:
Giri thank you, if you have time, it will be very helpful if also the setup.py has the same list. As such, it will be fantastic if setup.py generates  it from requirements.txt, in such a way we only maintain one list.

PR #48, opened 2018-12-29
Fix install pip version msg

ZoranPandovski:


torrmal:
Thank you @zoranpandovski

PR #77, opened 2019-02-06
New stats generator

George3d6:
Added a bunch of "score" based on which we can rate every column.

Added a quality score for each column, based on all the other scores.

Added a logging function that displays some of the more worrisome or interesting scores to the user and (will) trigger the backend to plot some of the distribution based on which they are created (e.g. data type distribution, outlier distribution)

Streamlined and cleaned the code a bit.

Notes:

Scores should be easy to add, remove or take out of the calculation of the overall quality score. The current scores are by no way definitive.

Currently the scores are `0` if the quality is perfect and `1` if it's horrible, we may want to invert this, if so, let me know.


mindsdb/__init__.py
```diff
@@ -2,11 +2,24 @@
...
+from mindsdb.libs.controllers.mindsdb_controller import MindsDBController
 
 name = "mindsdb"
 MindsDB = MindsDBController
-MDS = DataSource # A Mindsdb Data Source
\ No newline at end of file
+MDS = DataSource # A Mindsdb Data Source
+
+
+class Anon:
```
torrmal: Please remove this. people can access all CONFIG variables 
as 
```
from mindsdb import CONFIG
```
and all constants as 
```
from mindsdb import CONST
```

torrmal:
@George3d6 I am getting this error

```
lof.py:236: FutureWarning: default contamination parameter 0.1 will change in version 0.22 to "auto". This will change the predict method behavior.
  FutureWarning)
```

torrmal:
Getting logs without context

```
'----------'
INFO:mindsdb-logger-6176fce0-2a80-11e9-8f3f-88e9fe8586dc:libs/data_types/mindsdb_logger.py:132 - ' Numeric: 5037'
INFO:mindsdb-logger-6176fce0-2a80-11e9-8f3f-88e9fe8586dc:libs/data_types/mindsdb_logger.py:136 - '----------'
```

Not sure what is related to.

torrmal:
It says
```
Column "number_of_rooms" has 99.92% of it\'s values duplicated !'
```
However, the distribution of the number_of_rooms is as follows

image

One would suspect that the duplication is 25%

George3d6:
> @George3d6 I am getting this warning
> 
> ```
> lof.py:236: FutureWarning: default contamination parameter 0.1 will change in version 0.22 to "auto". This will change the predict method behavior.
>   FutureWarning)
> ```

Fixed in: 66d3a5232501b1d420993f49969fcc8c75476f6a

Just changed the parameter to auto, since that's the new recommended default and quite frankly, at this point, we haven't played around with this enough to optimize any hyperparameters, it's more of a POC to see if people are interested in the data.

George3d6:
> Getting logs without context
> 
> ```
> '----------'
> INFO:mindsdb-logger-6176fce0-2a80-11e9-8f3f-88e9fe8586dc:libs/data_types/mindsdb_logger.py:132 - ' Numeric: 5037'
> INFO:mindsdb-logger-6176fce0-2a80-11e9-8f3f-88e9fe8586dc:libs/data_types/mindsdb_logger.py:136 - '----------'
> ```
> 
> Not sure what is related to.

This is from `infoChart`, I thought it would be helpful to plot the data distribution server side.

Since for now infoChart only work client-side, I left it in the stdout as a POC for the kind of things I'd want to plot in the stats generator phase.

In the future we can remove it and make it send only.

I added `.info` line above that places a heading for the graph

Though we should consider creating a title for the plots from either the `uid` or an extra `title` parameter.

For now hopefully the fix in: b8da9e2f5fc35b634ffb77bbec0200ab783b7585 is enough

George3d6:
> It says
> 
> ```
> Column "number_of_rooms" has 99.92% of it\'s values duplicated !'
> ```
> 
> However, the distribution of the number_of_rooms is as follows
> 
> image
> 
> One would suspect that the duplication is 25%

Ok, I think we have different definitions for data duplication, I'll take this one to slack :)

mindsdb/libs/phases/stats_generator/stats_generator.py
```diff
@@ -223,32 +181,202 @@ def getParamsAsDictionary(self, params):
...
+        }
+
+        if stats[col_name][KEYS.DATA_TYPE] != DATA_TYPES.CATEGORICAL and stats[col_name][KEYS.DATA_TYPE] != DATA_TYPES.DATE:
+            data['duplicate_score'] = data['duplicates_percentage']/100
+        else:
+            data['duplicate_score'] = 0
+
+        return data
+
+    def empty_cells_score(self, stats, columns, col_name):
```
torrmal: since we are moving it all to underscores, in order to differentiate methods from variables, methods should be verbs, so calculate_empty_cells_score

mindsdb/libs/phases/stats_generator/stats_generator.py
```diff
@@ -223,32 +181,202 @@ def getParamsAsDictionary(self, params):
...
+        return {
+            'correlation_score': prediction_score * highest_correlated_column
+            ,'highest_correlation': max(corr_scores)
+            ,'most_correlated_column': other_feature_names[corr_scores.index(max(corr_scores))]
+        }
+
+    def data_quality_score(self, stats, columns, col_name):
+        scores_used = 0
+        scores_total = 0
+        scores = ['correlation_score', 'cummulative_lof_score', 'cummulative_z_score', 'data_type_distribution_score', 'empty_cells_score', 'duplicate_score', 'similarity_score', 'value_distribution_score']
```
torrmal: question: are all these scores uniformly normalized/scaled (example: all possible values of all scores are floats between 0 to 1)? are all these scores such that the scales are mapped to 1 means something like very bad, 0 means something like very good (example is correlation_score = 1 a bad thing just as similarity_score=1 also a bad thing)?

mindsdb/libs/phases/stats_generator/stats_generator.py
```diff
@@ -151,38 +83,64 @@ def getTextType(self, data):
...
 
+        return curr_data_type, type_dist
 
 
 
-    def getWordsDictionary(self, data, full_text = False):
+
+
+
+    def get_words_dictionary(self, data, full_text = False):
```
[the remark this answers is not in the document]
George3d6: will do

mindsdb/libs/phases/stats_generator/stats_generator.py
George3d6: again, will do

mindsdb/libs/phases/stats_generator/stats_generator.py
George3d6: All scores should be normalized between 0 and 1, where 1 is worst and 0 is best, I'll make sure once again that it applies to all of them.

mindsdb/libs/phases/stats_generator/stats_generator.py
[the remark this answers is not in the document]
George3d6: Fixed in #9880798afad27509d19568c36cdd9a5732c3aeb8

mindsdb/libs/phases/stats_generator/stats_generator.py
George3d6: Fixed in #9880798afad27509d19568c36cdd9a5732c3aeb8

mindsdb/libs/phases/stats_generator/stats_generator.py
George3d6: Confirmed that all the scores follow the pattern where 0 is good and 1 is bad, and added comments to all the scoring functions in #0cd13921838fef53140677f9fd47618154508703

PR #87, opened 2019-02-18
null Data Processing

duanzhihua:
mindsdb/libs/helpers/norm_denorm_helpers.py，mindsdb/libs/phases/stats_generator/stats_generator.py :Processing of adding null value


George3d6:
Fixed conflicts and merge, your help with spotting & fixing this is much appreciated :)

PR #91, opened 2019-02-21
Dev gh

George3d6:
The probabilistic validator is meant to give the user a likelihood that the prediction is correct (based on the missing features and prediction bucket). It's a bayesian model (ComplementNB atm) we fit on the validation set during `train`.

There's some merging issues at the moment AND the model is not getting properly saved to disk.

But, since I'm close to finishing this (hopefully) I'm opening the PR for discussion and to have a place where we can easily diff the changes.


George3d6:
One thing I should mention here, currently, to determine a features existence, I use the [-1] (last value) from the array representing said features.

As far as  I understood this is ok for numerical value (since it should be 0 if it doesn't exist and 1 if it does).

But is this also true for 1-hot encoded text values ? Or should i instead look for an array of all-zeros to determine that said column is null ?

docs/examples/nlp/train.py
```diff
@@ -1,11 +1,9 @@
 from mindsdb import *
 
 # We tell mindsDB what we want to learn and from what data
-MindsDB().read_csv(filepath="shakes.train.numLines.csv" ,delimiter=',' ,encoding='GB18030')\
```
torrmal: It should be Predictor

mindsdb/libs/controllers/transaction.py
```diff
@@ -135,6 +135,7 @@ def _execute_learn(self):
             self._call_phase_module('ModelTrainer')
 
             self._call_phase_module('ModelAnalyzer')
+            self.persistent_model_metadata.update()
```
torrmal: This should be inside the ModelAnalyzer

mindsdb/libs/phases/model_predictor/model_predictor.py
```diff
@@ -48,14 +75,16 @@ def run(self):
                         continue
 
                     actual_row = j + offset
-                    confidence = self.getConfidence(cell, confusion_matrix)
+                    # @TODO: Do we scrap this ?
+                    #confidence = self.getConfidence(cell, confusion_matrix)
```
torrmal: yes we can now scrap this as this Bayess implementation you have, is the right way to estimate this.

mindsdb/libs/phases/model_analyzer/model_analyzer.py
```diff
@@ -3,33 +3,62 @@
 from mindsdb.libs.phases.base_module import BaseModule
 from mindsdb.libs.data_types.sampler import Sampler
 from mindsdb.libs.ml_models.pytorch.libs import base_model;
```
torrmal: you can remove this

torrmal:
```
from mindsdb.libs.ml_models.pytorch.libs import base_model;
```
I mean this (remove)

mindsdb/libs/phases/model_analyzer/model_analyzer.py
```diff
@@ -3,33 +3,62 @@
...
+from mindsdb.libs.helpers.norm_denorm_helpers import denorm
 
+import pandas as pd
 
+#@TODO: Define a way to save self.transaction.persistent_model_metadata.probabilistic_validator
 class ModelAnalyzer(BaseModule):
 
     phase_name = PHASE_MODEL_ANALYZER
 
     def run(self):
```
torrmal: please add some comments for the code explaining whats been done here

mindsdb/libs/controllers/transaction.py
George3d6: Why are the other .update() calls not inside the phases ?
Either way, will move it

mindsdb/libs/phases/model_analyzer/model_analyzer.py
George3d6: Explaination should be found within the newly added comments in the probabilistic validator class and model analyzer phase.

Essentially TinyDB doesn't behave well when a dictionary contains binary, and pickle encoding is weird, so we just pickle each probabilistic validator, then decode it with `latin1` encoding (so that it becomes a python `str` object) and then TinyDB can save that do a file.

Later, when we need to use the model, we can simply unpickle the saved string after encoding it using `latin1` (sinc pickle only works on binary data)

Take a look at the 

`pickle` and `unpickle` methods in ProbabilisticValidator, it's not complex at all, just a bit messy and unintuitive (but it's the easiest way I found to serialize an arbitrary object into TinyDB).

torrmal: i understand, if it works its fine :)

PR #98, opened 2019-02-27
V1 prob dist

torrmal:
updates to the way we calculate the probability distribution of the result being correct


mindsdb/libs/ml_models/probabilistic_validator.py
```diff
@@ -52,22 +54,31 @@ def _closest(arr, value):
         """
         :return: The index of the member of `arr` which is closest to `value`
         """
-        aux = []
-        for ele in arr:
-            aux.append(abs(value-ele))
-        return aux.index(min(aux))
+
+        for i,ele in enumerate(arr):
+            if ele > value:
```
George3d6: Hmh, this implementation will always return the "larger" bucket.

So if we have value 5.1 and histogram 1-3-5-7-9 , it wil return 7, not 5. Is that intended ?

PR #104, opened 2019-03-16
Various improvements to timeseries and others

George3d6:
1. Timeseries should now work properly with the API (using an array of all the relevant past values per row instead of just passing one value at a time).

2. Changed the interface, added backend parameter to select model backend and the `learn` method now account for two types of windows:

* one on seconds (i.e. rows over x seconds spanned by the ts column, only works with date type timeseries col)
* one on samples (i.e. how many previous rows to take into account for each of the new timeseries rows)

3. Small changes to the data generator to produce more natural ascii strings

4. Other small changes (e.g. CSV sniffing now only take into account the header, since it seems to be recommended to do it that way if the csv contains text based on random SO posts I read and trying to fix issues with sniffing a "perfectly normal" randomly generate file)

Note:

Still not handling splitting of data into train/test/validation based on group_by properly (discussion around that in slack atm)

This approach is inefficient, a more efficient approach would probably be to treat the whole timeseries as one huge row (or multiple rows based on the sampling window, if it's quite small) and expand our to-predict label/value into len(row) columns. (This is in terms of memory efficiency and also, presumably, efficiency for ludwig)


George3d6:
Also added a poor fix to make the probabilistic validator work with text types.

I think it could be improved, but in order to do that we need to have some sort of similarity-based histograms for text.

PR #122, opened 2019-03-26
Support timeseries cell

### LINKED ISSUE #108: Support cells/columns with sequential values
If we get a cell such as `1 2 3 5 6` or `[1,2,3,5,6]` or `1,2,3,5,6` we should interpret that as a sequential value, rather than `TEXT` (which is what would happen now).

George3d6:
Fixed a few issues relating to hardcoded separator and a missing gitignore file from the previous PR.

This PR resolves #108

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -75,3 +75,4 @@ docs/Makefile
 docs/source
 integration_testing/video.py
 integration_testing/canvas.py
+test.out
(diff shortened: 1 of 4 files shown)


George3d6:
small changes and essential for the new PR to work properly, so I'll just go ahead and merge w/o review.

PR #132, opened 2019-03-30
Fix tensorflow requirements

ZoranPandovski:
This PR updates tensorflow requirements to use latest stable 1.13.1 version.


George3d6:
change should also apply to the `-win` requirements file, will do that in a sec.

PR #133, opened 2019-03-31
Interface revamp

### LINKED ISSUE #125: Finalize interface before the v.1.0 release
We really need to merge v.1.0 into master sooner rather than later, so we should finalize the interface for `predict` and `learn`, since we can fix any arising bugs and add features without breaking BC if we are happy with the interface.

Also somewhat relevant to produce the final website docs and video docs.

Personally, I'm happy with the current interface, but if anything comes to mind I'll mention it here.

George3d6:
Final interface for v.1.0, includes working implementations for `load` and `export` functions.

Also a fix regarding how number were processed in the ludwig backend. We can now handle the case when the number as something like `12134,243` we cast to the respective type (that is to say, int or float) after replacing the `,` with a `.`.

Closes #125


George3d6:
Should be noted that this should close #133

PR #156, opened 2019-04-12
Storage changes

### LINKED ISSUE #140: Change presistent data
A few changes should be made to the shared transaction state:

* Light persistent metadata (current to the similar persistent_transaction_metadata). Easy to save and transfer to a server (in order to generate pretty looking logs and aggregate statistics about various models). Ideally, this shouldn't go above 10-100MB or so.

* Heavy persistent metadata. This would include stuff like the Probabilistic Validator, or the weights of the ML model itself (unless the backend saves it separately).

* Non persistent metadata. This would include what's currently in the transaction metadata, but with less duplication (currently there's a lot of duplication between the persistent and non persistend metadata)

Changes will be made as to how this is also saved:

* The 2 persistent states will be pickled throughout the training and predicting and saved to files in mindsdb's storage directory. This should also deal with the concurrency issues outlined in #13.

* We'll also have some helper functions to move this pickled metadata into various databases (mongo, mariadb, postgres... etc), in case users want to use a database instead of a directory for storage. But they will be treated as binary blobs (unlike now, when the data in mongo is an actual BSON) and simply unpickled when they are read and overwritten when they are inserted/updated.

George3d6:
Split metadata into:

* light model metadata
* heavy model metadata

I use `lmd` and `hmd` to refer to the variables representing them in the transaction, since they are referenced a lot and I wanted to save visaul space.

Switched storage sch that it uses pickles instead of mongo/tindydb. This Resolves #140 

Removed a lot of unused code, mainly in the form of whole files and imports... always good to keep the codebase clean.


torrmal:
good, please update the requirements.txt

torrmal:
There are some conflicts
Conflicting files
mindsdb/libs/backends/ludwig.py

Can you update the requirements.txt
thank you

George3d6:
Resolved conflicts.

What do you mean by updating `requirements.txt` ? Prune dependencies that might no longer be needed ?

If so, I will do it as part of the following PR (replacing both types of metadata with plain python dictionaries).

PR #165, opened 2019-04-19
Ludwig online train

George3d6:
* Sourcing training stats (accuracy) from ludwig
* Allow models to be trained further by adding the `rebuild_model=False` argument to the `learn` method of the Preidctor

Initially the intention was to switch to using ludwig's `online_train`, but it's apparently garbage in many ways (see comments in the ludwig code and issue https://github.com/uber/ludwig/issues/295), so for now we stuck with `train`.

sadly enough, this means we can't really control how long training takes or the number of epochs or gather our own information about the model whilst it's being trained :(


George3d6:
Note, this PR is in relation to: #162 but doesn't actually close it, still there's apparently no way to use online_train, neither initially or for re-training the model.

torrmal:
im starting to get confused what does hmd and lmd stand for in future lets use descriptive variables

PR #181, opened 2019-04-30
Fix to adapted metadata

George3d6:
A small fix to the way our metadata was being adapted for the mindsdb server which cause histograms to be processed incorrectly.


George3d6:
Doesn't work due to ludwig dependency issues,whops, merged it with the ludwig related PR

PR #191, opened 2019-05-05
Add details about tkinter (fixes #152)

### LINKED ISSUE #152: ModuleNotFoundError: No module named '_tkinter'
**Describe the bug**
Train new model with the latest mindsdb version throws an error. LudwigModel uses matplotlib that depends on tkinter backend.
With different OS tkinter is not shipped with Python, or if users install python from source.

**Screenshots**
Screenshot from 2019-04-07 16-24-12

**Desktop (please complete the following information):**
 - python 3.7.3 version

**Additional context**
According to [tkinter docs](https://wiki.python.org/moin/TkInter):
If it fails with "No module named _tkinter", your Python configuration needs to be modified to include this module (which is an extension module implemented in C). Do not edit Modules/Setup (it is out of date). You may have to install Tcl and Tk (when using RPM, install the -devel RPMs as well) and/or edit the setup.py script to point to the right locations where Tcl/Tk is installed. If you install Tcl/Tk in the default locations, simply rerunning "make" should build the _tkinter extension.

fabaff:
Add details collected in #152 to the documentation.

### DIFF
--- a/docs/website/docs/Installing.md
+++ b/docs/website/docs/Installing.md
@@ -5,15 +5,24 @@ title: Installing MindsDB
 
 ## Prerequisites
 
-Before you begin, you need **[python>=3.7](https://realpython.com/installing-python/)** or **[Conda Python3](https://www.anaconda.com/download/)**, and make sure you have the **latest pip3**
+Before you begin, you need **[python>=3.7](https://realpython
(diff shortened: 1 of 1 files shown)


George3d6:
thank you for adding this, better than nothing for now, thought I still find it weird that the version of matplotlib ludwig uses requires tk...

PR #194, opened 2019-05-07
Split file type finder

George3d6:
* Split the functionality of sniffing file type and parsing headers into independent functions to use on the server

* Fixed a bug in the stats generator that was causing crashes where all of a column's data was `None`


torrmal:
I dont know, this is working, it has been stable for a long time, I dont want to make this change now.

PR #197, opened 2019-05-09
fix_predict method

alexsukhrin:
do not do this = it is very difficult to catch

### DIFF
--- a/mindsdb/libs/backends/ludwig.py
+++ b/mindsdb/libs/backends/ludwig.py
@@ -8,6 +8,7 @@ import os, sys
 
 from ludwig.api import LudwigModel
 import pandas as pd
+import typing
 
 # @TODO: Define generci interface, similar to 'base_module' in the phases
 class LudwigBackend():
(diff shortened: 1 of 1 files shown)


George3d6:
Sorry, what's the exact purpose of this change... I don't quite get it ? 

It just seems to change the style in which optional arguments are listed to a more verbose but safer one... which is fine, but if we wanted to do it this way we'd have to re-write it all over mindsdb to keep the internals somewhat consistent.

Was `predict` causing you weird errors specifically due to the way arguments were given default values ?

alexsukhrin:
see example

>>> def saver(x=[]):
...     x.append(1)
...     print(x)
...
>>> saver([2])
[2, 1]
>>> saver()
[1]
>>> saver()
[1, 1]
>>> saver()
[1, 1, 1]

judging by how you write the code, it feels like you are a beginner, I just
gave you a hint how to avoid trouble

пт, 10 мая 2019 г. в 17:11, George <notifications@github.com>:

George3d6:
I think you may have miss-read what the function does, unless I am mistaken.

Both the `ignore_columns` and the `mode` arguments are treated as immutable in the function (as in, they are only read, not written to).

If you find any cases where this is no the case (i.e. where a default argument happens to be a heap assigned object that is then modified in the body of the function), please point them out to me.

The example you gave is simply how python's non-int object work in order to avoid extra memory consumption. 

For example: if the default arg for a function was `[1,2,3]` and the function was called a few thousand times this would avoid a new list being pointlessly allocated each time.

Whether or not this is a good design choice or not is something you should take with python's designers, I can certainly see the appeal of using strongly typed languages to make the code more clear and the appeal of having `=` guarantee the construction/creation of a new object each time, such as would be the case in C++ or Rust, where `a=b` would, by default, result in a copy of the heap memory being pointed to by `b`, rather than of the pointer `b` itself.

However, the choice has been made long ago to write mindsdb in python with some of the quirks that causes in regards to assigning variable.

If you wish to make everything typed and avoid any possibly pointer copy where deep copy was intended, well.... than that would require a re-write of the whole codebase and would make the code more cumbersome to write and/or read.

Whilst I do agree with you, in principle, that such mistakes are to be avoided, this sort of design should be applied to *all* of the codebase rather than a single function.

If you wish to engage in such a re-write, be my guest, the project is free to clone by anyone, but right now we don't have the time nor any strong reasons to do this ourselves I'm afraid.

PR #211, opened 2019-05-15
Proper cat detection

### LINKED ISSUE #187: Categorical value get interpreted as texts
Some of the categorical values from the kaggel suicide dataset get interpreted as `text` rather than `categorical`. This might have to do with the stats generator sample size being too small.

See #182 in order to see how this can be replicated.

George3d6:
Added more aggressive type detection for category and corrected instances where single(binary) and multiple categories were not distinguished between properly.

... spend more than 1 hour chasing a bug that didn't exist (`get_model_metadata` in the travis test was running on the wrong model which had an old metadata version), corrected the travis test to avoid this in the future.

For now I would say this resolves #187, though I think there's more room for improvement as far as category detection goes, progress towards fixing all of the issues described by and found with the help of #182


George3d6:
Fixed the merge conflicts

PR #229, opened 2019-05-22
Rename function added to predictor class

aliakbar09a:


George3d6:
Hi, thank you for the PR, but sadly enough this rename function doesn't quite do everything that's intended.

The name of the model is currently used to name & load the model files (which have to be renamed)

And the metadata held within those files ( the light metadata and the heavy metadata) have to be modified:

`light_transaction_metadata['name']` and `heavy_transaction_metadata['name']`

The `ludwig_save_path` (inside the light metadata), the ludwig model name `model_name` property of the ludwig model and the ludwig file (same name as `ludwig_save_path` variable) also have to be modified.

There might be a few other places where the `self.name` variable is propagated... but I can't think of or see any right now.

George3d6:
No activity for a month => Will close.

If you wish to add this feature again in the future please feel free to open a new PR :)

PR #239, opened 2019-06-07
Lightwood backend integration

George3d6:
Integration of lightwood backend.

Only `train` is integrated at the moment and timeseries data isn't being handled.

Lightwood starts running but crashes with the error:

```
 '    from lightwood.config.config import CONFIG\n'
 "ModuleNotFoundError: No module named 'lightwood.config'\n")
```

So we probably need to fix lightwood before moving forward (using the latest version on master at the moment).

Still, this lays out a lot of the integration's foundations (updated requirements, changes to the ModelInterface, most of the lightwood backend... etc). So I think we can merge into master and go from there, keeping ludwig as the default backend.

Alternatively, I can keep working on this branch until the integration is complete.


torrmal:
will look int the error you are getting

George3d6:
I didn't realize the crash happened when lightwood was imported, I see that just now, I thought the crash happened when `.train` was called via the lightwood API.

Since merging this would break master I will close the PR until lightwood is fixed.

PR #241, opened 2019-06-11
Better date types handling

George3d6:
Fixed a few bugs in the stats generator (or possible causes for confusion).

The main one being that date and timestamp types were not being analyzer correctly once the type was determine (due to incorrect iteration through a column in the data frame using `enumerate`)


George3d6:
Also added a fix in `transaction.py` for a module which is not present with all python distros apparently.

PR #295, opened 2019-08-29
Check if column is Boolean before casting #294

### LINKED ISSUE #294: ValueError: could not convert string to float: 'True'
**Your Environment**
* Python version: 3.7.3
* Operating system: Ubuntu 18.04
* Python environment used (e.g. venv, conda): venv
* Mindsdb version: 1.4.6

**Describe the bug**
Predict input value e.g `Number of times pregnant: 1`  is somehow converted to True instead of 1.
Inside [base_module](https://github.com/mindsdb/mindsdb/blob/f7ca6b842e2d0b29c9a0a465454bf772097431ee/mindsdb/libs/phases/base_module.py#L54) if you check the *kwargs data_frame value contains True as an input value. So the actual mapping error should be somewhere in transaction.py

**To Reproduce**
Steps to reproduce the behavior, for example:
1. Link to [full-example](https://github.com/ZoranPandovski/mindsdb-examples/tree/master/diabetes_example) 

**Additional context**
Screenshot from 2019-08-20 16-14-32

ZoranPandovski:
This PR checks if the data frame value is boolean before casting it to 'True' or 'False'. Fixes #294

### DIFF
--- a/mindsdb/libs/phases/data_extractor/data_extractor.py
+++ b/mindsdb/libs/phases/data_extractor/data_extractor.py
@@ -4,6 +4,7 @@ from mindsdb.libs.phases.base_module import BaseModule
 from mindsdb.libs.data_types.mindsdb_logger import log
 from mindsdb.libs.helpers.text_helpers import hashtext
 from mindsdb.external_libs.stats import calculate_sample_size
+from pandas.api.types import is_numeric_dtype
 
 import random
 import traceback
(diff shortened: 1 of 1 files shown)


George3d6:
Why is this extra check required ?
Is there a situation where the original way of doing replace fails to replace all True/False values ? or a situation in which we want to keep True and False values in the dataframe as python bools instead of as strings ?

George3d6:
Also, you only want to do the casting on the column it should be something like `df[col] = df[col].replace(... etc`, right ? Or am I miss-understanding what you are trying to do here :?

George3d6:
Ah, nevermind, I see the issue described in #294 , but my second comment still stands. Also, this doesn't fix cases of this bugs where pandas read a float column that's all 0 and or 1 in the when/when_data dataset and turns that into a bool. I think the easier fix will be:

If the column type is string/object in pandas, do the replace as is.
If the column type is numerical/bolean, do the replace using the dict: ` {True: 1, False: 0}`

Does that seem like a good solution to you @ZoranPandovski  ?

ZoranPandovski:
I've made the required changes as we've talked.

PR #302, opened 2019-09-09
fix getting started link throwing 404

mohsinkhansymc:
Getting started link in footer is redirecting to an invalid url and git displays `404 page not found`. After this PR it is redirecting to the valid getting started page

image


George3d6:
Huh, nice find, thanks a lot.

PR #306, opened 2019-10-01
Add comparison between Mindsdb and other libraries

### LINKED ISSUE #45: Provide comparison to other DL frameworks
**Is your feature request related to a problem? Please describe.**
Not immediately clear to newcomer how this compares/better to just using torque/keras etc

**Describe the solution you'd like**
Provide clear and concise description of how mindsdb differ pointed to straight in the README

sk-ip:
This documentation is added for comparing how Mindsdb
is different from other libraries and frameworks. It
contains some key points and code samples of different
libraries for comparison. closes #45

### DIFF
--- a/docs/website/docs/ComparsionWithOtherLibraries.md
+++ b/docs/website/docs/ComparsionWithOtherLibraries.md
@@ -1,0 +1,166 @@
+---
+id: comparison-mindsdb
+title: Comparison with other Libraries
+---
+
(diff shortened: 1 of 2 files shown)


George3d6:
Thank you for adding this, the check are failing at the moment but that's due to a dependency issue with a dependency of a dependency of a dependency :p 

Will merge this and it will get auto-published alter tonight once I fix the issue.

Thanks for providing the code comparison, that might really help a lot of people figure out what mindsdb is about.

Greatly appreciated :)

PR #317, opened 2019-10-23
Add an error check for \r\n and \n type of strings

### LINKED ISSUE #316: Unbound local error
**Your Environment**
anaconda 
* Python version: 3.7.4
* Pip version:19.2.2
* Operating system:Windows
* Python environment used (e.g. venv, conda): conda
* Mindsdb version you tried to install:1.6.15
* Additional info if applicable:

**Describe the bug**
got Unbound local error while running this example 
[https://github.com/ZoranPandovski/mindsdb-examples/tree/master/air_quality](https://github.com/ZoranPandovski/mindsdb-examples/tree/master/air_quality)
**To Reproduce**
Steps to reproduce the behavior, for example:
1. clone the repository
2. Run that example code in a jupyter notebook and you should see the error as presented in the screenshot.

**Expected behavior**
It should start the training.

**Additional context**

sk-ip:
problem :
the file_ds.py was not able to find the type of file
given to it due to first_few_lines giving \r\n
so it was not able to find what type of delimiter is used and throws an
error.

Solution :
to fix this an if statement is added to check if line is \r\n or
\n then it will continue else it will append that line in
first_few_lines.

closes #316

### DIFF
--- a/mindsdb/libs/data_sources/file_ds.py
+++ b/mindsdb/libs/data_sources/file_ds.py
@@ -124,8 +124,10 @@ class FileDS(DataSource):
             first_few_lines = []
             i = 0
             for line in data:
-                i += 1
+                if line in ['\r\n','\n']:
+                    continue
                 first_few_lines.append(line)
+                i += 1
                 if i > 0:
                     break
 


George3d6:
I fail to see how this addresses anything @ZoranPandovski @sk-ip 
What file was this tested on ? It doesn't seem to be the case that air quality has the first line == `\n`, nor would I see the case for supporting such a weird format.

Could you provide me an example file that this PR addresses so I can look at the behavior.

PR #318, opened 2019-10-24
Improve extractor

George3d6:
## Main things

* Data shuffling during extraction now uses a seed (total length of the data) to obtain more consistency between mindsdb training runs

* Got rid of the indexes external to the data frame from everywhere *but* the data extractor where I still keep them because I want to re-work that more thoroughly and it's tied in with re-factoring how time-series work and for now I don't want to break it... for now it works, and the external-index based logic is de-coupled from where where else:
  1. Got rid of external index usage in ludwig backend
  2. Got rid of external index usage in data transformer
  3. Got rid of external indexes in the `TransactionData` object

* Decoupled Stats generator from needing a train/test/validation split, now it just relies on whatever dataframe gets past to it. This will allow us to run it before most of the logic in the Extractor (but, it would require some minimalist splitting of the extractor), thus the extractor could take into account the target variable's value distribution when creating the testing/validation/training set.

### Also
* Fix to accuracy testing
* Removed some dead code

### DIFF
--- a/mindsdb/__about__.py
+++ b/mindsdb/__about__.py
@@ -1,6 +1,6 @@
 __title__ = 'MindsDB'
 __package_name__ = 'mindsdb'
-__version__ = '1.6.18'
+__version__ = '1.7.0'
 __description__ = "MindsDB's goal is to make it very simple for developers to use the power of artificial neural networks in their projects."
 __email__ = "jorge@mindsdb.com"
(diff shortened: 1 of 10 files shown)


George3d6:
Note, labeled as WIP but I will probably merge if tests pass and continue the targe-variable splitting in another PR.

PR #325, opened 2019-11-04
Add PyPI downloads

ZoranPandovski:
Add badge for monthly downloads from PyPI


George3d6:
In the future please update the version in `__about__.py` for any PR, otherwise we'll get travis errors and it won't be deployed. In this case I'll merge my PR right after this one, so it'll be fine.

PR #339, opened 2019-11-25
ci tests formatted and added few

surendra1472:


George3d6:
What's the advantage being added here by using the `BasicTest` object rather than just keeping the test as a simple function ?

surendra1472:
@George3d6 I used python unittest package. The main advantage is that all tests will run parallelly so our ci pipeline will be faster. Apart from that, we can get test reports and ofc readability, easily we can trace which test case is failing(debug that test case only to fix) and add more test cases easily.

surendra1472:
And I added more test cases.

PR #340, opened 2019-11-26
ci tests:main method to run

surendra1472:


George3d6:
Ok, this logic is overly complicated, too long, and makes running tests locally harder. It adds nothing compared to *not* using the Unittest module, I will not merge this and revert back to the way we ran tests previously + add the new adapted metadata test cases.

PR #342, opened 2019-11-30
Update PredictorInterface.md

ZoranPandovski:
Include new Predictor parameter in the documentation


ZoranPandovski:
Chocolatey is still Experiencing disruptions, so the windows build will fail until they resolve the issues.

PR #350, opened 2019-12-11
Fix scikit version to work on windows

ZoranPandovski:
Fix scikit version that works on windows. Same fix as lightwood.


George3d6:
Added version bump and modified lightwood requirements so that install on window due to old lightwood versions will no longer fail. Will merge once tests pass (Or feel free to do so @ZoranPandovski once all tests are green).

PR #373, opened 2020-01-08
Travis Build not failing for script failure

ritwik12:
Solves #352.

Travis build was not failing as there was no option set in the `script` tag of travis.yml for taking actions if it fails. At present `script` tag is only executing tests script and as those scripts are running so there are no issues found and build will pass.

As per present scenerio the build will fail only if the scripts are not excuted such as the file which was executed using those scripts was not present there.

I guess adding `set -e` option will cause it to fail if there occurs an error in the script.
This was the only solution that came to my mind. We can use `pytest` or other tests rather than just running scripts manually if we want Travis to catch it automatically. For the present code, I think this should work.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -39,6 +39,7 @@ install:
 before_script: cd tests/ci_tests
 # run tests
 script:
+  - set -e
   - if [ "$TRAVIS_OS_NAME" = "windows" ]; then travis_wait 60 python fast_test.py; fi
   - if [ "$TRAVIS_OS_NAME" != "windows" ]; then travis_wait 60 python3 full_test.py; fi
   - cd ../..


ritwik12:
@George3d6 ^^

ritwik12:
Build is failing for OSX in travis `Services are not supported on osx`. I guess we should remove services when running travis on OSX hosts.

ZoranPandovski:
Let me restart the job just to make sure. I think it should pass on OSX also.

George3d6:
Rather weird since the failure seems to happen before the run the `full_test.py` file, so no changes should have happened there.

ritwik12:
Yeah that's weird @George3d6

George3d6:
Fixed the OSX issues, not related to this PR, merging now. Once again, thank you @ritwik12

ritwik12:
@George3d6 Thanks :)

PR #391, opened 2020-02-19
Add SageMaker integration docs

ZoranPandovski:
This PR includes new navbar menu Integrations with sections about:
*  Aws SageMaker and Mindsdb integration
* MindsDB container details
* Use MindsDB container in Sage
* Call endpoint code example


ZoranPandovski:
@George3d6 It should be valid. "aws-sagemaker" and {"type": "subcategory",} are both items in the "Integrations" array, we don't assign them as a key value.

PR #408, opened 2020-03-13
Fix delete model

ZoranPandovski:
This PR fixes the delete Predictor action. This also affects the GUI delete option


George3d6:
#407 already had the fix but I forgot to merge it, so no need for this PR, sorry :(

PR #425, opened 2020-04-27
Cleanup exceptions

ZoranPandovski:
This PR removes the sys.exit logic, and use ValueError and Exception when needed. @George3d6 Maybe it would be better if we raise RuntimeError error in __ini__.py for the wrong python version?


George3d6:
@ZoranPandovski , as always, bump the version when making any changes to mindsdb itself.

Other than that looks fine, I'll merge once the version is increased.

PR #441, opened 2020-05-13
Cleanup and getColPermutations logic fix

maximlopin:
`mindsdb/scraps/getColPermutations()`'s name says **permutations** while in docstring it says **combinations**, so I made two functions both for combinations and permutations that work according to their description.

**PEP-8:**
_Object type comparisons should always use isinstance() instead of comparing types directly._
_Use `.startswith()` and `.endswith()` instead of string slicing to check for prefixes or suffixes._

### DIFF
--- a/mindsdb/libs/backends/ludwig.py
+++ b/mindsdb/libs/backends/ludwig.py
@@ -275,14 +275,14 @@ class LudwigBackend():
                     data[tf_col].append(unix_ts)
 
                 elif data_subtype in (DATA_SUBTYPES.FLOAT):
-                    if type(row[col]) == str:
-                        data[tf_col].append(float(str(row[col]).replace(',','.')))
+                    if isinstance(row[col], str):
+                        data[tf_col].append(float(str(row[col]).replace(',', '.')))
                     else:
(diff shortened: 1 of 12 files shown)


George3d6:
Ah, I'm awfully afraid that the function in scraps.py are basically old stuff that I took out and ended up not deleting because I thought they might be worth pasting back in at some point :/

But, I think we'll probably be using that function soon enough again for some things, so good thing that it's refactored.

Other than that I agree with the changes for starting matching and type detection.

However, I prefer checks for `None` to always be explicit, i.e. `if order_by` should be `if order_by is None`. See here: https://github.com/mindsdb/mindsdb/pull/441/files#diff-0885559c059524995d7ce333c369cb0cR594

This happens a few times. If possible please change this, if not I will do it soon(TM) and merge.

George3d6:
Actaully, I just now see the `if group_by` and `if order_by` check were in the code before, so my fault for having that kind of silly logic.

Anyway, to be changed for the 2.0 release when we'll be refactoring this file heavily anything.

Merging this now, thanks for the changes.

PR #443, opened 2020-05-15
[WIP] Feature/414: deduplicate data by default

manesioz:
Adding the default step of dataframe deduplication as part of the DataTransformer as per #414

### DIFF
--- a/mindsdb/libs/backends/ludwig.py
+++ b/mindsdb/libs/backends/ludwig.py
@@ -275,14 +275,14 @@ class LudwigBackend():
                     data[tf_col].append(unix_ts)
 
                 elif data_subtype in (DATA_SUBTYPES.FLOAT):
-                    if type(row[col]) == str:
-                        data[tf_col].append(float(str(row[col]).replace(',','.')))
+                    if isinstance(row[col], str):
+                        data[tf_col].append(float(str(row[col]).replace(',', '.')))
                     else:
                         data[tf_col].append(row[col])
 
                 elif data_subtype in (DATA_SUBTYPES.INT):
-                    if type(row[col]) == str:
-                        data[tf_col].append(round(float(str(row[col]).replace(',','.'))))
+                    if isinstance(row[col], str):
+                        data[tf_col].append(round(float(str(row[col]).replace(',', '.'))))
                     else:
                         data[tf_col].append(row[col])
 
--- a/mindsdb/libs/controllers/predictor.py
+++ b/mindsdb/libs/controllers/predictor.py
@@ -559,7 +559,7 @@ class Predictor:
         return self.get_model_data(model_name=None, lmd=light_transaction_metadata)
 
 
-    def learn(self, to_predict, from_data, test_from_data=None, group_by = None, window_size = None, order_by = None, sample_margin_of_error = 0.005, ignore_columns = None, stop_training_in_x_seconds = None, stop_training_in_accuracy = None, backend='lightwood', rebuild_model=True, use_gpu=None, disable_optional_analysis=False, equal_accuracy_for_all_output_categories=True, output_categories_importance_dictionary=None, unstable_parameters_dict=None):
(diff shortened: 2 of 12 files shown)


mindsdb/libs/phases/data_transformer/data_transformer.py
```diff
@@ -5,13 +5,14 @@
 
 import pandas as pd
 
-from mindsdb.libs.constants.mindsdb import *
-from mindsdb.libs.phases.base_module import BaseModule
-from mindsdb.libs.helpers.text_helpers import clean_float
-from mindsdb.libs.helpers.debugging import *
+from mindsdb.mindsdb.libs.constants.mindsdb import TRANSACTION_LEARN, DATA_TYPES, DATA_SUBTYPES
```
manesioz, line 8: For some reason every import path was not working on my VS code editor (probably a relative vs absolute import path issue), so if the prepended `mindsdb` is not needed let me know.

George3d6:
Alright, so the core concept is here, if you want I can merge this into a different branch and do the edits I mentioned bellow, if you have time and want to finish this then feel free to do them :)

### Required change:

Adding that extra `log` argument in `BaseModule` seems to have broken things, it's no necessary to add it since the `log` object is already present in the transaction object ( see bellow the line `self.log = self.transaction.log`).

The `DataTransformer` object should already have a `self.log` by virtue of inheriting from the `BaseModule`.

Also, this `self.log.info('NO DUPLICATE ROWS DETECTED')` should be remove, no need to tell the user we did nothing.

And this `self.log.info('DUPLICATE ROWS DETECTED -- AUTOMATICALLY REMOVING')` should be replaced with a log line that tells the user how many rows we dropped that's importantly *not* in all caps :P

There's no need for the `duplicated` check you do, instead just call the drop function, register the length of the dataframe before and after, if it's the same, no rows were dropped, otherwise some rows were dropped, should be cleaner and faster.

Unless I'm misunderstanding how the `duplicated` and `drop_duplicates` functions work, in which case please correct me.

Finally, the last *big* issue that I only now realize is that `input_data.data_frame` is the object we want to operate on while predicting, but not when training, when training we have:

```
input_data.train_df
input_data.test_df
input_data.validation_df
```

But by this point, since we've split the data, we can't really de-duplicate, so it might be worth adding this function to the `DataSplitter`  module instead, even if it's a bit uglier, something like:

`self.transaction.input_data.data_frame = self._deduplicate_dataframe(self.transaction.input_data.data_frame)`

Note, if you think that these changes are not necessarily required or if I'm misunderstanding something please tell me so :)

manesioz:
Hey @George3d6 I really appreciate your detailed and prompt response, sorry it took me so long to take a look at this. 

I agree with everything, sorry about breaking travis I'll fix that right away. 
Lastly and more importantly, thats a great observation regarding the fact that it seems that deduplicating data at the `DataTransformer` phase is too late since it's already been split for training. Agreed that `DataSplitter` seems to be the phase in which deduplication would make sense. 

Finally, I was wondering if you had any thoughts on the possible effects deduplication would have on various models performance? Do you think that it's assuming too much, thus possibly leading to a premature loss of information, or is it reasonable to assume that duplicates are generally the result of badly cleaned data? I suppose thats more of a discussion for the issue itself (not PR), but I'm curious to hear your thoughts.

maximlopin:
What about just adding an `if` statement to check whether it's a prediction or a learn transaction? 
Based on that, deduplicate either `input_data.data_frame` or all: `input_data.train_df`, `input_data.test_df`, `input_data.validation_df`.
Or maybe it even should be its own new phase coming right before `DataSplitter`.

Also, what if my data is thousands rows of random input to an AND gate? It will only keep 4 rows of data, so user should be `log.warned` about deduplication and must have an option to disable it when calling `Predictor.learn`.

George3d6:
@maximlopin the problem with that is that we need to clean the duplicates from all the dataset, i.e. a duplicate rows in the form of the same row in the test and train data frame is still an issue.

@manesioz you can use [this](https://github.com/mindsdb/mindsdb/blob/de99a7e90c33c6c3e1f762fed4247bae306a4c4d/mindsdb/libs/phases/data_splitter/data_splitter.py#L8) recently added function as a template. Also, if possible, pull the new `master` and bump the minor version in __about__.py

As for your question on the effect of duplicate removal.... there are situations, although rare, when I can see the user wanting to put emphasis/require overfitting of certain rows.

Duplicating them would not be an ideal way to do it, since they can't control how many will go in train, how many in test and how many in validation. A better way to do it is weighted sampling of specific rows during training, which we can do. But implementing an option for this would be annoying and the interface would be... complicated. So I'll make a note of it but I don't think it's something worth bothering with unless you or someone else needs this.

manesioz:
Going to start this from a fresh branch

PR #445, opened 2020-05-16
Numerical accuracy changes

George3d6:
* Calculating numerical accuracy in the `ProbabilisticValidator` based on whether or not the real values falls within the predicted range.

* Making the "will fall in the numerical range" the target of the Bayesian model used for confidence determination in the `ProbabilisticValidator` for numerical targets.

* Getting a sample of max 400 predicted-ranges + real-values from the validation datasets, this sample can be used to plot the accuracy graph for numerical values instead of the confusion matrix.


George3d6:
Oh, also, this resolves #434

PR #446, opened 2020-05-18
Add import error messages

ZoranPandovski:
This PR adds ImportErrors and informative messages for the missing extra_data_sources requirements.

### DIFF
--- a/mindsdb/__about__.py
+++ b/mindsdb/__about__.py
@@ -1,6 +1,6 @@
 __title__ = 'MindsDB'
 __package_name__ = 'mindsdb'
-__version__ = '1.17.6'
+__version__ = '1.17.7'
 __description__ = "MindsDB's goal is to make it very simple for developers to use the power of artificial neural networks in their projects."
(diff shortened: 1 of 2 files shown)


George3d6:
@ZoranPandovski please

1. Merge current master
2. bump the version
3. replace `except ImportError:` with `except:`

ZoranPandovski:
@George3d6 The way I've added it wouldn't work because the __init__ will be called whenever you are using extra_data_source or not. That means the ImportError exception will be raised all the time. Maybe it would be better if we just log a warning in `except` like :
```
try:
    from mindsdb.libs.data_sources.postgres_ds import PostgresDS
except:
    log.warning("PostgresDS Datasource is not available by default. Please install psycopg2 or mindsdb[extra_data_sources]")
```

George3d6:
That works, but don't add "please install" instead say "If you wish to use it, please install... etc"

George3d6:
That's because a lot of people might not need those data sources.

PR #449, opened 2020-05-20
feat/414: add _deduplicate_data() func as preprocessing step in DataSplitter

### LINKED ISSUE #414: Deduplicate data
Remove duplicates row from the training and testing set.

If the user is intentionally duplicating rows in order to introduce bias in the way the model learns it should also be helpful to warn him that there's already option to do that via the weight map.

<Also a reminder to check the logic for removing rows with null targets, because I'm pretty sure that's implemented based on  behavior but I forget where>

manesioz:
Fixes #414 

Incorporated discussion from #443

### DIFF
--- a/mindsdb/libs/phases/data_splitter/data_splitter.py
+++ b/mindsdb/libs/phases/data_splitter/data_splitter.py
@@ -1,20 +1,31 @@
-from mindsdb.config import CONFIG
-from mindsdb.libs.constants.mindsdb import *
-from mindsdb.libs.phases.base_module import BaseModule
-from mindsdb.libs.data_types.mindsdb_logger import log
+from mindsdb.mindsdb.config import CONFIG
+from mindsdb.mindsdb.libs.constants.mindsdb import TRANSACTION_LEARN
+from mindsdb.mindsdb.libs.data_types.mindsdb_logger import log
+from mindsdb.mindsdb.libs.phases.base_module import BaseModule
 
 
 class DataSplitter(BaseModule):
     def _cleanup_w_missing_targets(self, df):
         initial_len = len(df)
         df = df.dropna(subset=self.transaction.lmd['predict_columns'])
-        no_dropped = len(df) - initial_len
+        final_len = len(df) 
+        no_dropped = final_len - initial_len
         if no_dropped > 0:
             self.log.warning(f'Dropped {no_dropped} rows because they had null values in one or more of the columns that we are trying to predict. Please always provide non-null values in the columns you want to predict !')
+        return df, final_len
+
+    def _deduplicate_data(self, df, df_len): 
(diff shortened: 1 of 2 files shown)


mindsdb/libs/phases/data_splitter/data_splitter.py
```diff
@@ -1,20 +1,31 @@
...
 class DataSplitter(BaseModule):
     def _cleanup_w_missing_targets(self, df):
         initial_len = len(df)
         df = df.dropna(subset=self.transaction.lmd['predict_columns'])
-        no_dropped = len(df) - initial_len
+        final_len = len(df) 
+        no_dropped = final_len - initial_len
         if no_dropped > 0:
             self.log.warning(f'Dropped {no_dropped} rows because they had null values in one or more of the columns that we are trying to predict. Please always provide non-null values in the columns you want to predict !')
+        return df, final_len
```
manesioz, line 15: To avoid computing the length of the same dataset twice, I passed this along to the next pre-processing step

mindsdb/libs/phases/data_splitter/data_splitter.py
```diff
@@ -1,20 +1,31 @@
-from mindsdb.config import CONFIG
-from mindsdb.libs.constants.mindsdb import *
-from mindsdb.libs.phases.base_module import BaseModule
-from mindsdb.libs.data_types.mindsdb_logger import log
+from mindsdb.mindsdb.config import CONFIG
+from mindsdb.mindsdb.libs.constants.mindsdb import TRANSACTION_LEARN
```
manesioz, line 2: I needed to prepend every import statement with `mindsdb` to overcome an import error -- can be removed if needed.

manesioz:
cc. @George3d6 @maximlopin

mindsdb/libs/phases/data_splitter/data_splitter.py
manesioz: This seems to be breaking travis, fixing

manesioz:
@maximlopin raised a great point here: 

> Also, what if my data is thousands rows of random input to an AND gate? It will only keep 4 rows of data, so user should be `log.warned` about deduplication and must have an option to disable it when calling `Predictor.learn`.

What are your guys thoughts on the best way to accomplish this? Currently it is just part of the `DataSplitter` preprocessing steps. 

Also, I'm not sure why Travis is failing to load the `DataSplitter` module... Any ideas?

ZoranPandovski:
@manesioz There is an IndexError in DataSpliter https://travis-ci.com/github/mindsdb/mindsdb/jobs/338248695#L647 that's why travis fails.

George3d6:
Closing this, if you want to work on this further please PR in https://github.com/mindsdb/mindsdb_native instead. Thanks :)

PR #452, opened 2020-05-28
Data Analytics Backend V2

### LINKED ISSUE #435: Various improvements to the Data Analysis
There will be various improvements to the Data Analysis phase for version 1.20.0 (compatible with scout version 2.0).

For now I will keep track of all of them here, if some of them require their own issue we can open them later.

Currently still discussing the changes around this internally, two big things would be:

* Re-work some of the scores
* Add predictive power based correlation score. i.e. how correlated/similar two columns are based on whether or not they can predict each other. could also be expanded further into creating some sort of causal graph between certain columns or column combinations.

### LINKED ISSUE #450: Numerical Int histogram - Int only buckets
For a data subtype of integer, our histogram x-axis and buckets should begin/end at integer values.

George3d6:
* Resolves #450 
* Resolves #435 (Not fully, but what it doesn't, is now part of other issues)
* Adds a new data cleaning phase that runs multiple times (much less complex than other phases).
* Major refactoring to the StatsGenerator that greatly simplify the logic and remove a bunch of bugs, that being said, the original `stats` object returned by it should remain unchanged, all the new stuff is inside `stats_v2`, which will soon (TM) fully replace it.

### DIFF
--- a/mindsdb/libs/phases/stats_generator/data_preparation.py
+++ b/mindsdb/libs/phases/stats_generator/data_preparation.py
@@ -1,4 +1,5 @@
 import random
+import numpy as np
 
 from mindsdb.external_libs.stats import calculate_sample_size
 from mindsdb.libs.constants.mindsdb import *
--- a/mindsdb/libs/phases/stats_generator/data_preparation.py
+++ b/mindsdb/libs/phases/stats_generator/data_preparation.py
@@ -3,9 +3,9 @@ import numpy as np
 
 from mindsdb.external_libs.stats import calculate_sample_size
 from mindsdb.libs.constants.mindsdb import *
+from mindsdb.libs.helpers.text_helpers import clean_float
 
-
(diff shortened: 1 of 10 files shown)


mindsdb/libs/phases/stats_generator/data_preparation.py
```diff
@@ -0,0 +1,36 @@
...
+    log.info(f'Analyzing a sample of {sample_size} from a total population of {population_size}, this is equivalent to {sample_size_pct}% of your data.')
+
+    return df.iloc[input_data_sample_indexes]
+
+
+def clean_int_and_date_data(col_data, log):
+    cleaned_data = []
+
+    for ele in col_data:
+        if str(ele) not in ['', str(None), str(False), str(np.nan), 'NaN', 'nan', 'NA', 'null'] and ele != '' and ele != '\r' and ele != '\n':
```
ZoranPandovski, line 26: We can use `and (not ele.isspace())` that will check for space, tab, newline, carriage return and feed'\f'

PR #456, opened 2020-06-02
Manual (Sub)type setting

### LINKED ISSUE #451: Support user-decided type
Allow the user to configure the data type/subtype, for now do it via an unstable param, in the future data skillet might be responsible for providing these (so they will essentially be part of the data source).

George3d6:
* Added a method to the `DataSource` superclass that all datasource inherit to allow for manual setting of the data subtypes of a certain column. We only allow subtype setting (instead of typesetting or the ability to set either/or type and subtype) because:

a) We'd have to change the `StatsGenerator` quite a lot to find a subtype under specific type constraints and this proved too difficult
b) Setting the subtype also dictates the data type, so setting a subtype is enough to set both
c) I don't see many cases in which a user would want and know what type to set for his data better than mindsdb but not what subtype.

* Made mindsdb able to use said types and subtypes, when specified, instead of those inferred by the `StatsGenerator` (the type inference step is skipped for columns where the subtype is specified)
* Added a decent test suite to check if this works.

Example usage:

```
data_source_mod = FileDS('https://raw.githubusercontent.com/mindsdb/mindsdb-examples/master/benchmarks/german_credit_data/processed_data/test.csv')
data_source_mod.set_subtypes({'credit_usage': 'Int', 'Average_Credit_Balance': 'Text','existing_credits': 'Binary Category'})
```

This resolves #451

### DIFF
--- a/mindsdb/__about__.py
+++ b/mindsdb/__about__.py
@@ -1,6 +1,6 @@
 __title__ = 'MindsDB'
 __package_name__ = 'mindsdb'
-__version__ = '1.18.2'
+__version__ = '1.18.3'
 __description__ = "MindsDB's goal is to make it very simple for developers to use the power of artificial neural networks in their projects."
 __email__ = "jorge@mindsdb.com"
 __author__ = 'MindsDB Inc'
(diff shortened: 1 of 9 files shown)


George3d6:
@ZoranPandovski good point, removed that and all references to it.

PR #457, opened 2020-06-02
Clickhouse datasource: fixed query endings with semicolon

StpMax:
When creating ClickHouse DS from query ended with semicolon is error raised with unclear message.


StpMax:
done

PR #470, opened 2020-06-05
Explicit exception when file type is unsupported

maximlopin:
It used to cause `reference before assignment error` on the variable `header`. Now it is an explicit `raise ValueError` . Also I changed variable name from `format` to `fmt`, because `format` is a built-in function name.

### DIFF
--- a/mindsdb/libs/data_sources/file_ds.py
+++ b/mindsdb/libs/data_sources/file_ds.py
@@ -158,32 +158,32 @@ class FileDS(DataSource):
 
         col_map = {}
         # get file data io, format and dialect
-        data, format, dialect = self._getDataIo(file)
+        data, fmt, dialect = self._getDataIo(file)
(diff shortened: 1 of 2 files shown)


mindsdb/libs/data_sources/file_ds.py
```diff
@@ -158,32 +158,32 @@ def _setup(self,file, clean_rows = True, custom_parser = None):
 
         col_map = {}
         # get file data io, format and dialect
-        data, format, dialect = self._getDataIo(file)
+        data, fmt, dialect = self._getDataIo(file)
```
George3d6, line 161: 1. As a general rule of thumb avoid name changes for the sake of name changing (i.e. if you can shorthand something that's used 20x times from 20 to 5 character do it, 6x times from 5 to 3 is pointless). It makes it harder to track things across versions.

PR #473, opened 2020-06-05
Mark outlier buckets

### LINKED ISSUE #460: Mark outlier buckets
In `stats_v2`, instead of returning the distinct outlier values in `stats_v2[col_name]['outliers']` we should instead have a list of `outlier_buckets` with buckets containing the most amount of outliers.

Where "amount of outliers" means over half the bucket is outlier and/or the bucket has at least one outlier and it's in the top 5% as far as outlier content.

maximlopin:
Resolves #460

### DIFF
--- a/mindsdb/libs/phases/stats_generator/stats_generator.py
+++ b/mindsdb/libs/phases/stats_generator/stats_generator.py
@@ -4,7 +4,7 @@ import warnings
 import imghdr
 import sndhdr
 import logging
-from collections import Counter
+from collections import Counter, defaultdict
 
 import numpy as np
 from scipy.stats import entropy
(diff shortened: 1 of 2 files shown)


mindsdb/libs/phases/stats_generator/stats_generator.py
```diff
@@ -647,6 +648,32 @@ def run(self, input_data, hmd=None, print_logs=True):
...
+                    bucket_outliers_num = len(outlier_values)
+
+                    total_outliers_num = len(stats_v2[col_name]['outliers']['outlier_values'])
+                    
+                    # Are half of values in the bucket outliers?
+                    predominantly_outlier = bucket_values_num > 0 and (bucket_outliers_num / bucket_values_num) > 0.5
+
+                    # Does number of outliers in the bucket make 95% of outliers
+                    # among all buckets?
+                    percentile_outlier = bucket_outliers_num > 0 and total_outliers_num > 0 and (bucket_outliers_num / total_outliers_num) > (1 - 0.05)
```
George3d6, line 672: Hmh, so this is not what we had in mind. What I mean by that was "the number of outliers in this bucket is in the 95th percentile".

Basically meaning that out of e.g. 100 buckets, the 95th percentile are the 5 with the highest number of outliers proportional to the nr of element in the bucket (see np.percentile for example/helper func)

George3d6: So this would have to be computed outside the for once you already know the nr of outliers in each bucket.

PR #475, opened 2020-06-09
#463 Splitting StatsGenerator into TypeDeductor and DataAnalyzer phases

btseytlin:
* TypeDeductor only infers types and computes type distributions in columns
* DataAnalyzer computes scores, but warnings based on scores were removed
* pytest and pytest-randomly introduced
* pytest included into CI


George3d6:
@btseytlin the tests seem to fail with a weird error: `self.transaction.model_backend = LudwigBackend(self.transaction)`

... strange, Ludwig should never be imported unless the optional-version that has ludwig is installed or ludwig is specified as a backend (which it shouldn't be in the tests).

Please look into it and try to get the continuous integration tests to run locally.

btseytlin:
Done, passed. Apparently running tests in the `ci_tests` dir from anywhere except that dir breaks everything

mindsdb/libs/controllers/predictor.py
```diff
@@ -37,7 +37,7 @@ def __init__(self, name, root_folder=CONFIG.MINDSDB_STORAGE_PATH, log_level=CONF
         if CONFIG.CHECK_FOR_UPDATES:
             try:
                 check_for_updates()
-            except:
+            except Exception:
```
George3d6, line 40: What's the point of explicitly catch `Exception` if we don't log it (e.g. `except Exception as e` + `print(e)`)  ?

I'm not saying we should log them, just not sure why making the `Exception` explicit is superior.

mindsdb/libs/helpers/text_helpers.py
```diff
@@ -15,15 +15,25 @@
...
 def clean_float(val):
     if isinstance(val, (int, float)):
         return float(val)
 
     if isinstance(val, numpy.float64):
         return val
 
-    val = str(val)
-    val = val.replace(',','.')
+    val = str(val).strip()
```
George3d6, line 35: Make the `.strip()` explicit (i.e. add the chracters being stripped)

tests/unit_tests/test_type_deductor.py
```diff
@@ -0,0 +1,151 @@
...
+        input_dataframe['numeric_float'].iloc[:2] = 'random string'
+        input_data = TransactionData()
+        input_data.data_frame = input_dataframe
+        type_deductor.run(input_data)
+
+        stats_v2 = lmd['stats_v2']
+        assert stats_v2['numeric_float']['typing']['data_type'] == DATA_TYPES.NUMERIC
+        assert stats_v2['numeric_float']['typing']['data_subtype'] == DATA_SUBTYPES.FLOAT
+        assert stats_v2['numeric_float']['typing']['data_type_dist'][DATA_TYPES.NUMERIC] == 97
+        assert stats_v2['numeric_float']['typing']['data_subtype_dist'][DATA_SUBTYPES.FLOAT] == 97
```
George3d6, line 151: I'm getting some warning about `no newline at the end of file` here, but I'm not sure this is relevant ?

George3d6:
Ok, I don't have time to review fully today, at least not now, will do so tonight and or tomorrow morning, sorry for the delay, but the PR is fairly large.

mindsdb/libs/controllers/predictor.py
btseytlin: This will swallow `KeyBoardInterrupt`:
```
try:
<whatever>
except:
<handle>
```
`except Exception` will not.

It's just annoying to have code not react to your Ctrl+C when running. Also just a good practice.

You can try this code:
```
while True:
  try:
   pass
  except:
   pass
```
This will hang forever and nothing can unhang it (except process kill). If you add `except Exception` it's fixed.

Thought it would be better to handle all exceptions of course.

tests/unit_tests/test_type_deductor.py
btseytlin: Ah yes, empty line at the end of file is a convention I always forget to follow :) fixed

btseytlin:
Added checks to ensure `lmd` is json-serializable after both phases

maximlopin:
1. Why check if key exists in the dictionary? It's never assigned to the dictionary but here.
https://github.com/mindsdb/mindsdb/blob/463-splitting-stats-generator/mindsdb/libs/phases/type_deductor/type_deductor.py#L293

2. Outlier buckets are missing (https://github.com/mindsdb/mindsdb/blob/master/mindsdb/libs/phases/stats_generator/stats_generator.py#L687).
I think they can become a score function.

3. Sample is computed twice.
https://github.com/mindsdb/mindsdb/blob/463-splitting-stats-generator/mindsdb/libs/phases/data_analyzer/data_analyzer.py#L200
https://github.com/mindsdb/mindsdb/blob/463-splitting-stats-generator/mindsdb/libs/phases/type_deductor/type_deductor.py#L249
I think we can either pass in through `transaction.lmd`, or create a wrapper around `pandas.DataFrame`, that will look something like this:
```
class DataObject():
    def __init__(self, df, margin_of_error, confidence_level):
        self._df = df
        self._sample_size = compute_sample_size(margin_of_error, confidence_level)
        self._sample = None

    @property
    def sample(self):
        if self._sample is None:
            self._sample = self._df.sample(self._sample_size)
        return self._sample
```
It would go through all phases, and each of them would be able to safely access sample without worrying if it has already been computed. This might be applicable to operations other than sample. What do you guys think?

4. Also, `stats_v2` doesn't contain outputs of scores, is that intentional?

PR #496, opened 2020-06-15
Convert (some) old tests to new structure

btseytlin:
Related: https://github.com/mindsdb/mindsdb/issues/482
First iteration of converting old tests to new structure:
* Predictor tests introduced
* `functional_tests.py` moved to pytest Predictor tests
* `DataAnalyzer.test_empty_values` is back
* Added ability to run slow tests in pytest via adding a `--runslow` flag. Without these flags these tests are not ran
* CI now prints top-20 slowest tests


btseytlin:
I will make an issue regarding the bug in ProbabilisticValidator that causes this test to fail:
https://github.com/mindsdb/mindsdb/pull/496/files#diff-27eb329a19e2e57da6f47204a201a255R97

The test serves as code to reproduce the bug

mindsdb/libs/phases/data_analyzer/data_analyzer.py
```diff
@@ -330,7 +328,7 @@ class DataAnalyzer(BaseModule):
     """
 
     # @TODO get rid of scores and stats entirely
-    def compute_scores(self, col_name, sample_df, full_df, stats):
+    def compute_scores(self, col_name, sample_df, full_data_dict, stats):
```
George3d6, line 331: Is the thing that's it's receiving a dict rather and a data frame ?

btseytlin: Yeah it is, so I renamed it to match

PR #498, opened 2020-06-16
Fix issue 495

StpMax:
This is an idea how to determine delimiter before read_csv (as it cant do it). It works, but may be someone know how do it better.

### DIFF
--- a/mindsdb/__about__.py
+++ b/mindsdb/__about__.py
@@ -1,6 +1,6 @@
 __title__ = 'MindsDB'
 __package_name__ = 'mindsdb'
-__version__ = '1.25.1'
+__version__ = '1.25.2'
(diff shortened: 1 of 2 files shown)


maximlopin:
`_getDataIo()` always returns a dialect if it's a .csv file (https://github.com/mindsdb/mindsdb/blob/master/mindsdb/libs/data_sources/file_ds.py#L161), so it can be simplified to:
```
try:
    return pd.DataFrame(file_list_data, columns=header), col_map
except Exception as e:
    return pd.read_csv(file, sep=dialect.delimiter), col_map
```

George3d6:
merging as is for now. @btseytlin and @maximlopin feel free to merge these in the future if you review them and they look correct.

It's enough for one of us to give a stamp of approval for anything but *very* important changes and that someone doesn't need to be me :)

PR #499, opened 2020-06-16
Converting integration tests to new structure

btseytlin:
* Moved two tests from `tests/integration_tests` to `test_predictor`.
* Dropped two other tests as they are somewhat redudant


George3d6:
@btseytlin can you just remove the `integration_tests` directory as a whole, since that will now contain nothing. I agree those tests are fairly outdated and we might as well get rid of them.

Once you do that feel free to merge this :)

PR #504, opened 2020-06-18
dont recompute dialect

maximlopin:


mindsdb/libs/data_sources/file_ds.py
```diff
@@ -183,23 +180,16 @@ def _setup(self,file, clean_rows = True, custom_parser = None):
...
             file_list_data = file_data
 
+        col_map = dict((col, col) for col in header)
+
         try:
             return pd.DataFrame(file_list_data, columns=header), col_map
-        except:
-            with open(file) as f:
-                f.readline()    # header
-                dialect = csv.Sniffer().sniff(f.readline())
```
George3d6, line 204: Is this bit of code no longer required ?

I.e. does the sniffing in `_getDataIo` work the exact same way thus making this one redundant ?

maximlopin: since we're calling pd.read_csv, we're assuming it's a csv file, therefore dialect must be defined as it's always returned by `_getDataIo` in case of a csv file (https://github.com/mindsdb/mindsdb/blob/master/mindsdb/libs/data_sources/file_ds.py#L138)

George3d6: Fair point, please bump the version in __about__.py and merge

maximlopin: Will do

PR #505, opened 2020-06-18
Sample once

### LINKED ISSUE #487: Sampling of data is done twice in `TypeDeductor` and `DataAnalyzer`
**Is your feature request related to a problem? Please describe.**
Both phases do a step of sampling data. Some time could be saved by computing it only once.
E.g. here:
https://github.com/mindsdb/mindsdb/blob/master/mindsdb/libs/phases/type_deductor/type_deductor.py#L249

**Describe the solution you'd like**
I am not familiar enough with the codebase to see which solution is best.

**Describe alternatives you've considered**
I see the following options:
* Cache the sample if it was already computed (can be done inside the sampling function or using a decorator)
* Do it as a separate phase and keep the sample in `transaction.heavy_meta_data`

**Additional context**
@maximlopin made an interesting suggestion on caching it inside a wrapper over `pd.DataFrame`:
https://github.com/mindsdb/mindsdb/pull/475#issuecomment-641952668

maximlopin:
Resolves #487 
As @George3d6 suggested, sampling is done once in `TypeDeductor` and stored in `transaction.input_data`

### DIFF
--- a/mindsdb/libs/phases/type_deductor/type_deductor.py
+++ b/mindsdb/libs/phases/type_deductor/type_deductor.py
@@ -244,14 +244,14 @@ class TypeDeductor(BaseModule):
         stats = defaultdict(dict)
         stats_v2 = defaultdict(dict)
 
-        self.transaction.sample_df = sample_data(
+        self.transaction.input_data.sample_df = sample_data(
             input_data.data_frame,
             self.transaction.lmd['sample_margin_of_error'],
             self.transaction.lmd['sample_confidence_level'],
             self.log
         )
 
-        sample_df = self.transaction.sample_df
+        sample_df = self.transaction.input_data.sample_df
 
         for col_name in sample_df.columns.values:
             col_data = sample_df[col_name].dropna()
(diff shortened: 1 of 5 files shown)


mindsdb/libs/phases/type_deductor/type_deductor.py
```diff
@@ -244,13 +244,14 @@ def run(self, input_data):
...
         stats_v2 = defaultdict(dict)
 
-        # Really bad that these parameters are implicitly passed through lmd
-        # Perhaps sampling can be moved somewhere upwards,
-        # so that it can be reused by all downstream phases?
-        sample_df = sample_data(input_data.data_frame,
-                                self.transaction.lmd['sample_margin_of_error'],
-                                self.transaction.lmd['sample_confidence_level'],
-                                self.log)
+        self.transaction.sample_df = sample_data(
```
George3d6, line 247: You modified the class of the `input_data` object not the `transaction` object.

So `self.transaction.sample_df` needs to be `self.transaction.input_data.sample_df`  ( I believe double-check the naming and try running all tests locally first)

George3d6:
@maximlopin tests still seem to be failing, please check out why.

PR #506, opened 2020-06-18
Docker for integration tests and replace datasource dependencies with pure-python

btseytlin:
This is a demonstration of how we could setup data source tests with docker.

Assuming you have Docker and docker-compose installed, to run the tests execute in mindsdb project dir:
```docker-compose -f integration-test-docker-compose.yml run --rm test_runner && docker-compose down```
This spins up MySQL, Postgres, Clickhouse containers, runs the datasource tests against them, and then removes everything and cleans up. Warning: the first time it runs it will take some time, as it will download the images for all databases and also mindsdb.

The project code is mounted inside the container that runs the tests, so you can change some code locally and rerun tests to see if it works, no rebuild needed. 

You can also run the ds tests locally as before. For this, you have to `cd tests/integration-tests` and run `pytest`. 

I think the most interesting is that with this we can add these tests to CI.

### CODE REVIEW

**Comment by ZoranPandovski** [comment]
This is great @btseytlin. Can we have multi-stage build in `tests/integration_tests/Dockerfile` because of the size of the Lightwood dependencies? It will minimize the size and improve run time when it is integrated in the CI.

**Comment by btseytlin** [comment]
> This is great @btseytlin. Can we have multi-stage build in `tests/integration_tests/Dockerfile` because of the size of the Lightwood dependencies? It will minimize the size and improve run time when it is integrated in the CI.

Do you mean saving the image to a registry, so that it doesn't have to be rebuilt from scratch every time?

**Comment by btseytlin** [comment]
Actually I think we can base the test runner image on the existing `mindsdb/mindsdb:latest` docker image and then it will save some time when installing dependencies

**Comment by George3d6** [comment]
I don't think time-saving is important. What I'm worried about is:

Even if the database itself runs in a docker container, we still need the `-dev` client dependencies (i.e. the ones with headers) for mysql and postgres in order to connect o mysql and postgres.

Does pip install handle these when running on the travis machiens ?

The tests seem to pass for now, but did you add the datasource integration/unit tests to travis or not yet ?

**Comment by btseytlin** [comment]
It seems `pip install -r optional_requirements_extra_data_sources.txt` installs all required dependencies for the mysql/postgres clients to work. 

I did not add these integration tests to travis yet

**Comment by btseytlin** [comment]
Good news!
We don't even need another Dockerfile for running tests. The workflow is simplified to be closer to what was there before.

Start databases:
```docker-compose up```
This runs databases with `network_mode: host`, so they run exactly as if they were launched locally. This can be used for both tests and development. You can run the databases locally as before, or you can use the docker-compose, but either way you don't need to change code at all.

And it runs the same way in tests/CI too, so we can actually test:
* Installing data source dependencies on each OS
* Running the tests on each OS

File: optional_requirements_extra_data_sources.txt [hunk 1]
```diff
@@ -1,3 +1,3 @@
 boto3 >= 1.9.0
```
George3d6, line 1: The idea behind these dependencies is that they should be install-abled by default, so feel free to add them to `requirements.txt` instead of here (and also make the datasources they pretain to imported in the main __init__.py by default, rather than inside a try/except)

tests/integration_tests/data_sources/test_postgres_ds.py
```diff
@@ -0,0 +1,32 @@
...
+    con = pg8000.connect(database=DBNAME, user=USER, password=PASSWORD,
+                         host=HOST, port=PORT)
+    cur = con.cursor()
+
+    cur.execute('DROP TABLE IF EXISTS test_mindsdb')
+    cur.execute(
+        'CREATE TABLE test_mindsdb(col_1 Text, col_2 Int, col_3 Boolean)')
+    for i in range(0, 200):
+        cur.execute(
+            f'INSERT INTO test_mindsdb VALUES (\'This is tring number {i}\', {i}, {i % 2 == 0})')
```
George3d6, line 23: I *think* pg8000 should have an insert syntax where arguments are passed separately *and* if not you can use `"""` instead of `'` to forgo the escaping of `'`... but whatever, minor nitpick, no need to change, just good to remember in the future.

.travis.yml
```diff
@@ -11,8 +12,6 @@ matrix:
       env:
         DEPLOYMENT_ENV="true"
         BADGE=linux
-      services:
```
George3d6, line 14: If we are only running docker tests on linux for now, I assume we should revert to docker only being a service undex linux ?

docker-compose.yml
```diff
@@ -0,0 +1,17 @@
+# This starts databases for local development or integration tests as if they were launched locally.
+# Run `docker-compose up -d`
+# Shut down and clean up: `docker-compose down`
+version: "3"
```
George3d6, line 4: In the next PR also add MariaDB as a 4th database here to run the test for it (even if they are the same as MySQL).

Also, is `alpine` the version for Postgres or just the linux-distro associated with the image? if the latter, it might be worth adding a specific version (say the last or second to last LTS release), but whatever, not high priority if it's difficult to find.

mindsdb/libs/data_sources/postgres_ds.py
```diff
@@ -13,37 +13,14 @@ def _setup(self, query=None, host='localhost', user='postgres', password='', dat
         if query is None:
             query = f'SELECT * FROM {table}'
 
-        con = psycopg2.connect(dbname=database, user=user, password=password, host=host, port=port)
+        con = pg8000.connect(database=database, user=user, password=password,
+                             host=host, port=port)
         df = pd.read_sql(query, con=con)
+        df.columns = [x.decode('utf-8') for x in df.columns]
```
George3d6, line 19: Hmhm, since it exhibits this weird behavior for strings it *might* just be worth it to add unit testing for a date-type to see what happens with those. And an array type if Postgres supports those by default (I forgot if it does)

tests/integration_tests/data_sources/test_postgres_ds.py
btseytlin: I agree, but for tests this will do since we don't need to protect form some injection here

docker-compose.yml
btseytlin: It's a linux distro, but its better to add the version too

mindsdb/libs/data_sources/postgres_ds.py
btseytlin: Good point, added

PR #513, opened 2020-06-22
Mariadb DS and test

btseytlin:


tests/integration_tests/data_sources/test_maria_ds.py
```diff
@@ -0,0 +1,62 @@
...
+
+    con = mysql.connector.connect(host=HOST,
+                                  port=PORT,
+                                  user=USER,
+                                  password=PASSWORD,
+                                  database=DATABASE)
+    cur = con.cursor()
+
+    cur.execute('DROP TABLE IF EXISTS test_mindsdb')
+    cur.execute('CREATE TABLE test_mindsdb(col_categorical Text, col_int BIGINT, col_bool BOOL, col_float FLOAT, col_date Text, col_ts Text, col_text Text)')
```
George3d6, line 24: Please check for the types: Date, Datetime, Timestamp, Time (see here: https://mariadb.com/kb/en/data-types/)

I think python datetime object will fit in the last 3 and the former should fit a python date (insert-able via calling `.date()` on  a datetime object) But you'd need to insert via the template syntax rather than using string format.

If template syntax doesn't work just test whichever are easy to test inserted this way.

tests/integration_tests/data_sources/test_maria_ds.py
```diff
@@ -0,0 +1,62 @@
...
+from mindsdb import Predictor
+from mindsdb.libs.constants.mindsdb import DATA_TYPES, DATA_SUBTYPES
+from mindsdb.libs.data_sources.maria_ds import MariaDS
+
+
+def test_maria_ds():
+    HOST = 'localhost'
+    USER = 'root'
+    PASSWORD = ''
+    DATABASE = 'mysql'
```
George3d6, line 13: Is this the default database for mariadb ?

mindsdb/libs/data_sources/maria_ds.py
```diff
@@ -0,0 +1,5 @@
+from mindsdb.libs.data_sources.mysql_ds import MySqlDS
+
+
+class MariaDS(MySqlDS):
```
George3d6, line 4: Copy-paste the code from the MysqlDS, since it's short *and* we might make divergent changes to these 2 later.

I know it's boilerplate but I plan on removing the boilerplate later from all database sources by having shared function between them, when we add more functionality, so that's fine.

tests/integration_tests/data_sources/test_maria_ds.py
```diff
@@ -0,0 +1,62 @@
...
+                       password=PASSWORD, database=DATABASE, port=PORT)
+    assert (len(maria_ds._df) == 200)
+
+    mdb = Predictor(name='analyse_dataset_test_predictor', log_level=logging.ERROR)
+    model_data = mdb.analyse_dataset(from_data=maria_ds)
+    analysis = model_data['data_analysis_v2']
+    assert model_data
+    assert analysis
+
+    assert analysis['col_categorical']['typing']['data_type'] == DATA_TYPES.CATEGORICAL
```
George3d6, line 43: Would be better to assert that the data-type distribution and the data subtype distribution contain *only* this type, since it may catch various edge cases (maybe?) and it's about the same length if using a counter or just checking like:

`assert(len(dist.keys()) == 1)`
`assert(len(dist[expected_type]) == nr_of_insert_itterations)`

btseytlin: Actually it takes a sample so data_type dist is not equal to the number of data points. I think we can check both

tests/integration_tests/data_sources/test_maria_ds.py
btseytlin: Seems like it

George3d6:
Perfect now, thanks :)

PR #514, opened 2020-06-23
Add warnings that TransactionOutputRow.explain() is deprecated

btseytlin:
Makes it harder to be confused by the old explainer, see:
https://github.com/mindsdb/mindsdb/issues/510

### DIFF
--- a/mindsdb/__about__.py
+++ b/mindsdb/__about__.py
@@ -1,6 +1,6 @@
 __title__ = 'MindsDB'
 __package_name__ = 'mindsdb'
-__version__ = '1.27.0'
+__version__ = '1.27.1'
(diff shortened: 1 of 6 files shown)


maximlopin:
are assertions like this: `assert explanation_new['predicted_value']`
supposed to check for `None`? If a value is equal to `0` then its boolean representation might trigger the `assert` statement. Explicit is better than implicit.
Everything else looks good to me.

btseytlin:
> are assertions like this: `assert explanation_new['predicted_value']`
> supposed to check for `None`? If a value is equal to `0` then its boolean representation might trigger the `assert` statement. Explicit is better than implicit.
> Everything else looks good to me.

Good point

PR #545, opened 2020-06-30
DB intergrations features

StpMax:
test for mariadb
$xxx_confidence meta columns
recreating mindsdb related tables on restart
other fixes


tests/integration_tests/test_mariadb.py
```diff
@@ -109,9 +115,180 @@ def setUpClass(cls):
...
+    while in_use is False and (time.time() - start_time) < timeout:
+        time.sleep(2)
+        in_use = is_port_in_use(port_num)
+
+    return in_use
+
 
 if __name__ == "__main__":
-    unittest.main()
+    maria_sp = subprocess.Popen(
```
btseytlin, line 270: Is running docker-compose from python necessary in this case? Seems like some very complicated logic, but it's really solved by having the user just run `./cli.sh` on their own and waiting for a bit

StpMax: Cant agree in this case.  Is much easy make one action (run test file) instead of three (start container and wait it ready, run test file, stop container)

George3d6:
Ok... if that's the only contention point I'm merging this, not enough of an issue to block the workflow here, but @btseytlin in the future please merge these if you are reviewing.

PR #554, opened 2020-07-01
fix #550 login with empty password with auth method change

### LINKED ISSUE #550: Issues connection from Mariadb
Inserting a predictor causes the following error:

```
2020-07-01 00:32:38,356 - INFO - handle new incoming connection
2020-07-01 00:32:38,356 - INFO - New connection [127.0.0.1:43208]
2020-07-01 00:32:38,356 - INFO - send HandshakePacket
2020-07-01 00:32:38,357 - INFO - Sending packet: HandshakePacket
2020-07-01 00:32:38,357 - INFO - Get packet: HandshakeResponsePacket
2020-07-01 00:32:38,357 - INFO - Got packet
2020-07-01 00:32:38,357 - INFO - b'\r\xa2:\x80\x00\x00\x00\x01-\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00mindsdb\x00\x00mindsdb\x00mysql_native_password\x00u\x03_os\x10debian-linux-gnu\x0c_client_name\x08libmysql\x04_pid\x041425\x0f_client_version\x0710.3.22\x0c_server_host\t127.0.0.1\t_platform\x06x86_64'
2020-07-01 00:32:38,357 - INFO - Check auth, user=mindsdb, ssl=False, auth_method=mysql_native_password: empty password
2020-07-01 00:32:38,357 - INFO - Check auth, user=mindsdb, ssl=False, auth_method=mysql_native_password: connecting to database mindsdb
2020-07-01 00:32:38,357 - INFO - Check auth, user=mindsdb: Ok
2020-07-01 00:32:38,357 - INFO - Sending packet: OkPacket
2020-07-01 00:32:38,357 - INFO - Get packet: CommandPacket
2020-07-01 00:32:38,357 - WARNING - Packet with less than 4 bytes in length: b''
2020-07-01 00:32:38,357 - INFO - Session closed by client
```

StpMax:
Issue 550 raised when user try connect with empty password and with 'mysql_native_password' auth method, when server default auth method is different. In this case it always need to make switch auth, what i did not do before (because i already know that  password is empty)
Other changes:
- fix deleting mindsdb database when init clickhouse interface
- add error raise when query to clickhouse returns not 200 code. Before it was not possible realize what some error raised in query (if query not return anything)


tests/integration_tests/common.py
```diff
@@ -17,3 +20,32 @@ def wait_port(port_num, timeout):
         in_use = is_port_in_use(port_num)
 
     return in_use
+
+def prepare_config(config):
```
George3d6, line 24: Does this need to be here, can't it just be replaced by the generate config functionality in the config wizard ? + some changes to the json file that generates ?

tests/integration_tests/common.py
```diff
@@ -17,3 +20,32 @@ def wait_port(port_num, timeout):
...
+                'storage_dir': str(TEMP_DIR.joinpath('predictors/'))
+            }
+        }
+    })
+
+    if not os.path.isdir(config['interface']['datastore']['storage_dir']):
+        os.makedirs(config['interface']['datastore']['storage_dir'])
+    
+    if not os.path.isdir(config['interface']['mindsdb_native']['storage_dir']):
+        os.makedirs(config['interface']['mindsdb_native']['storage_dir'])
```
George3d6, line 45: Mindsdb should create these directories on boot, if it doesn't, that's an issue and should be fixed, if it does, these lines can be removed.

tests/integration_tests/common.py
StpMax: I think it needed here, because this function only for integration tests. This function is add two changes to current config:
- switch off all integrations, except current tested (now hardcoded 'default_mariadb', but this function wiull use in clichhouse tests too)
- change predictors and datasources dirs to some temp dir

tests/integration_tests/common.py
StpMax: issue added
https://github.com/mindsdb/mindsdb/issues/557

PR #559, opened 2020-07-03
Added _max _min and _explain columns

### LINKED ISSUE #547: Allow selecting from `.explain`
Discussed during a call

Gist of it is, assume you are prediciting two columns `foo` and `bar`. Assume `foo` is numerical and `bar` is categorical.

1. All the values (including the predicted value) should come from the `.explain` property of the prediction
2. The values we expose should be:

`foo`
`bar`
`foo_confidence`
`bar_confidence`
`foo_max`
`foo_min`

Where min and max are the first and last value of the `confidence_interval` found in `.explain->foo` and `confidence` is also a property of  `.explain`

In addition to this, maybe expose two more properties:

`foo_explain` and `bar_explain` that are just strings containing a json representation of `.explain` (which is just a python dictionary).

StpMax:
Close #547. Now in result of sql-query will returns for each predicted columns *_explain column, and for Numeric columns *_min and *_max columns.

Additionally fixed very hard to find bug with keeping packets order number

### DIFF
--- a/mindsdb/README.md
+++ b/mindsdb/README.md
@@ -38,7 +38,7 @@ SELECT rental_price FROM mindsdb.rentals_predictor WHERE initial_price=900 and n
 Or, you can select some other data from a similar table (or the same table) and get a bunch of predictions.
 
 ```
-SELECT rental_price FROM mindsdb.rentals_predictor WHERE `$select_data_query`='SELECT * FROM default.home_rentals WHERE days_on_market > 60';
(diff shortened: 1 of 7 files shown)


George3d6:
_confidence was also added, which I feel like it's an important thing to note, as it was a critical bit here, so good to see it's in there.

George3d6:
I can find no nitpick with this, not matter how hard I try, @StpMax please go ahead and merge.

PR #569, opened 2020-07-06
Ci testing

### LINKED ISSUE #521: Compile installers using Travis and Deploy them to s3
The installers should be compiled (for those which require it) via travis and deployed to a public s3 bucket every time we release the `stable` branch.

George3d6:
* Refactored the HTTP tests to work with the new Mindsdb.
* Added tests and linux installe deployment to CI.
* Various bug fixes.

This closes  #521


George3d6:
Ok, tests are failing now, but just because we need to release a new mindsdb_native version, should be reviewed and merged.

mindsdb/__main__.py
```diff
@@ -37,7 +37,7 @@ def close_api_gracefully(p_arr):
 
     print(f'Using configuration file: {config_path}')
     config = Config(config_path)
-
+    print(args.api)
```
btseytlin, line 40: Some leftover print

tests/integration_tests/api/test_http.py
```diff
@@ -0,0 +1,132 @@
...
+import signal
+from random import randint
+
+
+@pytest.fixture(scope="module", autouse=True)
+def ds_name():
+    rand = randint(0,pow(10,12))
+    return f'hr_ds_{rand}'
+
+@pytest.fixture(scope="module", autouse=True)
```
btseytlin, line 17: I think these don't need `autouse` as they are only in effect when you use them anyway

PR #571, opened 2020-07-06
new ssl-cert  for each connection

### LINKED ISSUE #539: Remove hardcoded cert from code
I am informed using a test SSL cert hardcoded inside public source code goes against certain best practices.

StpMax:
close #539
now if field 'certificate_path' not in config.json, then for each connection will be make new temp ssl-cert

### DIFF
--- a/mindsdb/api/mysql/mysql_proxy/mysql_proxy.py
+++ b/mindsdb/api/mysql/mysql_proxy/mysql_proxy.py
@@ -234,12 +234,15 @@ class MysqlProxy(SocketServer.BaseRequestHandler):
             log.info('switch to SSL')
             self.session.is_ssl = True
 
-            use_temp_cert = not isinstance(CERT_PATH, str) or not pathlib.Path(CERT_PATH).is_file()
+            use_temp_cert = isinstance(CERT_PATH, str) is False or len(CERT_PATH) == 0
             cert_path = self.make_temp_cert() if use_temp_cert else CERT_PATH
             if use_temp_cert:
                 log.info(f'use temp ssl cert: {cert_path}')
             else:
                 log.info(f'use user defined ssl cert: {cert_path}')
+                if not pathlib.Path(cert_path).is_file():
+                    log.error(f'File is not SSL cert: {cert_path}')
+                    return False
 
             ssl_socket = ssl.wrap_socket(
                 self.socket,
--- a/mindsdb/api/mysql/mysql_proxy/mysql_proxy.py
+++ b/mindsdb/api/mysql/mysql_proxy/mysql_proxy.py
@@ -17,6 +17,9 @@ import re
 import traceback
 import json
 import atexit
+from datetime import datetime, timedelta
+import tempfile
+import pathlib
 
 from moz_sql_parser import parse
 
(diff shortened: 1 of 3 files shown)


StpMax:
Method to make cert tested on linux and windows.

mindsdb/api/mysql/mysql_proxy/mysql_proxy.py
```diff
@@ -181,12 +233,30 @@ def get_fast_auth_password():
         if handshake_resp.type == 'SSLRequest':
             log.info('switch to SSL')
             self.session.is_ssl = True
+
+            use_temp_cert = isinstance(CERT_PATH, str) is False or len(CERT_PATH) == 0
+            cert_path = self.make_temp_cert() if use_temp_cert else CERT_PATH
```
George3d6, line 238: I don't think this is the way we want to go about thing. Why not generate a cert on install and stick with it ?

Generating a new cert for every ssl request we get seem very wasteful. Am I getting something wrong here ?

mindsdb/api/mysql/mysql_proxy/mysql_proxy.py
```diff
@@ -130,8 +133,57 @@ def isAuthOk(self, user, orig_user, password, orig_password):
             log.error(f'Check auth, user={user}: ERROR')
             log.error(traceback.format_exc())
 
+    def make_temp_cert(self):
```
George3d6, line 136: Can't we use library e.g. the one linked in the issue instead of doing all of this manually? Or is the fine-grain control required to get it looking exactly like the kind of certs MySQL works with?

mindsdb/utilities/wizards.py
```diff
@@ -63,64 +63,10 @@ def cli_config(python_path,pip_path,predictor_dir,datasource_dir,config_dir,use_
...
-zGvTvYkjMFLRtJ13SkRyfiMJkfJcM89czOVu4X/dljiHhGePfdbCUuGs1Gw759a8
-3l7b506sujWQEmuSe6UdOUws+gR82H7kb8n7qxcOa5HXiIE2MRdfHXx8AS0LGPsa
-n0PAUDF7eqI/kYskiWUX
------END CERTIFICATE-----
-                 """)
-
     mysql = _in('Enable MYSQL API ? [Y/N]','Y',use_default)
     if mysql in ['Y','y']:
         config['api']['mysql'] = {
-            "certificate_path": crt_path
```
George3d6, line 122: See previous point, this should remain here and the certificate should be generate in this path on bootup (in __main__.py) if it doesn't exist.

mindsdb/utilities/wizards.py
```diff
@@ -63,64 +110,13 @@ def cli_config(python_path,pip_path,predictor_dir,datasource_dir,config_dir,use_
...
-zGvTvYkjMFLRtJ13SkRyfiMJkfJcM89czOVu4X/dljiHhGePfdbCUuGs1Gw759a8
-3l7b506sujWQEmuSe6UdOUws+gR82H7kb8n7qxcOa5HXiIE2MRdfHXx8AS0LGPsa
-n0PAUDF7eqI/kYskiWUX
------END CERTIFICATE-----
-                 """)
-
     mysql = _in('Enable MYSQL API ? [Y/N]','Y',use_default)
     if mysql in ['Y','y']:
+        crt_path = os.path.join(config_dir, 'cert.pem')
+        make_ssl_cert(crt_path)
```
George3d6: Make it regardless of disabled/enabled, since the user might enable it later.

Altenratively, don't make it in the wizard, make it in the `__main__.py` or in the mysql API (probably a better place to do so anyway).

mindsdb/api/mysql/mysql_proxy/mysql_proxy.py
StpMax: changed to make cert on install

mindsdb/api/mysql/mysql_proxy/mysql_proxy.py
StpMax: That easiest pure-python cert-generation what i can find

mindsdb/utilities/wizards.py
StpMax: cert generation returned to wizard

PR #575, opened 2020-07-08
Dynamic config

George3d6:


mindsdb/interfaces/database/database.py
```diff
@@ -4,37 +4,46 @@
...
         self.config = config
-        self.integration_arr = []
-
-        for db_alias in config['integrations']:
-            if config['integrations'][db_alias]['enabled']:
-                if config['integrations'][db_alias]['type'] == 'clickhouse':
-                    self.integration_arr.append(Clickhouse(config,db_alias))
-                if config['integrations'][db_alias]['type'] == 'mariadb':
-                    self.integration_arr.append(Mariadb(config,db_alias))
+        self._get_integrations()
```
StpMax, line 9: In this case initialization mindsdb database and registration predictors will be on each instantiating of DatabaseWrapper. That means what it will be and on MindsdbNative instantiating too. That can bring racing condition issues. For example:
1. we start mindsdb with --api=mysql,http
2. in http and mysql threads starts instantiating MindsdbNative() in same time. 
3. in thread 'http' executed: 'drop table predictor.xxx', 
4. in thread 'mysql' executed: 'drop table predictor.xxx'
5. in thread 'http' executed: 'create table predictor.xxx', 
6. in thread 'mysql' executed: 'create table predictor.xxx' <- ERROR

mindsdb/utilities/config.py
```diff
@@ -1,41 +1,57 @@
...
+            if file_hash != self._config_hash:
+                self._read()
+                self._config_hash = self._gen_hash()
 
     def __getitem__(self, key):
+        self._set_updated(key)
         return self._config[key]
 
     def get(self, key, default=None):
+        self._set_updated(key)
```
StpMax, line 43: this never will works, because 'key' is always top level keys of config (api, integration, interface, etc) and it never will be in self._config['integrations']

George3d6: My bad that was meant to read just `['integrations']`

mindsdb/interfaces/database/database.py
George3d6: If we do delete predictors, yes, that is an issue. As it stands we get around it by not deleting predictors on initialization

In the future, we could have separate logic in `__main__` for deleting predictors on boot from the databases.

Race conditions in terms of trying to create the same table are not an issue, we can just have them fail silently, unideal but I can't think of another easy way to do this without coupling the **** out of the config and DatabaseWrapper.

George3d6: As in, the alternative is that whenever a config update comes through we run it through the DatabaseWrapper first.

But idk, if you have thoughts on how to do this better speak them out.

George3d6:
Fyi, tests passing but locally, we need the new mindsdb_native release for them to run properly on travis

PR #580, opened 2020-07-09
patch

StpMax:
fixed pack of bugs

### DIFF
--- a/mindsdb/api/mysql/mysql_proxy/datahub/datanodes/mindsdb_datanode.py
+++ b/mindsdb/api/mysql/mysql_proxy/datahub/datanodes/mindsdb_datanode.py
@@ -127,10 +127,7 @@ class MindsDBDataNode(DataNode):
             for key in min_max_keys:
                 row[key + '_min'] = res[i].explanation[key]['confidence_interval'][0]
                 row[key + '_max'] = res[i].explanation[key]['confidence_interval'][-1]
+            row['select_data_query'] = select_data_query
             data.append(row)
-
-        if select_data_query is not None:
-            for row in data:
-                row['select_data_query'] = select_data_query
 
         return data
--- a/mindsdb/interfaces/native/predictor_process.py
+++ b/mindsdb/interfaces/native/predictor_process.py
@@ -38,7 +38,7 @@ class PredictorProcess(ctx.Process):
                 'name': name,
                 'predict_cols': to_predict,
                 'data_analysis': stats
-            }])
+            }], setup = False)
 
         if trx_type == 'predict':
             if isinstance(from_data, dict):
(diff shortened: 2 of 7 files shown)


mindsdb/api/mysql/mysql_proxy/datahub/datanodes/mindsdb_datanode.py
```diff
@@ -127,10 +127,7 @@ def select(self, table, columns=None, where=None, where_data=None, order_by=None
             for key in min_max_keys:
                 row[key + '_min'] = res[i].explanation[key]['confidence_interval'][0]
                 row[key + '_max'] = res[i].explanation[key]['confidence_interval'][-1]
+            row['select_data_query'] = select_data_query
```
George3d6, line 130: Potentially good to bring back the check here for when we support `external_data_source` instead of `select_data_query` ?

mindsdb/interfaces/native/predictor_process.py
```diff
@@ -38,7 +38,7 @@ def run(self):
                 'name': name,
                 'predict_cols': to_predict,
                 'data_analysis': stats
-            }])
+            }], setup = False)
```
George3d6, line 41: I believe this needs to be set to `True` here for now, I'm afraid.

As in, assuming a new connection is added, we need to register the prediction to it, unless the connections are re-established every time, we can't (at the moment) make that happen. Am I wrong in assuming this ?

StpMax: This block should not executed if connection added. It executed just when predictor trained. And there was error, what was fixed by adding 'setup = False':
1. user start learn predictor by send to db statement 'insert into mindsdb.predictors ....'
2. when learning finished, then this 'DatabaseWrapper.register_predictors' starts. 
3. before 'setup' parameter was added, it each time execute '_setup_integrations' what follows to 'integration.setup' where executed 'drop database if exists mindsdb', what follow to db lock.

mindsdb/api/mysql/mysql_proxy/datahub/datanodes/mindsdb_datanode.py
StpMax: `external_data_source` instead of `select_data_query` ? i think we should use both, or not?

If talk about `select_data_query` than it will be better do here so here:
```
    if came_from is not None:
        for row in data:
            row['select_data_query'] = select_data_query
```
but: above is method `getTableColumns`, and it returns for each predictor table column `select_data_query` in list. That not  very good, because this column will be return in plain sql query to mindsdb too. That need to fix in future, but for now we must set `select_data_query` in each row.

PR #589, opened 2020-07-14
Tests update

StpMax:
All three tests (clickhouse, mariadb, http) now passing in Travis.
Notes:
1. looks like using 'fuser' brokes travis
2. Sometime when http_test run is error raised: address already in use. I dont know why, is no place before where http-interface started. So i put in travis `ps -aux | grep $(lsof -t -i:47334)` and print(...) to http/start.py to try figure out what happened when it will be again.


tests/integration_tests/flows/test_clickhouse.py
```diff
@@ -44,92 +56,226 @@ def query_ch(query, database='default'):
...
+        if TEST_PREDICTOR_NAME in models:
+            cls.mdb.delete_model(TEST_PREDICTOR_NAME)
+
+        query('create database if not exists test')
+        test_tables = query('show tables from test')
+        test_tables = [x['name'] for x in test_tables]
+        if TEST_DATA_TABLE not in test_tables:
+            print('creating test data table...')
+            query(f'''
+                CREATE TABLE test.{TEST_DATA_TABLE} (
```
George3d6, line 84: Can we switch back t creating the table over the network:

```
CREATE TABLE {ds_name} (number_of_rooms String, number_of_bathrooms String, sqft Int64, location String, days_on_market Int64, initial_price Int64, neighborhood String, rental_price Float64)  ENGINE=URL('https://raw.githubusercontent.com/mindsdb/mindsdb-examples/master/benchmarks/home_rentals/dataset/train.csv', CSVWithNames)
```

makes things much simpler to read

tests/integration_tests/flows/test_clickhouse.py
```diff
@@ -44,92 +56,226 @@ def query_ch(query, database='default'):
...
-
-        #query_ch('DROP DATABASE mindsdb')
-        query_ch(f"""
-        CREATE TABLE {ds_name} (number_of_rooms String, number_of_bathrooms String, sqft Int64, location String, days_on_market Int64, initial_price Int64, neighborhood String, rental_price Float64)  ENGINE=URL('https://raw.githubusercontent.com/mindsdb/mindsdb-examples/master/benchmarks/home_rentals/dataset/train.csv', CSVWithNames)
+        cls.mdb = MindsdbNative(config)
+
+        models = cls.mdb.get_models()
+        models = [x['name'] for x in models]
+        if TEST_PREDICTOR_NAME in models:
+            cls.mdb.delete_model(TEST_PREDICTOR_NAME)
```
George3d6, line 76: can be replaced with `MindsdbNative(config).delete_model(TEST_PREDICTOR_NAME)`

tests/integration_tests/flows/test_mariadb.py
```diff
@@ -299,63 +302,50 @@ def test_7_delete_predictor_by_delete_statement(self):
...
             print(f'Failed by timeout. MariaDB started={maria_ready}, MindsDB started={api_ready}')
             raise Exception()
         unittest.main(failfast=True)
-        print('Tests passed !')
+        print('Tests passed!')
     except Exception as e:
-        print('Tests Failed !')
-        print(e)
-    print('done')
+        print(f'Tests Failed!\n{e}')
```
George3d6, line 351: Could we move this logic to the setup method of the class, similarly to how we do it for clickhouse ?

I know it's not much of an advantage, but a bit cleaner and will allow for running with pytest later if we end up switching.

tests/integration_tests/flows/test_clickhouse.py
StpMax: agree, file small, we can reload it each time

tests/integration_tests/flows/test_clickhouse.py
StpMax: .delete_model raise error if predictor not exists, so need check it before try to delete

tests/integration_tests/flows/test_mariadb.py
StpMax: ok, will move

tests/integration_tests/flows/test_clickhouse.py
StpMax: But may be better to keep MergeTree-engine table for tests.

StpMax: after some experiments i conclude what ENGINE=URL('', CSVWithNames) very buggy:  you can specify any types of columns and table will be created. But you can access only rows what can by cast to that types, other will be missed, or it will be error if you try to access them. Example:
1. create table `CREATE TABLE test.xxx3 (number_of_rooms String, number_of_bathrooms String, sqft String, location String,days_on_market Int16,initial_price Int32,neighborhood String,rental_price Int32) ENGINE=URL('https://s3.eu-west-2.amazonaws.com/mindsdb-example-data/home_rentals.csv', CSVWithNames)` 
number_of_rooms, number_of_bathrooms, sqft strings, overwise will be error in any access.
2.`select * from test.xxx3 order by rental_price limit 1;` - error by timeout
3. `select * from test.xxx3 limit 1;` works, but  `select * from test.xxx3 limit 10;` fail by `Expected end of line.`
so i keep old table creation

tests/integration_tests/flows/test_clickhouse.py
George3d6: fair enough, still no point in attaching the `mdb` object to `cls`.

tests/integration_tests/flows/test_clickhouse.py
George3d6: Fair enough. I can't say I like it but I agree on the fact that engine switching is easier this way and if you think the URL impl is buggy since it breaks only when selecting rather than at creation... than, yeah, I see your point.

tests/integration_tests/flows/test_clickhouse.py
StpMax: mdb object used in tests to check what test predictor really created/removed

PR #594, opened 2020-07-15
Capture information about

### LINKED ISSUE #555: Capture information about the host database on receiving queries
The Mysql mime proxy should capture information about the sender databases when it receives a query (e.g. via the connect engine)... at least if at all possible. Things like:

What's the IP from which the query was sent ?
What's the port the database is running on (unlikely we can get this) ?
What's the type of the database (again, maybe not, since @StpMax you were mentioning `clickhouse` send `mariadb` as the name *and* `mariadb` sends `mysql`...) ?
... basically, any identifying information we should try to capture.

Alternatively, since this *might* be hard or impossible, whenever we create a `predictors` and `commands` table in the databases, we should add some sort of magic value that  is a hash which informs us of the database's identity.

@StpMax let me know what are your thoughts on this ?

Obviously easy to do when we only have 2 integrations (clickhouse and mysql) but I'm thinking in the scenario where we could integrate with 3 or 4 clickhouse databases but want to know e.g. which one of them to query data from when a learn/predict query comes through.

StpMax:
Close #555
Capture information about query source using db username

### DIFF
--- a/mindsdb/api/mysql/mysql_proxy/classes/sql_query.py
+++ b/mindsdb/api/mysql/mysql_proxy/classes/sql_query.py
@@ -60,12 +60,14 @@ class SQLQuery():
 
         return dict(zip(columns, values))
 
-    def __init__(self, sql, default_dn=None):
+    def __init__(self, sql, default_dn=None, session=None):
         # parse
-
+        self.session = session
         self.default_datanode = None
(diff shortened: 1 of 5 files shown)


George3d6:
Unsure why the tests are failing, maybe just has to do with the server restart timing ? restarted them to see if they will succeede this time.

PR #606, opened 2020-07-22
Release 2.1.0

StpMax:
features:
- config migration and validity check at start
- if no specify in config 'certificate_path' or set it to empty string, then on start will be created temp ssl cert.

fixes:
- mindsdb ignored storage_dir for mindsdb_native passed in --config arg and used it from default config instead.
- escape predictors name when create related tables in databases. 

tests:
- added tests for 'external_datasource'


mindsdb/utilities/config.py
```diff
@@ -16,13 +18,58 @@ def __init__(self, config_path):
...
+            if 'datasources' in config['api']['mysql']:
+                del config['api']['mysql']['datasources']
+            config['config_version'] = '1.1'
+            return config
+
+        migrations = {
+            '1.0': m1_0
+        }
+
+        current_version = self._parse_version(self._config['config_version'])
```
maximlopin, line 36: there is packaging,version.Version class that takes a version string in constructor and supports `<=, ==, >=`operators, it might be useful here

StpMax: yes, 'packaging' good for version comparison. But adding additional dependency for one function can be a bit overhead. So i made my own :)

PR #621, opened 2020-07-28
MySQL integration

StpMax:
Integration with MySQL


mindsdb/api/mysql/mysql_proxy/classes/sql_query.py
```diff
@@ -65,6 +65,12 @@ def parse_insert(sql):
         values = p['literal'] if isinstance(p, dict) else p
         if isinstance(values, list) is False:
             values = [values]
+        for i, v in enumerate(values):
+            if isinstance(v, dict) and 'literal' in v:
```
maximlopin, line 69: Is `isinstance(v, dict) and 'literal' not in v` case supposed to be ignored?
Making it explicit will remove such doubts for anyone reading the code:
```
for i, v in enumerate(values):
            if isinstance(v, dict):
                if 'literal' in v:
                     v = v['literal']
                else:
                    pass
```

mindsdb/api/mysql/mysql_proxy/classes/sql_query.py
```diff
@@ -65,6 +65,12 @@ def parse_insert(sql):
...
+                v = v['literal']
+            elif v == 'null':
+                v = None
+            values[i] = v
```
maximlopin, line 73: in theory `v` can be undeclared because the `else-if` statement above doesn't have `else`

mindsdb/integrations/mysql/mysql.py
```diff
@@ -0,0 +1,149 @@
...
+            except Exception:
+                print(f'Error: cant convert type {col_subtype} of column {name} to mysql type')
+
+        return column_declaration
+
+    def _escape_table_name(self, name):
+        return '`' + name.replace('`', '``') + '`'
+
+    def _query(self, query):
+        con = mysql.connector.connect(host=self.config['integrations'][self.name]['host'], port=self.config['integrations'][self.name]['port'], user=self.config['integrations'][self.name]['user'], password=self.config['integrations'][self.name]['password'])
```
maximlopin, line 45: much easier to read:
```
con = mysql.connector.connect(
    host=self.config['integrations'][self.name]['host'],
    port=self.config['integrations'][self.name]['port'],
    user=self.config['integrations'][self.name]['user'],
    password=self.config['integrations'][self.name]['password']
)
```

mindsdb/integrations/mysql/mysql.py
```diff
@@ -0,0 +1,149 @@
...
+    def setup(self):
+        self._query('DROP DATABASE IF EXISTS mindsdb')
+
+        self._query('CREATE DATABASE IF NOT EXISTS mindsdb')
+
+        connect = self._get_connect_string('predictors')
+
+        q = f"""
+                CREATE TABLE IF NOT EXISTS mindsdb.predictors
+                (name VARCHAR(500),
```
maximlopin, line 81: I'm seeing this `"VARCHAR(500)"` in multiple files, it might make sense to declare a variable in constants like `LONG_VARCHAR = "VARCHAR(500)"` and use it everywhere else, it will be easier to change this magic value `500` in future

mindsdb/interfaces/datastore/datastore.py
```diff
@@ -86,7 +86,7 @@ def save_datasource(self, name, source_type, source, file_path=None):
                 ,'kwargs': {}
             }
         elif source_type == 'clickhouse':
-            user = self.config['integrations']['default_clickhouse']['user']
+            user = self.config['integrations']['default_clickhouse']['user']            # FIXME
```
maximlopin, line 89: `# FIXME`
unclear what should be fixed :))

George3d6:
@StpMax I like the boilerplate removal you did for testing, I'll leave the review of this up to @maximlopin since he's already on it :+1:

tests/integration_tests/flows/test_mysql.py
```diff
@@ -0,0 +1,283 @@
...
+        host=config['integrations']['default_mysql']['host'],
+        port=config['integrations']['default_mysql']['port'],
+        user=config['integrations']['default_mysql']['user'],
+        passwd=config['integrations']['default_mysql']['password'],
+        db='mindsdb'
+    )
+
+    cur = con.cursor(dictionary=as_dict)
+    cur.execute(q)
+    res = True
```
maximlopin, line 38: this is strange because I think all tests expect output of `query()` to be an iterable

mindsdb/api/mysql/mysql_proxy/classes/sql_query.py
StpMax: yeah, may be it will be easier to read

mindsdb/api/mysql/mysql_proxy/classes/sql_query.py
StpMax: If `v` is not 'null' and not dict then it should not be changed. By default parser returns strings, so i think it correct here

mindsdb/integrations/mysql/mysql.py
StpMax: done

mindsdb/interfaces/datastore/datastore.py
StpMax: good notice :)

tests/integration_tests/flows/test_mysql.py
StpMax: `create database ...` and `create table ...` does not returns iterable.

mindsdb/integrations/mysql/mysql.py
StpMax: it used in 'mariadb' and 'mysql'. I will change this when whill change this files in next time

mindsdb/api/mysql/mysql_proxy/classes/sql_query.py
maximlopin: right, I didn't see the for loop

PR #632, opened 2020-08-05
Db improvements

### LINKED ISSUE #631: As a API consumer I must have all integration details in one single endpoint
Hello, 
it's possible to have all integration details in one endpoint, for example
```
GET http://0.0.0.0:47334/config/integrations

Response:
[
{
  "conection_name": "conection_name_1"
  "enabled": true,
  "host": "localhost",
  "password": "mypass",
  "port": 3306,
  "type": "mariadb",
  "user": "root"
},
{
  "conection_name": "conection_name_2"
  "enabled": true,
  "host": "HTTP://host.com/midb",
  "password": "secret",
  "port": 3306,
  "type": "mariadb",
  "user": "root"
},
{
  "conection_name": "conection_name_3"
  "enabled": true,
  "host": "HTTP://host.com/midb",
  "password": "123",
  "port": 3306,
  "type": "mariadb",
  "user": "root"
},
]

```

George3d6:
1. Added `/config/all_integrations` endpoint that lists all integrations with the full data associated with all of them in a dict keyed on the integration ids

2. Added `/integrations/<name>/check` (where `<name>` is the integration's id) that allows you to check whether an integration is functional (i.e. if mindsdb can connect to the database)

3. Added the ability to add database integration based datasource to the `/datasources/<name> -- PUT` endpoiont:

Simply call the endpoint and replace `<name>` with the name of our datasource then give two parameters in the json post body:

`query` - the sql query to get the actual data
`integration_id` - the id of the database integration

4. To see what database integration a given datasource comes from, simply check the `source_type` parameter of the datasource, if it's not `file` or `url` it means it comes from a database integration and `source_type` is the actual id/name

5. Added `date_last_updated` and `database_name` parameter support to the database integrations under `/config`

6. Added tests for all new features

Should resolve #631 and #630

### DIFF
--- a/mindsdb/api/http/namespaces/config.py
+++ b/mindsdb/api/http/namespaces/config.py
@@ -8,10 +8,17 @@ from mindsdb.interfaces.database.database import DatabaseWrapper
 
 @ns_conf.route('/integrations')
 @ns_conf.param('name', 'List all database integration')
-class Integration(Resource):
+class ListIntegration(Resource):
     @ns_conf.doc('get_integrations')
     def get(self):
         return {'integrations': [k for k in ca.config_obj['integrations']]}
+
+@ns_conf.route('/all_integrations')
+@ns_conf.param('name', 'List all database integration')
+class AllIntegration(Resource):
+    @ns_conf.doc('get_all_integrations')
+    def get(self):
+        return ca.config_obj['integrations']
 
 @ns_conf.route('/integrations/<name>')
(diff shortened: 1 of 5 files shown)


StpMax:
I didn't run this branch, but code looks good. Only question: why need 'database_name'? just alias to integration[KEY]?

alejandrovillegas:
@StpMax database_name it's only a name that we will show in the scout

PR #643, opened 2020-08-13
Read passwords from config as strings

StpMax:
If user specified password in config as int, it followed to errors when connect to DB. Now all passwords will reading as strings, empty or  = null will read as empty string.


ZoranPandovski:
**CLA Assistant Lite** All Contributors have signed the CLA.

PR #646, opened 2020-08-14
Deleting datastore dir on learing error

### LINKED ISSUE #609: Datastore dir created when error happens
The user can include wrong parameters or have a typo in INSERT OR SELECT query which will result in an error but the directory inside datastore will be created, so the second time user will get `[Errno 17] File exists: '/mindsdb-examples/tescl/datastore/airq_predictor'`.

StpMax:
Fixes #609

1. Datasource dir will be removed if error raised.
2. If trained columns not in datasource columns list, error for user will be raised
3. fixed 'use db' query
4. changed clickhouse array type to text, because CH cant get array from mysql.


ZoranPandovski:
**CLA Assistant Lite** All Contributors have signed the CLA.

PR #677, opened 2020-09-08
Downloading GUI from S3

StpMax:


mindsdbadmin:

mindsdb/api/http/initialize.py
```diff
@@ -16,8 +24,87 @@ class Swagger_Api(Api):
...
+    ACCESS_KEY = ""
+    SECRET_KEY = ""
+
+    try:
+        s3 = boto3.client('s3', aws_access_key_id=ACCESS_KEY, aws_secret_access_key=SECRET_KEY)
+        s3.download_file('mindsdb-web-builds', 'index.html', str(static_path.joinpath('index.html')))
+        css_zip_path = str(static_path.joinpath('css.zip'))
+        js_zip_path = str(static_path.joinpath('js.zip'))
+        s3.download_file('mindsdb-web-builds', 'css-V' + gui_version_lv.vstring + '.zip', css_zip_path)
+        s3.download_file('mindsdb-web-builds', 'js-V' + gui_version_lv.vstring + '.zip', js_zip_path)
```
George3d6, line 82: Don't include boto3 for the sake of downloading some files, since these will be public anyway.

It's an extra dependency that is to be avoided, but specifically, in the case of boto3, it's an extra dependency that can interact badly with a lot of other things the user might have in his env since it's sometimes problematic to installed and required by many packages.

mindsdb/api/http/initialize.py
```diff
@@ -16,8 +24,87 @@ class Swagger_Api(Api):
     def specs_url(self):
         return url_for(self.endpoint("specs"), _external=False)
 
+
+def initialize_static():
+    this_file_path = os.path.abspath(inspect.getfile(inspect.currentframe()))
+    static_path = Path(this_file_path).parent.joinpath('static/')
```
George3d6, line 30: In the future, remember to move this to the `var` dir where we store predictors, datasources ... etc.

PR #688, opened 2020-09-10
Fix paths

StpMax:
Removed creating paths from HTTP start.
Predictors files will be stored in right directory.


mindsdbadmin:

George3d6 (changes requested):
A welcome refactor, the fact that we were still using those variables in `CONFIG` was certainly wrong, not sure why that was even working.

Not 100% sure about using env variable to pass those paths around though. Couldn't we just source them directly from the config ? If there's issues with that though, feel free to merge as is.

PR #722, opened 2020-09-25
fixes for mongo 4.4

StpMax:
It should make possible connect to MongoDB v4.4.

### DIFF
--- a/mindsdb/api/mongo/op_msg_responders/add_shard.py
+++ b/mindsdb/api/mongo/op_msg_responders/add_shard.py
@@ -1,0 +1,40 @@
+from mindsdb.api.mongo.classes import Responder
+import mindsdb.api.mongo.functions as helpers
+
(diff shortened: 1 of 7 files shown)


mindsdbadmin:

mindsdb/api/mongo/op_msg_responders/add_shard.py
```diff
@@ -0,0 +1,40 @@
...
+#     },
+#     "$clusterTime": {
+#         "clusterTime": 6876008835761307650,
+#         "signature": {
+#             "hash": 00:00:00:00:00:00:00:00:00:00:00:00:00:00:00:00:00:00:00:00,
+#             "keyId": 0
+#         }
+#     },
+#     "operationTime": 6876008835761307649
+# }
```
George3d6, line 29: Is it worth it to keep these large bits of logic commented out ?

George3d6 (approved):
Not my area of expertise, so I can't opine on it much, if the tests pass go ahead and merge :)

mindsdb/api/mongo/op_msg_responders/add_shard.py
StpMax: yes, i think we should keep it, for a time at least. Im not sure all problems with v4.4 gone. That not easy get dump of real answer. So if we will have problems with this answer - will be much easier find reason.

PR #746, opened 2020-10-01
python version check

shobhitsundriyal:
used exit() if python requirement is not met


George3d6:
Already handled in another PR, so closing this, sorry , will try to organize better (assign issues to people that are interested) as to avoid conflicting PRs in the future :(

PR #792, opened 2020-10-12
latest/fixed version installers (windows)

maximlopin:


distributions/windows/install.py
```diff
@@ -11,11 +11,17 @@
 GET_PIP_URL = 'https://bootstrap.pypa.io/get-pip.py'
 VC_REDIST_URL = 'https://aka.ms/vs/16/release/vc_redist.x64.exe'
 
+# must be replaced in another script
+NAME = '$name'
+VERSION = '$version'
+
+assert NAME != '$' + 'name'
+assert VERSION != '$' + 'version'
```
George3d6, line 19: Why are these here ?

distributions/windows/install.py
```diff
@@ -11,11 +11,17 @@
...
+
 if len(sys.argv) < 2:
-    INSTALL_DIR = 'C:/Program Files/MindsDB'
+    INSTALL_DIR = os.path.join('C:\PROGRA~1', NAME)
```
George3d6, line 22: Is `PROGRA~1` the correct location ? looks like a weirdly named dir

George3d6 (changes requested):

distributions/windows/install.py
maximlopin: it resolves to Program Files, I'm using it because otherwise there is a weird error because of spaces in the dir name

distributions/windows/install.py
George3d6: Ok, so it's basically a feature for when testing a new version, got it.

PR #841, opened 2020-10-20
Static file fix(?)

George3d6:
Fix to static file path to be `''` instead of `'/'`. Apparently `'/'` sometimes doesn't work (https://github.com/pallets/flask/issues/3134)... though flask says it should. Maybe it's only an issue with some versions or environments, but if I encountered it chances are other people might, so we might as well use `''` if there is not harm @StpMax , thoughts ?

Also added a few files to the `.gitignore` with this ocassion.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -22,3 +22,7 @@ distributions/docker-compose/*/storage
 venv/*
 root
 static
+tmp
+predictors
+datasources
+.pytest_cache
--- a/.gitignore
+++ b/.gitignore
@@ -20,3 +20,5 @@ tests/docker/mongodb/storage_*
 .vscode
 distributions/docker-compose/*/storage
 venv/*
+root
+static
--- a/mindsdb/api/http/initialize.py
+++ b/mindsdb/api/http/initialize.py
@@ -184,6 +184,10 @@ def initialize_flask(config):
         static_folder=config.paths['static']
     )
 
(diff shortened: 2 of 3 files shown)


.gitignore
```diff
@@ -20,3 +20,9 @@ tests/docker/mongodb/storage_*
 .vscode
 distributions/docker-compose/*/storage
 venv/*
+root
+static
+tmp
+predictors
+datasources
```
StpMax, line 27: why you add this? you set in config file {storage_path} to mindsdb path?

George3d6: I usually auto-generate my configuration files and it generates them there (sometimes).

Honestly I'm not sure why all of those are there (e.g. not sure where `root`) comes from, but they were lying around, I assumed they were auto-generate by some script or another, and I didn't want to get them into the project.

Could adding any of them to gitignore cause issues ?

StpMax: No, it should not be any issues with adding this lines to gitignore. You can add it. I just surprised, how you get this folders in project root. By default it should be in `{root}/var` https://github.com/mindsdb/mindsdb/blob/staging/mindsdb/utilities/fs.py#L23
I make fresh install in venv, run mindsdb without `--config` option, and get folder `{root}/var`.

StpMax (approved):

PR #858, opened 2020-10-22
Caching data analysis forever

George3d6:
Removed the caching refresh limit of 10 seconds for data analysis, due to datasource versioning we can just cache this forever.

### DIFF
--- a/mindsdb/api/http/namespaces/datasource.py
+++ b/mindsdb/api/http/namespaces/datasource.py
@@ -152,6 +152,8 @@ class Analyze(Resource):
         if name in ds_analysis:
             if ds_analysis[name] is None:
                 return {'status': 'analyzing'}, 200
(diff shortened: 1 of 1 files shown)


StpMax:
What it should fix?
I see two issues with this change:
1. if remove ds and upload new with same name - then you will see analysis of first ds.
2. if it is db-ds, then if you change data in database, you will still see analysis of old data.

George3d6:
@StpMax the fact that it takes way too long to run analysis, basically the cache doesn't work with a 10 second timeout.
I think we can think about refreshing the analysis later, for now we have to take the practical route and cache the analysis.

StpMax:
The main idea was not to block the user's request. Then client go to analysis, request just start analysis, and each 3 seconds ping it again, and get analysis right after it finished. But if you sure it not working, then sure, change it.

StpMax (approved):

PR #866, opened 2020-10-24
fixed address already in use when running mindsdb

### LINKED ISSUE #823: Fix address already in use when running mindsdb
We need help with find and fix bug:
When you run mindsdb in console, for each interface (specified during --api argument) it needs to run the separate process. When you interrupt the main process by ctrl-c, all other processes should stop too. But sometimes it doesn't happen. We don't know why, maybe it depends on OS, or terminal, or something else. You can check it easy:
1. for your convenience create and activate venv (commands below will be slightly different on windows):
```
python3 -m venv venv
source venv/bin/activate
```
2. install mindsdb:
```
pip3 install mindsdb
```
3. run mindsdb as:
```
python3 -m mindsdb
```
by default will be stated 2 interfaces: HTTP on port 47334 and MySQL on port 47335.
4. Interrupt it by ctrl-c
5. Run mindsdb again. If you see the error 'address already in use', then well, you caught the error :)

All processes start and close in __main__.py https://github.com/mindsdb/mindsdb/blob/staging/mindsdb/__main__.py
We assume on 'ctrl-c' interrupt should be executed in close_api_gracefully function https://github.com/mindsdb/mindsdb/blob/staging/mindsdb/__main__.py#L23

yash2189:
Fixes #823 

  - Previously, there was a bug that when interrupting the main process by ctrl-c, all other processes should have stopped too. 
    But sometimes that didn't happen in some cases.
  - In order to solve this issue , this PR is an improvement to the existing code that adds `try - except` to  `close_api_gracefully` 
    function so that the graceful exit can take place.

Changes can be found [here](https://github.com/yash2189/mindsdb/blob/fix%23823/mindsdb/__main__.py)
mindsdb

### DIFF
--- a/mindsdb/__main__.py
+++ b/mindsdb/__main__.py
@@ -27,13 +27,15 @@ from mindsdb.utilities.log import initialize_log
 
 
 def close_api_gracefully(apis):
-    for api in apis.values():
-        process = api['process']
-        sys.stdout.flush()
-        process.terminate()
-        process.join()
-        sys.stdout.flush()
-
+    try: 
+        for api in apis.values():
+            process = api['process']
+            sys.stdout.flush()
+            process.terminate()
+            process.join()
+            sys.stdout.flush()
+    except KeyboardInterrupt:
+        sys.exit(0)
 
 if __name__ == '__main__':
     version_error_msg = """


yash2189:
@ZoranPandovski I think we may even use a signal handler alternatively, although I have used `try-except` here.

ZoranPandovski:
recheckcla

mindsdbadmin:

George3d6:
@yash2189 this might help a lot, thank you :)

yash2189:
@George3d6 I have enrolled to be a beta tester for MindsDB and sent the mail for getting added into the slack channel. Could you please tell me whether I need to do anything else?

PR #885, opened 2020-10-29
fix datetime types cast

StpMax:
small fix to dt types, to show it correct in mysql api

### DIFF
--- a/mindsdb/utilities/functions.py
+++ b/mindsdb/utilities/functions.py
@@ -17,9 +17,9 @@ def cast_row_types(row, field_types):
     keys = [x for x in row.keys() if x in field_types]
     for key in keys:
         t = field_types[key]
-        if t == 'Timestamp' and (isinstance(row[key], int) or isinstance(row[key], float)):
+        if t == 'Timestamp' and isinstance(row[key], (int, float)):
             timestamp = datetime.datetime.utcfromtimestamp(row[key])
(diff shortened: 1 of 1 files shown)


mindsdb/utilities/functions.py
```diff
@@ -17,9 +17,9 @@ def cast_row_types(row, field_types):
     keys = [x for x in row.keys() if x in field_types]
     for key in keys:
         t = field_types[key]
-        if t == 'Timestamp' and row[key] is not None:
+        if t == 'Timestamp' and (isinstance(row[key], int) or isinstance(row[key], float)):
```
maximlopin, line 20: `isinstance(row[key], int) or isinstance(row[key], float)` simplifies to `isinstance(row[key], (int, float))`

mindsdb/utilities/functions.py
```diff
@@ -17,9 +17,9 @@ def cast_row_types(row, field_types):
...
             timestamp = datetime.datetime.utcfromtimestamp(row[key])
             row[key] = timestamp.strftime('%Y-%m-%d %H:%M:%S')
-        elif t == 'Date':
+        elif t == 'Date' and (isinstance(row[key], int) or isinstance(row[key], float)):
```
maximlopin, line 23: isinstance(row[key], int) or isinstance(row[key], float) simplifies to isinstance(row[key], (int, float))

PR #903, opened 2020-11-02
Tests using remote databases

StpMax:
Integration tests can be run in two modes, defined by env variable USE_EXTERNAL_DB_SERVER:
1. if USE_EXTERNAL_DB_SERVER="0", than on run test will be runned docker container with target databasde.
2. if USE_EXTERNAL_DB_SERVER="1", then ssh-tunnels to test database server will be up. For that should exists:  
* `.mindsdb_credentials.json` file in home dir.
* `~/.ssh/db_machine` with key for db machine.


George3d6 (changes requested):
I see that the `run_docker_db` logic is still in, together with the dockerfile. Any reason these shouldn't be removed @StpMax  ?

The idea here would be that everybody has access to remote testing machines, so docker running locally should no longer be necessary.

George3d6 (approved):

PR #937, opened 2020-11-16
Check GUI is served

StpMax:
I hope GUI downloading will be async soon, so i put test to the end and set timeout.


ilia-tsyplenkov (approved):

ilia-tsyplenkov:
LGTM

PR #939, opened 2020-11-16
Install mindsdb from pypi in deploy step

### LINKED ISSUE #804: Pip install after deploy
As part of our github action testing suite we need to add a step/script that runs *after* the deploy to pypi and tries to install the latest `mindsdb` version deployed there.

This is because, in certain scenarios, the pypi version install can be different from the way we install it in when testing and can thus still fail.

This should only happen when merging to stable, i.e. when a release to pypi is made.

ilia-tsyplenkov:
Fixes #804 

Release image (which created in a `deploy` step) needs to contain MindsDB version installed from PyPi repository (not from local source files) In this case we ensure that release image and PyPi have same version.

### DIFF
--- a/.github/workflows/mindsdb.yml
+++ b/.github/workflows/mindsdb.yml
@@ -68,6 +68,11 @@ jobs:
         CHECK_FOR_UPDATES: False
         DB_MACHINE_KEY: ${{secrets.DB_MACHINE_KEY}}
         DATABASE_CREDENTIALS: ${{secrets.DATABASE_CREDENTIALS}}
+    - name: Install latest version from pypi (just testing)
+      run: |
+        sleep 20
+        pip install mindsdb
+        pip show mindsdb
   deploy_windows_installer:
     runs-on: windows-latest
     needs: test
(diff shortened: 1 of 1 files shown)


.github/workflows/mindsdb.yml
```diff
@@ -68,6 +68,11 @@ jobs:
         CHECK_FOR_UPDATES: False
         DB_MACHINE_KEY: ${{secrets.DB_MACHINE_KEY}}
         DATABASE_CREDENTIALS: ${{secrets.DATABASE_CREDENTIALS}}
+    - name: Install latest version from pypi (just testing)
+      run: |
+        sleep 20
+        pip install mindsdb
+        pip show mindsdb
```
George3d6, line 75: Was this for testing when you made the PR, if so please remove before merging. all else looks fine.

George3d6 (changes requested):

.github/workflows/mindsdb.yml
ilia-tsyplenkov: Done

PR #950, opened 2020-11-18
Async download Scout on start

### LINKED ISSUE #940: Async download Scout on start
When mindsdb-http api starts, it checks new version of gui is available. If it available, mindsdb starts download it. Issue here - is mindsdb-http server is not start until gui-downloading finishes. Because of this launch time of http-server can be long.
To fix it, need run initialize_static (https://github.com/mindsdb/mindsdb/blob/staging/mindsdb/api/http/start.py#L24) function in separate process/thread.
Also, it will be good to make concurrent download of GUI files here: https://github.com/mindsdb/mindsdb/blob/staging/mindsdb/api/http/initialize.py#L141

ilia-tsyplenkov:
Fixes #940

### DIFF
--- a/mindsdb/api/http/initialize.py
+++ b/mindsdb/api/http/initialize.py
@@ -8,6 +8,7 @@ from zipfile import ZipFile
 from pathlib import Path
 import logging
 import traceback
+import concurrent.futures
 
 from flask import Flask, url_for
 from flask_restx import Api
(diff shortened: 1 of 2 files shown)


mindsdb/api/http/initialize.py
```diff
@@ -138,11 +139,23 @@ def initialize_static(config):
...
+            except Exception as e:
+                return e
+            return None
+
+        # to make downloading faster download each resource in a separate thread
+        with concurrent.futures.ThreadPoolExecutor(max_workers=5) as executor:
+            future_to_url = {executor.submit(get_resources, r): r for r in resources}
+            for future in concurrent.futures.as_completed(future_to_url):
+                res = future.result()
+                if res:
```
George3d6, line 157: Check `if res is not None`, usually we prefer to be explicit in check unless we are using an actual `True` or `False` boolean as the condition.

George3d6 (changes requested):
overall looks good, other than that minor comment.

PR #971, opened 2020-11-25
OSX installer prototype with pyinstaller base

### LINKED ISSUE #962: OSX Installer using pyinstaller

ilia-tsyplenkov:
Fixes #962 

This is OSX installer prototype.
Since we can't to downgrade/upgrade OS python the main idea to install required (for mindsdb correct work) python version in a custom directory (configure, compile and install). Next steps are similar with windows installer (install torch in specified version range, create a primitive mindsdb launcher).
The installer also checks that require programs/packages/utils installed and stops work with related message otherwise.


George3d6 (changes requested):
Building the installer should also be added to the CI (`.github/workflows/mindsdb.yml` and `release.yml`)

ilia-tsyplenkov:
> Building the installer should also be added to the CI (`.github/workflows/mindsdb.yml` and `release.yml`)

Sure, but I decided that general idea needs to be discussed first. Going to add required changes.

.github/workflows/mindsdb.yml
```diff
@@ -172,10 +204,6 @@ jobs:
         SOURCE_DIR: 'distributions/linux/dist'
         DEST_DIR: 'mindsdb-installer/linux'
 
-    - name: Build OSX sh installer
-      run: |
-        python distributions/osx/rename_installers.py beta
```
George3d6, line 177: Don't remove this, we can still keep the script installer.

George3d6 (changes requested):

.github/workflows/mindsdb.yml
ilia-tsyplenkov: Done

PR #989, opened 2020-12-01
Windows installer pip version == 20.2.4

maximlopin:


George3d6:
Don't think it;'s needed here, since python 3.7 should already come with an older pypi version.

PR #1008, opened 2020-12-14
added telemetry endpoint

g3gayatree:
Fixes Issue 993

  - Added an endpoint to enable/disable telemetry in the form of `config/telemetry/<true or false>` + associated http tests
  - functionality as below:
image
image

mindsdb


gayatrishandar:
hi @George3d6, just wanted know if you/whoever assigned could take a look at this PR + associated question on the ticket?

George3d6:
Looks ok, we still have to add the actual telemtry-disabling to the endpoint though (and that requires some change in mindsdb_native)... but at least the endpoint is there now and the fix to the readme is also welcome. Cheers :)

PR #1009, opened 2020-12-15
added predict from query endpoint

adzyu100:
Fixes [Issue 808](https://github.com/mindsdb/mindsdb/issues/808)

  - Added a predict endpoint which takes in a SQL query and an integration id, and returns the prediction for the query.
  - Specifically, it creates a datasource from the query and integration id, feeds it to native's predict, and returns the results.
  - Url of the form `/api/predictors/:name/predict_query`, where `name` is the name of the predictor we want to send the query to.
  - Takes in a JSON like the following (taken from the "AI Tables in MySQL" Tutorial)
Screen Shot 2020-12-14 at 11 54 55 PM

mindsdb


mindsdb/api/http/namespaces/predictor.py
```diff
@@ -290,6 +293,131 @@ def post(self, name):
...
+                    'collection': data['collection'],
+                    'query': data['find'],
+                    'user': integration['user'],
+                    'password': integration['password'],
+                    'host': integration['host'],
+                    'port': integration['port']
+                }
+            }
+
+            ds = dsClass(**picklable['kwargs'])
```
George3d6, line 400: A lot of this is repeat logic that is contained elsewhere (i.e. it's already done by the component used to create a datasource from a sql query).

It would be better to use the logic that's used to create a datasource from a sql query (maybe seprarated in it's own function if it currently isn't) and then separate the predict logic into another function, and then couple them into this endpoint.

Otherwise we end up with too much alsmot-copy-paste code that needs maintaining,

George3d6 (changes requested):

mindsdb/api/http/namespaces/predictor.py
adzyu100: Hi! We ended up separating the logic for creating a datasource from a sql query into a function in the DataStore class, as the code came from the save_datasource function. Let us know if there's a better place to put that function, as it doesn't really have much to do with the data storage system tbh.

Also, the prediction logic was separated into a new function in the same predictor.py file, and we've changed the three predict endpoints to use that function instead of repeating the same predict logic.

PR #1013, opened 2020-12-16
Database-based storage

George3d6:
*Note: WIP, but it has so many changes I thought it might be useful to put it up here so that it's public, I'm hoping to finish it by Thursday night, including testing*

## Why
* We need a centralized state containing datasources, configs and predicotrs that can be used by multiple mindsdb
* We need this centralized state to be easy to synchronize

## How
* In short, by moving the "logic" data layer of mindsdb to a database (implemented via the sqlalchemy ORM)

## Extra
* Removed installer code
* Removed a bunch of already dead code
* Removed code that was no longer necessary


George3d6:
All tests pass successfully locally but for some reason still fail when ran remotely

.gitignore
```diff
@@ -18,3 +18,7 @@ root
 static
 tmp
 .pytest_cache
+*mindsdb.sqlite3.db
+*mindsdb.sqlite.db
+mindsdb.sqlite3.db
+mindsdb.sqlite.db
```
StpMax, line 24: double lines

mindsdb/__init__.py
```diff
@@ -1,38 +1,38 @@
...
-    pass
+    config_path = None
 
 try:
-    mindsdb_config = Config(config_path)
-except Exception as e:
-    print(str(e))
-    sys.exit(1)
+    with open(args.db_config, 'r') as fp:
+        db_config = json.loads(fp)
```
StpMax, line 16: is no `import json` above

StpMax (changes requested):

StpMax:
need change `error` to `info` here https://github.com/mindsdb/mindsdb/blob/01a9ceeaacabc92022e990b37a8233022c5bf1d6/mindsdb/api/http/initialize.py#L231

StpMax:
`GET datasources/` returns x5 copies of every DS. I upload one `hr1`, then `GET datasources/` return to me:
image

StpMax:
probably, this should be removed:
```
concrete_strength_predictor/
datasources_None_test_ds.zip.tar.gz
predictor_None_test_ds.zip.tar.gz
test_ds/
test_ds_csv/
test_predictor/
test_predictor_csv/
```

StpMax:
If i have current staging version of mindsdb, with predictors and datasources, and then update to your branch, i get two issues:
1. i dont see old datasources in GUI
2. i see old predictors in GUI (but should not), but query to them return error (here https://github.com/mindsdb/mindsdb/blob/distribution_1/mindsdb/interfaces/state/state.py#L70).
May be need add some 'initial migration': if is no `mindsdb.sqlite.db`, then read datasources and predictors and insert them to database?

George3d6:
Closing in favor of: https://github.com/mindsdb/mindsdb/pull/1039

PR #1037, opened 2021-01-12
improve port forwarding closing

StpMax:
**why**

Sometimes error appears on test start: `mysql cant connect to 127.0.0.1 bla bla bla`. The reason is unsuccessful port forwarding before. My guess, that can happens if someone other used same port before and dont close forwarding properly (if, for example, exception raised during test). 

**how**

1. closing all current port-forwarding on test start (need comment this if several tests will be ran in same time)
2. trying open port forwarding in cycle, untill success (no more 10 times)

### DIFF
--- a/tests/integration_tests/flows/common.py
+++ b/tests/integration_tests/flows/common.py
@@ -182,15 +182,16 @@ if USE_EXTERNAL_DB_SERVER:
 
     close_all_ssh_tunnels()
 
-    port_forwarding_status = 1
-    i = 10
-    while port_forwarding_status != 0 and i > 0:
-        i = i - 1
+    for _ in range(10):
         r = requests.get('http://127.0.0.1:5005/port')
         if r.status_code != 200:
             raise Exception('Cant get port to run mindsdb')
         mindsdb_port = r.content.decode()
-        port_forwarding_status = open_ssh_tunnel(mindsdb_port, 'R')
+        status = open_ssh_tunnel(mindsdb_port, 'R')
+        if status == 0:
+            break
+    else:
+        raise Exception('Cant get empty port to run mindsdb')
 
     print(f'use mindsdb port={mindsdb_port}')
     config._config['api']['mysql']['port'] = mindsdb_port
--- a/tests/integration_tests/flows/common.py
+++ b/tests/integration_tests/flows/common.py
@@ -6,6 +6,7 @@ import atexit
 import asyncio
 import shutil
 import csv
+import re
 from pathlib import Path
 
 import requests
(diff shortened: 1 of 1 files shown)


tests/integration_tests/flows/common.py
```diff
@@ -136,33 +137,61 @@ def prepare_config(config, mindsdb_database='mindsdb', override_integration_conf
...
     # sp = subprocess.Popen(f'for pid in $(lsof -i :{port} -t); do kill -9 $pid; done', shell=True)
-    sp = subprocess.Popen(f'ssh -S /tmp/.mindsdb-ssh-ctrl-{port} -O exit ubuntu@3.220.66.106', shell=True)
+    sp = subprocess.Popen(f'ssh -S /tmp/mindsdb/.mindsdb-ssh-ctrl-{port} -O exit ubuntu@3.220.66.106', shell=True)
     sp.wait()
 
 
 def open_ssh_tunnel(port, direction='R'):
-    cmd = f'ssh -i ~/.ssh/db_machine -S /tmp/.mindsdb-ssh-ctrl-{port} -o StrictHostKeyChecking=no -o UserKnownHostsFile=/dev/null -fMN{direction} 127.0.0.1:{port}:127.0.0.1:{port} ubuntu@3.220.66.106'
+    path = Path('/tmp/mindsdb')
+    if not path.is_dir():
```
ilia-tsyplenkov, line 158: propably `path.exists` would be better?

tests/integration_tests/flows/common.py
```diff
@@ -136,33 +137,61 @@ def prepare_config(config, mindsdb_database='mindsdb', override_integration_conf
...
     wait_port(5005, timeout=10)
-    r = requests.get('http://127.0.0.1:5005/port')
-    if r.status_code != 200:
-        raise Exception('Cant get port to run mindsdb')
-    mindsdb_port = r.content.decode()
-    open_ssh_tunnel(mindsdb_port, 'R')
+
+    close_all_ssh_tunnels()
+
+    port_forwarding_status = 1
```
ilia-tsyplenkov, line 185: dont you think that `for` looks better here?
```
for _ in range(10):
        r = requests.get('http://127.0.0.1:5005/port')
        if r.status_code != 200:
            print("attempt to get a port is failed")
            continue
        mindsdb_port = r.content.decode()
        port_forwarding_status = open_ssh_tunnel(mindsdb_port, 'R')
        if port_forwarding_status == 0:
               break
else:
      raise Exception('Cant get port to run mindsdb')

ilia-tsyplenkov (changes requested):

tests/integration_tests/flows/common.py
StpMax: Cant agree, `path.exists` returns `True` on any file.

tests/integration_tests/flows/common.py
StpMax: changed

ilia-tsyplenkov (approved):

PR #1038, opened 2021-01-12
Timeseries prediction latency test

### LINKED ISSUE #1035: Timeseries-focused prediction latency test
Create a test script inside the `tests` directory, which we can run manually and that does the following two things:

1. Trains models from 2x timeseries datasets (@paxcema please recommend the two datasets to be used :d) ... unless the models already exist, in which case no need to re-train every time the script runs

2. For both models run a bunch of predict queries via mindsdb native's `predict` function:
    * 1 by 1 predictions using data in a clickhouse table (i.e. using a `ClickhouseDS` datasource)
     * bulk predictions using data in a clickhouse table (i.e. using a `ClickhouseDS` datasource) ... bulk meaning selecting, say, between 20 and 100 rows
     * 1 by 1 predictions using data passed in the form of a dictionary/pandas dataframe/csv file, including some historical data
     * bulk preidctions using data passed in the form p a pandas dataframe or csv file ... again, bulk here predicting for something like 20 to 100 rows.
     
3. Run the above 4 test cases (or at least the first 2) using the mysql proxy via a clickhouse integration (i.e. sending `SELECT` queries from within clickhouse to the predictor tables created within the mindsdb database)

4. The same thing as 3 and 2, but this time using the python sdk

When this is all done, generate a report detailing how long each of the tasks took in order to compare the latency between these 2 interfaces and mindsdb_native itself, as well as to benchmark the time native needs to run timeseries predictions.

ilia-tsyplenkov:
Close #1035 
How: by creating initial benchmark version

please write `python test.py --help` for details
do not forget to install requirements.txt
please note that indexes show number of rows used in prediction data:

image


George3d6 (changes requested):
Looks fantastic :)

A few small changes if I may ask:

* Test with 1, 5, 10, 20, 50, 100 rows, for the sake of time.
* Write `nr of rows` above the first column if it's easy enough to do.

George3d6:
Also `AutoML` should be called `AITable`, that's how we call our database predictors, AI Tables.

AutoML is just the term we use generally to describe what mindsdb does (Automatic machine learning)

George3d6:
Also, please add the instructions for running in order to produce the outputs shown above

George3d6:
Also, for when `rows == 1` add a testcase called `AITablesManualWhere` in which you specify the values in there `WHERE` query for every column.

I.e. instead of:

`f"SELECT {predict_targets[self.dataset]} FROM mindsdb.{self.dataset} WHERE select_data_query='%s'"`

Have

`f"SELECT {predict_targets[self.dataset]} FROM mindsdb.{self.dataset} WHERE col_1=some_val, col_2=some_other_val, col_3=yet_another_value`

Where `col_1`, `col_2` and `col_3` and so on are the columns of your datasets (i.e. all the headers of the csv file used, minus the one that represents the prediction target).

Hopefully I'm explaining the above well, if not, let me know here or in slack

ilia-tsyplenkov:
@George3d6 Could you please check it again. All notes have been fixed:
image

PR #1044, opened 2021-01-15
Better model importing

George3d6:
Now registering newely imported models

### DIFF
--- a/mindsdb/interfaces/native/mindsdb.py
+++ b/mindsdb/interfaces/native/mindsdb.py
@@ -103,9 +103,8 @@ class MindsdbNative():
         self.dbw.register_predictors(self.get_model_data(new_name), setup=False)
 
     def load_model(self, fpath):
-        F.import_model(model_archive_path=fpath)
-        # @TODO How do we figure out the name here ?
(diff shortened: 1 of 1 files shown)


mindsdb/interfaces/native/mindsdb.py
```diff
@@ -103,9 +103,8 @@ def rename_model(self, name, new_name):
         self.dbw.register_predictors(self.get_model_data(new_name), setup=False)
 
     def load_model(self, fpath):
-        F.import_model(model_archive_path=fpath)
-        # @TODO How do we figure out the name here ?
-        # dbw.register_predictors(...)
+        name = F.import_model(model_archive_path=fpath)
+        self.dbw.register_predictors(self.get_model_data(new_name), setup=False)
```
StpMax, line 107: `name` instead `new_name`?

StpMax (changes requested):

mindsdb/interfaces/native/mindsdb.py
George3d6: Error indeed, we should add test to cover it...

PR #1048, opened 2021-01-16
Better Database Wrapper

George3d6:
* Simplified `get_data` method of the datasource interface
* Removed unused analyze subset route
* Decoupled the integration setup and predictor registration of the DatabaseWrapper


George3d6:
There was an issue before with the Postgres test trying to connect for 1-2+ hours and then timing out, seems to be happening again, I can't quite remember the cause back then though.

PR #1062, opened 2021-01-20
Disable frontend telemetry

StpMax:
**why**

At this moment to disable telemetry need do two things: enter konami code at frontend and set env variable at backend. That can be simplified

**how**

At http api start we can inject to index.html code to disable telemetry, if it disabled for backend. 
Additionally, at backend start will be console output if telemetry is disabled

### DIFF
--- a/mindsdb/__main__.py
+++ b/mindsdb/__main__.py
@@ -24,6 +24,8 @@ from mindsdb.utilities.ps import is_pid_listen_port
 from mindsdb.interfaces.database.database import DatabaseWrapper
 from mindsdb.utilities.functions import args_parse, get_all_models_meta_data
 from mindsdb.utilities.log import initialize_log
+from mindsdb.utilities.telemetry import is_telemetry_file_exists, disable_telemetry
+
 
 def close_api_gracefully(apis):
     try:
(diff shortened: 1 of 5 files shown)


mindsdb/__main__.py
```diff
@@ -44,6 +44,10 @@ def close_api_gracefully(apis):
 
     config = Config(os.environ['MINDSDB_CONFIG_PATH'])
 
+    disable_telemetry = os.getenv('CHECK_FOR_UPDATES').lower() in ['0', 'false']
+    if disable_telemetry:
+        print('\n ✓ telemetry is disabled \n')
```
George3d6, line 49: We should be using the `TELEMETRY_FILE` here (see in utils.py -- name is `telemtry.lock`

The logic should be:

On start, if `telemtry.lock` file is present => set CHECK_FOR_UPDATES to 0 (using the function in uitls.py... there should be one)
If `CHECK_FOR_UPDATES` is 0, false, False ... etc => create the `telemtry.lock` file (using the function in utils.py)

Also, we should be re-updating the telemtry value in the gui every time `telemtry.lock` is changed, this is fairly easy since it gets changed with a function in utils.py, so just updated the `index.html` when that happens.

George3d6 (changes requested):

mindsdb/__main__.py
StpMax: done

PR #1080, opened 2021-02-02
add _class_distribution to apis output

StpMax:
**why**

To output `*_class_distribution` in api, for models trained with `output_class_distribution: True`

**how**

added virtual fields for `*_class_distribution` to mysql and mongo apis


StpMax:
Gonna ti close that, since i added class_distribution into 'explain' of prediction https://github.com/mindsdb/mindsdb_native/pull/430.

PR #1083, opened 2021-02-04
Fix for checking connection to atlas data lake

StpMax:
**why**

Checking connection to data lake always failed. That, probably, because connection to data lake should be with `tls=true`.

**how**

if `mongo.net` in connection sting, then we set that option. Also, for `mongo.net` set `tlsCAFile=certifi.where()` since Alejo cant connect without it.

### DIFF
--- a/mindsdb/integrations/mongodb/mongodb.py
+++ b/mindsdb/integrations/mongodb/mongodb.py
@@ -1,3 +1,6 @@
+import re
+
+import certifi
 from pymongo import MongoClient
 
 from mindsdb.integrations.base import Integration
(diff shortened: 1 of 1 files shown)


mindsdb/integrations/mongodb/mongodb.py
```diff
@@ -19,12 +22,34 @@ def unregister_predictor(self, name):
...
+
+            kwargs = {}
+
+            if isinstance(user, str) and len(user) > 0:
+                kwargs['username'] = user
+
+            if isinstance(password, str) and len(password) > 0:
+                kwargs['password'] = password
+
+            if re.match(r'\/\?.*tls=true', host.lower()):
```
ilia-tsyplenkov, line 37: `host.lower().endswith('tls=true')` looks a little bit simpler, or not?

ilia-tsyplenkov (changes requested):

PR #1087, opened 2021-02-04
Allow dot in mongo find

StpMax:
**why**

to allow use 'dot' notation in mongo `find`: db.predictor.find({'a.b': 1})

**how**

added additional check for field exists

### DIFF
--- a/mindsdb/api/mongo/responders/find.py
+++ b/mindsdb/api/mongo/responders/find.py
@@ -40,9 +40,7 @@ class Responce(Responder):
 
             for key in where_data:
                 if key not in columns:
-                    sub_key = key[:key.find('.')] if '.' in key else key
-                    if sub_key not in columns:
-                        raise Exception(f"Unknown column '{key}'. Only columns from this list can be used in query: {', '.join(columns)}")
+                    columns.append(key)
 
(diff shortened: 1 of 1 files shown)


mindsdb/api/mongo/responders/find.py
```diff
@@ -40,7 +40,9 @@ def result(self, query, request_env, mindsdb_env, session):
 
             for key in where_data:
                 if key not in columns:
-                    raise Exception(f"Unknown column '{key}'. Only columns from this list can be used in query: {', '.join(columns)}")
+                    sub_key = key[:key.find('.')] if '.' in key else key
+                    if sub_key not in columns:
+                        raise Exception(f"Unknown column '{key}'. Only columns from this list can be used in query: {', '.join(columns)}")
```
George3d6, line 45: I think that for now it might be best if we just skip this bit for Mongo, it makes more sense for SQL, but mongo is unstructred anyway, people might well want to predict with data different from the training set, even if EOD the columns there will be ignored.

So unless you're strongly against (in which case let's discuss) I'd say drop the schema check for mongo entirely.

George3d6 (changes requested):

George3d6:
Note: Tests seem to fail due to connection issues.

StpMax:
Removed schema check. Also, any fields in 'find' will be in result.

PR #1106, opened 2021-02-12
change logging levels

StpMax:
**why**

to reduce messages count

**how**


George3d6 (commented):
small change to no longer pass config to `initialize_log`. Ideally we can stop config from being passed around for now.

PR #1108, opened 2021-02-15
close learn process gracefully

### LINKED ISSUE #1058: Close learn process gracefuly
`close_api_gracefully` only closes the API, but not any ongoing learn processes, which are left orphaned and consuming loads of memory.

ilia-tsyplenkov:
Fixes #1058 

  ## What:
 - ```close_api_gracefully only closes the API, but not any ongoing learn processes, which are left orphaned and consuming loads of memory.```
  ## How:
 - Each learning process when start working stories its own pid as empty file in `storage_dir/in_learning` directory and will delete it at the end if no interruption happens
 - `close_api_gracefully` read list of files in `storage_dir/in_learning` and sends SIGTERM to each of this process.

mindsdb

### DIFF
--- a/mindsdb/__main__.py
+++ b/mindsdb/__main__.py
@@ -16,25 +16,19 @@ from mindsdb.interfaces.custom.custom_models import CustomModels
 from mindsdb.api.http.start import start as start_http
 from mindsdb.api.mysql.start import start as start_mysql
 from mindsdb.api.mongo.start import start as start_mongo
-from mindsdb.utilities.ps import is_pid_listen_port
+from mindsdb.utilities.ps import is_pid_listen_port, get_child_pids
 from mindsdb.interfaces.database.database import DatabaseWrapper
 from mindsdb.utilities.functions import args_parse, get_all_models_meta_data
 from mindsdb.utilities.log import log
 
 
-def close_api_gracefully(apis, config):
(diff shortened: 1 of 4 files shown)


StpMax:
What a reason was to choose this solution?
May be just go recursively through processes children at the end of main process?
```
import psutil
p = psutil.Process()
c = p.children(recursive=True)
for p in c:
    p.kill()
```

ilia-tsyplenkov:
> @ilia-tsyplenkov What a reason was to choose this solution?
> May be just go recursively through processes children at the end of main process?
> 
> ```
> import psutil
> p = psutil.Process()
> c = p.children(recursive=True)
> for p in c:
>     p.kill()
> ```
Yes, this is first that I decided to implement. But since psutil is not fully cross-platform package, I decided to implement approach which is definitely will work on linux/mac/windows

StpMax:
I checked that on macos - looks like it work. Can you try on windows?
I think getting process's children should be safe, so OS should not block that.

ilia-tsyplenkov:
> I checked that on macos - looks like it work. Can you try on windows?
> I think getting process's children should be safe, so OS should not block that.

Sorry but I don't have Windows

StpMax (approved):

PR #1109, opened 2021-02-16
'check connection' not worked for existing connection

### LINKED ISSUE #1082: 'check connection' not worked for existing connection
We have 'check connection' button at 'edit connection' screen of GUI. Button is not working, something broken at backend.

ilia-tsyplenkov:
Fixes #1082

**What:**
 - right now it is possible to check connection for `real` integrations that is why check connection (put /integration request with {test: true}) fails

**How**:
 - Provide connection checker for each integration type as a mixin to decouple strong conjunction between Integration abstraction and particular call to different database types


George3d6:
@ilia-tsyplenkov I like the idea, but to avoid code duplication you should just add a `create_connection(self)` method to the database classes that creates then returns a connection based on the information contained within `self.`  Other than that looks good.

PR #1111, opened 2021-02-17
checking mongo password

StpMax:
**why**

To validate password on client and server

**how**


mindsdb/api/mongo/responders/sasl_start.py
```diff
@@ -17,8 +21,10 @@ def result(self, query, request_env, mindsdb_env, session):
                 'payload': responce.encode(),
                 'ok': 1
             }
-        except Exception:
+        except Exception as e:
+            print(e)
```
George3d6, line 25: use logging please

George3d6 (changes requested):
I think `scram` needs to be added to requirements.txt

Overall looks good.

PR #1115, opened 2021-02-20
sql 'create model' command

StpMax:
**why**

add support for new commands for mindsdb mysql api:
- [x] 'create model'
- [ ] 'create database'
...

**how**


George3d6:
@StpMax How do you detect the dialect used (standard MySQL vs custom mindsdb SQL), can just look for `CREATE PREDICTOR` or `CREATE AI TABLE` and that's enough not to have parsing logic conflicts?

StpMax:
@George3d6 Which 'logic conflicts' you mean? At this moment that work as you wrote: at first step detected type of query: 'create predictor', 'create ai table', 'delete' etc. Then query parsed by specific parser.

PR #1123, opened 2021-02-26
Route to check native process exists

StpMax:
**why**

To resolve https://github.com/mindsdb/mindsdb/issues/1121

**how**

By set title for native process, and check that title at route handler.

### DIFF
--- a/mindsdb/api/http/namespaces/util.py
+++ b/mindsdb/api/http/namespaces/util.py
@@ -1,6 +1,7 @@
 from flask import request
 from flask_restx import Resource
 from flask import current_app as ca
+import psutil
 
 from mindsdb.utilities.log import log
 from mindsdb.api.http.namespaces.configs.util import ns_conf
@@ -19,6 +20,20 @@ class Ping(Resource):
         '''Checks server avaliable'''
         return {'status': 'ok'}
 
+
+@ns_conf.route('/ping_native')
+class PingNative(Resource):
+    @ns_conf.doc('get_ping_native')
+    def get(self):
+        ''' Checks server use native for learn or analyse.
+            Will return right result only on Linux.
+        '''
+        for p in psutil.process_iter(['name']):
+            if p.info['name'] == 'mindsdb_native_process':
+                return {'native_process': True}
+        return {'native_process': False}
+
+
 @ns_conf.route('/report_uuid')
 class ReportUUID(Resource):
     @ns_conf.doc('get_report_uuid')
@@ -28,6 +43,7 @@ class ReportUUID(Resource):
         return {
             'report_uuid': predictor.report_uuid
         }
+
 
 @ns_conf.route('/telemetry')
 class Telemetry(Resource):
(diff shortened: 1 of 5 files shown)


mindsdb/interfaces/datastore/datastore.py
```diff
@@ -26,7 +25,20 @@ def __init__(self):
...
+
+            analysis = self.mindsdb_native.analyse_dataset(self.get_datasource_obj(name))
+
+            try:
+                setproctitle.setproctitle(original_process_title)
+            except Exception:
+                pass
+
+            datasource_record.analysis = json.dumps(analysis)
             session.commit()
```
George3d6, line 42: Should also be inside the proc title guard

mindsdb/interfaces/native/learn_process.py
```diff
@@ -46,13 +47,25 @@ def run(self):
...
                 to_predict=to_predict,
                 **kwargs
             )
-        except Exception as e:
+        except Exception:
+            pass
+
+        try:
+            setproctitle.setproctitle(original_proc_title)
+        except Exception:
```
George3d6, line 68: This should be set when the learn process starts and unset when it ends (or rather, no need to ever unset it)

George3d6: That is because we don't want to e.g. start learning, then the tilte changes, and then the predictor never gets put in the database or dumped to s3

mindsdb/interfaces/datastore/datastore.py
```diff
@@ -26,7 +25,20 @@ def __init__(self):
...
+                setproctitle.setproctitle('mindsdb_native_process')
+            except Exception:
+                pass
+
+            analysis = self.mindsdb_native.analyse_dataset(self.get_datasource_obj(name))
+
+            try:
+                setproctitle.setproctitle(original_process_title)
+            except Exception:
+                pass
```
George3d6, line 39: This approach can actually be used to "guard" predict calls as well (as part of the native interface)

mindsdb/api/http/namespaces/util.py
```diff
@@ -19,6 +20,20 @@ def get(self):
...
+class PingNative(Resource):
+    @ns_conf.doc('get_ping_native')
+    def get(self):
+        ''' Checks server use native for learn or analyse.
+            Will return right result only on Linux.
+        '''
+        for p in psutil.process_iter(['name']):
+            if p.info['name'] == 'mindsdb_native_process':
+                return {'native_process': True}
+        return {'native_process': False}
```
George3d6, line 34: overall (given that we also implement the "predict guard" this might be sufficient to be able to tell wether a mindsdb instance is working or not, with some caveates (i.e. maybe we keep pinging this for ~5 minutes before closing) :)

George3d6 (changes requested):

mindsdb/interfaces/datastore/datastore.py
StpMax: I added this to 'predict' method, but actually it not called in separate thread/process at this moment.

PR #1136, opened 2021-03-05
Delete broken datasources + loggin errors

StpMax:
**why**

1. to add more errors logging on ds creation fails
2. i think it will be good to remove DS if it is broken, instead ignore it
3. there was possible get 'broken' ds if something happen in `def save_datasource` (even with `sesson.delete(ds_record)` in exception handler, because by default autocommit is false).

**how**

### DIFF
--- a/mindsdb/interfaces/datastore/datastore.py
+++ b/mindsdb/interfaces/datastore/datastore.py
@@ -51,6 +51,8 @@ class DataStore():
             datasource_record_arr = session.query(Datasource).filter_by(company_id=self.company_id)
         for datasource_record in datasource_record_arr:
             try:
+                if datasource_record.data is None:
+                    continue
                 datasource = json.loads(datasource_record.data)
                 datasource['created_at'] = datasource_record.created_at
                 datasource['updated_at'] = datasource_record.updated_at
@@ -59,7 +61,6 @@ class DataStore():
                 datasource_arr.append(datasource)
             except Exception as e:
                 log.error(e)
-                self.delete_datasource(datasource_record.name)
         return datasource_arr
 
     def get_data(self, name, where=None, limit=None, offset=None):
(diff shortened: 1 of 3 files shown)


mindsdb/interfaces/datastore/datastore.py
```diff
@@ -62,6 +59,7 @@ def get_datasources(self, name=None):
                 datasource_arr.append(datasource)
             except Exception as e:
                 log.error(e)
+                self.delete_datasource(datasource_record.name)
```
George3d6, line 62: This we shouldn't do, as in the current process a "broken" datasource could be "under construction".
Also, there should be no reason why a datasource could be "Broken" so for now let's keep them in the database to see what can go wrong to cause them to be so.

mindsdb/interfaces/datastore/datastore.py
```diff
@@ -104,18 +102,12 @@ def delete_datasource(self, name):
...
-
         if source_type == 'file' and (file_path is None):
             raise Exception('`file_path` argument required when source_type == "file"')
 
         ds_meta_dir = os.path.join(self.dir, name)
         os.mkdir(ds_meta_dir)
 
-        session.add(datasource_record)
-        session.commit()
-        datasource_record = session.query(Datasource).filter_by(company_id=self.company_id, name=name).first()
```
George3d6, line 117: In principle I agree with only creating the datasource in a delayed manner in the database, the issue here will be one of locking, i.e. atm we have to mechanism to "lock" the ability for multiple clients to work on an entitiy (e.g. datasource) with the same name, unless it's already in the database, so this could cause issues.

But, regardless, I think it's unlikely to happen. So keep this as it.

Still, please don't delete "broken" datasources, keep them in the database so we can figure out why they are broken later.

mindsdb/interfaces/datastore/datastore.py
```diff
@@ -252,23 +244,30 @@ def save_datasource(self, name, source_type, source, file_path=None):
...
+            session.commit()
 
         except Exception as e:
+            log.error(f'{e}')
             if os.path.isdir(ds_meta_dir):
                 shutil.rmtree(ds_meta_dir)
-            session.delete(datasource_record)
-            raise e
+            session.rollback()
+            raise
```
George3d6, line 269: We could just call the internal `delete_datasource` function here, I think, + log the error.

George3d6 (changes requested):

PR #1154, opened 2021-03-11
Memory optimization and ray usage

George3d6:
Optimizing memory and using ray to run native calls in parallel and (in the future) on multiple machines.


mindsdb/interfaces/model/model_controller.py
```diff
@@ -0,0 +1,262 @@
...
+        elif pred_format == 'dict':
+            predictions = [p.as_dict() for p in predictions]
+        elif pred_format == 'dict&explain':
+            predictions = [[p.as_dict() for p in predictions], [p.explain() for p in predictions]]
+        else:
+            raise Exception(f'Unkown predictions format: {pred_format}')
+
+        if ray_based:
+            return predictions
+        return xmlrpc.client.Binary(pickle.dumps(predictions))
```
StpMax, line 142: may be add to class
```
def _return(self, obj):
        if ray_based:
            return obj
        return xmlrpc.client.Binary(pickle.dumps(obj))
```
and change returns to call to that function? just to make it slightly more beauty :)

PR #1172, opened 2021-03-25
Detecting learn processes

StpMax:
**why**

to check is any learn process exists

**how**

Detecting learn process if ray in use is little bit tricky then without it. Ray setting his own process names, so using 'setproctitle' to mark learn processes is impossible. Best what i can found is creating temp files for all learn time.


George3d6 (changes requested):
Ok, short clarification:

Detecting predict processes is important,

I'd just like to be able to differentiate between predict and learn.

So keep the predict logic, but make it so that we can get `native_process: 'learn'` and `native_process: 'other'`

PR #1179, opened 2021-03-30
Support Kafka streams

ilia-tsyplenkov:
Support Kafka streams at it was done for Redis

### DIFF
--- a/mindsdb/integrations/kafka/kafkadb.py
+++ b/mindsdb/integrations/kafka/kafkadb.py
@@ -1,0 +1,58 @@
+import os
+import json
+import kafka
+
+from threading import Thread
+from mindsdb.utilities.config import STOP_THREADS_EVENT
+from mindsdb.utilities.log import log
+from mindsdb.integrations.base import Integration
+
+
+class KafkaConnectionChecker:
+    def __init__(self, **kwargs):
+        self.host = kwargs.get('host')
+        self.port = kwargs.get('port', 9092)
+
+    def _get_connection(self):
+        return kafka.KafkaAdminClient(bootstrap_servers=f"{self.host}:{self.port}")
+    def check_connection(self):
+        try:
+            client = self._get_connection()
+            client.close()
+            return True
+        except Exception:
+            return False
+
+
+class Kafka(Integration, KafkaConnectionChecker):
+    def __init__(self, config, name):
+        Integration.__init__(self, config, name)
+        intergration_info = self.config['integrations'][self.name]
+        self.host = intergration_info.get('host')
+        self.port = intergration_info.get('port', 9092)
+        self.control_topic_name = intergration_info.get('topic')
+        self.client = self._get_connection()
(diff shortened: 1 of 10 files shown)


mindsdb/integrations/kafka/kafkadb.py
```diff
@@ -0,0 +1,172 @@
...
+
+    def work(self):
+        self.consumer = kafka.KafkaConsumer(bootstrap_servers=f"{self.host}:{self.port}",
+                                            consumer_timeout_ms=1000)
+
+        self.consumer.subscribe([self.control_topic_name])
+        log.error(f"Integration {self.name}: subscribed  to {self.control_topic_name} kafka topic")
+        while not self.stop_event.wait(0.5):
+            try:
+                # break if no record about this integration has found in db
```
George3d6, line 109: If this comment is necessary, than the method should be called `integration_in_db`  or something similar, if the method is or will be checking for more than db presence, then remove this comment

mindsdb/integrations/kafka/kafkadb.py
```diff
@@ -0,0 +1,172 @@
...
+        self.streams = {}
+        self.stop_event = STOP_THREADS_EVENT
+
+    def should_i_exist(self):
+        config_record = session.query(Configuration).filter_by(company_id=self.company_id).first()
+        if config_record is None:
+            return False
+        integrations = json.loads(config_record.data)["integrations"]
+        if self.name not in integrations:
+            return False
```
George3d6, line 46: This check is being done the wrong way, please check the configuration object directly since that already does the merging of the file config with the db config.

If we can't access the config here then we are doing something wrong and the control flow should be changed, but I think we can.

George3d6 (changes requested):

mindsdb/integrations/kafka/kafkadb.py
```diff
@@ -0,0 +1,172 @@
...
+
+
+    def _query(self):
+        pass
+
+    def register_predictors(self, model_data_arr):
+        pass
+
+    def unregister_predictor(self, name):
+        pass
```
George3d6, line 172: Why are these here, why not just skip streaming integration in the respective checks? Or is it easier this way?

George3d6 (changes requested):

mindsdb/integrations/kafka/kafkadb.py
ilia-tsyplenkov: Fixed

mindsdb/integrations/kafka/kafkadb.py
ilia-tsyplenkov: Renamed to `exist_in_db`

mindsdb/integrations/kafka/kafkadb.py
ilia-tsyplenkov: Thanks. Changed on `Config()` usage

PR #1194, opened 2021-04-07
fix static paths

StpMax:
**why**

it was some issues with GUI opening if 'storage_dir' is relative

**how**

move all static_paths to absolute


ZoranPandovski (approved):

ZoranPandovski:
Fixes #1193

PR #1215, opened 2021-04-20
Fix xmlrpc timestamp conversion

StpMax:
**why**

xmlrpc can not transfer complex types, like pandas timestamp


George3d6 (changes requested):
Whatever function of XMLRPC the dataframe is passed to should instead accept the dataframe as a pickle (this is already done for some functions)

PR #1217, opened 2021-04-21
Add timezone to vars

maximlopin:
Closes https://github.com/mindsdb/mindsdb/issues/1214

`api/config/vars` now returns a new key `'timezone'`


mindsdb/api/http/namespaces/config.py
```diff
@@ -179,7 +182,18 @@ def get(self):
...
+        local_time = datetime.datetime.now(tzlocal())
+        local_timezone = local_time.tzname()
+
+        return {
+            'mongo': mongo,
+            'telemtry': telemtry,
+            'cloud': cloud,
+            'timezone': {
+                'name': local_timezone,
+                'time': local_time,
```
George3d6, line 194: Sending the local time isn't required, timzeone is enough, extra data makes things confusing, just send `'timezone': local_timezone`, and that should suffice :)

George3d6 (changes requested):

PR #1225, opened 2021-04-23
Company independent

George3d6:


mindsdb/interfaces/model/model_controller.py
```diff
@@ -23,10 +23,9 @@
 
 class ModelController():
     def __init__(self, ray_based):
```
StpMax, line 25: should `company_id` be in args?

StpMax (changes requested):

PR #1253, opened 2021-05-13
Timeseries fix + various

George3d6:
Mindsdb won't return any rows if the `data` resulting from the `select` function in the datanode has the values for the received columns changed. (when using from within clickhouse)

E.g. if we get ` WHERE A=1 AND B='2'` but then try to return the data dict:
`{'A': None, 'B': '2', 'to_pred': 13242}`, then we will see `0` rows being returned to clickhouse.

The original values that the prediction is generated from being made "null" was happening inside the native `predict` call and is an unknown bug, but in case this happens in the future it might make sense to implement a "fix" for it mindsdb side, which this PR does. I assume this might have to do with clickhouse doing the `WHERE` filtering on their side too, but am unsure.

Also:

* Fixed `get_model` to work before `name` gets set as a property in the local metadata.
* Fixed `error` logging to also print the traceback.

### DIFF
--- a/mindsdb/api/mysql/mysql_proxy/datahub/datanodes/mindsdb_datanode.py
+++ b/mindsdb/api/mysql/mysql_proxy/datahub/datanodes/mindsdb_datanode.py
@@ -266,6 +266,7 @@ class MindsDBDataNode(DataNode):
             return data
         else:
             pred_dicts, explanations = self.model_interface.predict(table, 'dict&explain', when_data=where_data)
+            # Fix since for some databases we *MUST* return the same value for the columns originally specified in the `WHERE`
             if isinstance(where_data, list):
                 for i in range(len(pred_dicts)):
                     for col in where_data[i]:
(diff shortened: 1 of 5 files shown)


mindsdb/api/mysql/mysql_proxy/datahub/datanodes/mindsdb_datanode.py
```diff
@@ -265,6 +266,13 @@ def select(self, table, columns=None, where=None, where_data=None, order_by=None
             return data
         else:
             pred_dicts, explanations = self.model_interface.predict(table, 'dict&explain', when_data=where_data)
+            # Fix since for some databases we *MUST* return the same value for the columns originally specified in the `WHERE`
+            if isinstance(where_data, list):
```
StpMax, line 270: Im think `where_data` can be `dict` in some cases. Should we process that case?

StpMax (approved):

mindsdb/api/mysql/mysql_proxy/datahub/datanodes/mindsdb_datanode.py
George3d6: Oh, ok

PR #1290, opened 2021-06-07
Refactor streams

maximlopin:
- all complex logic is moved to `BaseStream` class, its subclasses (`KafkaStream`, `RedisStream`) only implement a few small methods to read/write to/from in/out streams
- `group_by` and `order_by` can now be lists
- when the number of records read from the input stream exceeds `ts_settings['window']`, only last `ts_settings['window']` records are used to predict, others are discarded. This way timeseries cache has constant space complexity so it's stored in a python `dict`.

### DIFF
--- a/mindsdb/integrations/base/integration.py
+++ b/mindsdb/integrations/base/integration.py
@@ -1,6 +1,4 @@
-COMPANY_ID = None
 import os
-from re import S
 from threading import Thread
 
 from mindsdb.utilities.config import STOP_THREADS_EVENT
@@ -35,8 +33,9 @@ class StreamIntegration(Integration):
         Thread(target=StreamIntegration._loop, args=(self, )).start()
 
     def _loop(self):
+        company_id = os.environ.get('MINDSDB_COMPANY_ID', None)
         while not STOP_THREADS_EVENT.wait(1.0):
-            stream_db_recs = db.session.query(db.Stream).filter_by(company_id=COMPANY_ID, integration=self.name).all()
+            stream_db_recs = db.session.query(db.Stream).filter_by(company_id=company_id, integration=self.name).all()
 
             # Stop streams that weren't found in DB
             indices_to_delete = []
@@ -59,14 +58,17 @@ class StreamIntegration(Integration):
             s.thread.join()
             print('3s', s)
 
+    def check_connection(self):
+        raise NotImplementedError
+
     def _make_stream(self, s):
         raise NotImplementedError
 
     def _query(self, query, fetch=False):
-        raise NotImplementedError
+        pass
 
     def register_predictors(self, model_data_arr):
-        raise NotImplementedError
+        pass
 
     def unregister_predictor(self, name):
-        raise NotImplementedError
+        pass
--- a/mindsdb/integrations/base/integration.py
+++ b/mindsdb/integrations/base/integration.py
@@ -1,115 +1,72 @@
+COMPANY_ID = None
 import os
-from abc import ABC, abstractmethod
-from mindsdb.utilities.config import Config, STOP_THREADS_EVENT
-from mindsdb.interfaces.storage.db import session, Stream
-from mindsdb.utilities.log import log
-from mindsdb.interfaces.database.integrations import get_db_integration
+from re import S
+from threading import Thread
+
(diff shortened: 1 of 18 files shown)


mindsdb/integrations/redis/redisdb.py
```diff
@@ -1,143 +1,42 @@
...
-        self.advanced_info = kwargs.get("advanced", {})
-        self.connection_params = {}
-        self.connection_params.update(self.connection_info)
-        self.connection_params.update(self.advanced_info)
-
-    def _get_connection(self):
-        return walrus.Database(**self.connection_params)
+    def __init__(self, **params):
+        self.connection_info = {
+            'host': params['connection']['host'],
```
ilia-tsyplenkov, line 11: Same notes as for Kafka Integration connection/auth parameters

mindsdb/integrations/redis/redisdb.py
```diff
@@ -1,143 +1,42 @@
...
+            'port': params['connection']['port'],
+            'password': params['connection']['password'],
+        }
 
     def check_connection(self):
-        client = self._get_connection()
         try:
+            client = walrus.Database(**self.connection_info)
```
ilia-tsyplenkov, line 18: It may be false-positive: for e.g connection might be successfully established, but you can't do anything because of lack of permissions (as example empty or wrong password)
```
import walrus
CONNECTION_PARAMS = {"host": "3.220.66.106", "port": 6379, "db": 0, "password": "WRONG_PASS"}
client = walrus.Database(**CONNECTION_PARAMS)
```

mindsdb/streams/kafka/kafka_stream.py
```diff
@@ -1,106 +1,24 @@
 import json
-from threading import Thread
-
 import kafka
 
-from mindsdb.utilities.log import log
-from mindsdb.utilities.cache import Cache
```
ilia-tsyplenkov, line 7: Please use `Cache` for TS caching. It is based on `shelve` python package or Redis

mindsdb/integrations/base/integration.py
```diff
@@ -1,109 +1,64 @@
...
-    @abstractmethod
-    def work(self):
-        pass
-
-    @abstractmethod
-    def start(self):
-        pass
+        Thread(target=StreamIntegration._loop, args=(self,)).start()
+
+    def _loop(self):
```
ilia-tsyplenkov, line 36: Generally each integration has to have opportunity to create Streams via native API (Redis, Kafka or others)  by sending messages in `control_stream`. Please add it in the `__loop` method.

tests/integration_tests/flows/test_kafka.py
```diff
@@ -30,32 +28,38 @@
 STREAM_OUT_TS = f"test_stream_out_ts_{STREAM_SUFFIX}"
 DS_NAME = "kafka_test_ds"
 
-def read_stream(stream_name, buf, stop_event):
```
ilia-tsyplenkov, line 33: With default Kafka configuration (when no groups exist). A consumer must be launched before a producer.
Otherwise all data written before consumer starts would be lost.

tests/integration_tests/flows/test_redis.py
```diff
@@ -81,163 +99,71 @@ def train_ts_predictor(self, ds_name, predictor_name):
...
-        while len(stream_in) and time.time() < threshold:
-            time.sleep(1)
-        time.sleep(10)
-        prediction = stream_out.read()
-        stream_out.trim(0, approximate=False)
-        stream_in.trim(0, approximate=False)
-        self.assertTrue(len(prediction)==2, f"expected 2 predictions, but got {len(prediction)}")
-
-
-    def test_6_create_redis_stream_native_api(self):
```
ilia-tsyplenkov, line 189: We need support native api. Please restore them back.

ilia-tsyplenkov (changes requested):
@maximlopin  Thanks a lot. New structure of classes looks pretty good. But there are several notes from my side. I don't think that it takes too much time from you to fix them.

mindsdb/integrations/redis/redisdb.py
maximlopin: so `client.dbsize()` is not enough to check the connection?

PR #1307, opened 2021-06-28
Update model controller for lightwood refactor

maximlopin:
https://github.com/mindsdb/mindsdb/issues/1299

this should be merged first https://github.com/mindsdb/lightwood/pull/443


mindsdb/interfaces/model/model_controller.py
```diff
@@ -100,56 +105,6 @@ def _try_outdate_db_status(self, predictor_record):
...
-            if join_learn_process is True:
-                p.join()
-                if p.exitcode != 0:
-                    delete_process_mark('learn')
-                    raise Exception('Learning process failed !')
-
-        delete_process_mark('learn')
-        return 0
-
     def predict(self, name, pred_format, when_data=None, kwargs={}, company_id=None):
```
George3d6, line 108: For predict we might need a "backwards compatible" flag for now, that converts the new lightwood format into the old format that mindsdb was using.

Basically:

`prediction` -> `target_name`
`confidence` -> `target_name_confidence`
`lower_bound` -> `target_name_lower_bound`

Double check the actual names returned by native and by lightwood, and do the conversion here, and have a `backwards_compatible` argument (set to `True` by default)

George3d6 (changes requested):

PR #1324, opened 2021-07-12
Don't allow deleting datasources if they are linked to a predictor

### LINKED ISSUE #1280: Don't allow deleting datasources if they are linked to a predictor
Don't allow a datasource to be deleted if a predictor that exists was trained from it, otherwise this will break update and (in the future) potentially retraining functionality.

We can toggle this behavior off in the config.

ilia-tsyplenkov:
Fixes #1280

Perform simple checking before delete a datasource.
By setting `force_datasource_removing` to True in config any datasource would be deleted despite on any linked models.

### DIFF
--- a/mindsdb/api/http/namespaces/datasource.py
+++ b/mindsdb/api/http/namespaces/datasource.py
@@ -28,8 +28,6 @@ from mindsdb.api.http.namespaces.entitites.datasources.datasource_missed_files i
 )
 from mindsdb.interfaces.database.integrations import get_db_integration
 
-from mindsdb.utilities.config import Config
-
 
 def parse_filter(key, value):
     result = re.search(r'filter(_*.*)\[(.*)\]', key)
@@ -77,11 +75,6 @@ class Datasource(Resource):
     def delete(self, name):
         '''delete datasource'''
 
-        if not Config()["force_datasource_removing"]:
-            models = request.native_interface.get_models()
-            linked_models = [x["name"] for x in models if x['data_source_name'] == name]
-            if linked_models:
-                return "Can't delete {} datasource because there are next models linked to it: {}".format(name, linked_models), 403
         try:
             request.default_store.delete_datasource(name)
         except Exception as e:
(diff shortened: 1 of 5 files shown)


mindsdb/api/http/namespaces/entitites/predictor_status.py
```diff
@@ -10,6 +10,7 @@
     # other attributes
     'is_active': fields.Boolean(required=False, description='Only one predictor by public_name can be active'),
     'data_source': fields.String(required=False, description='The data source it\'s learning from'),
+    'data_source_name': fields.String(required=False, description='The name of the datasource it\'s learning from'),
```
StpMax, line 13: is this not the same as 'data_source'?

mindsdb/api/http/namespaces/datasource.py
```diff
@@ -74,6 +76,12 @@ def get(self, name):
...
     def delete(self, name):
         '''delete datasource'''
+
+        if not Config()["force_datasource_removing"]:
+            models = request.native_interface.get_models()
+            linked_models = [x["name"] for x in models if x['data_source_name'] == name]
+            if linked_models:
+                return "Can't delete {} datasource because there are next models linked to it: {}".format(name, linked_models), 403
         try:
             request.default_store.delete_datasource(name)
```
StpMax, line 86: It would be better to check 'force_datasource_removing' in 'delete_datasource', because datasource can be removed not only through http api

StpMax (approved):

mindsdb/api/http/namespaces/entitites/predictor_status.py
ilia-tsyplenkov: No, in general `data_source` points to the .csv file with data

mindsdb/api/http/namespaces/datasource.py
ilia-tsyplenkov: Thanks, that is a really good note. Fixed

PR #1404, opened 2021-08-18
Migration support

### LINKED ISSUE #1220: Add sqlalchemy migration
We should add a migration routine that gets executed whenever mindsdb starts to support backwards compatibility for older predictors/datasource (or at least delete the no-longer-supported rows as part of the migration?).

Though as it stands cloud kind of relies on mindsdb always having the same schema, so not sure how much this will help, since we have to try and stop modifying all existing schemas anyway.

ilia-tsyplenkov:
Fixes #1220

  - Add a separate `make_migration.py` file to create new migrations and upgrade/downgrade mindsdb database
  - Add required functions and attributes in `storage/db.py` to accept interface of flask_sqlalchemy.SQLAlchemy instance

## How to work wit migrations:
 - activate mindsdb virtual env
 - `cd mindsdb/mindsdb`
 - `export FLASK_APP=make_migrations.py`
 - `flask db upgrade` to launch migrations
 - `flask migrate -m "MIGRATION INFO"` to add new migration
 - `flask downgrade` - to revert the latest migration
please note: In case if `flask migrate` or `flask upgrade` returns `NameError ... unknown 'mindsdb'` need to import `mindsdb` package in migrations/versions/<migration_file>.py manually 
- 
## Auto-upgrade
Db upgrade running in each start except cloud case


mindsdb/__init__.py
```diff
@@ -2,11 +2,12 @@
...
-from mindsdb.__about__ import __version__
-from mindsdb.__about__ import __version__ as mindsdb_version
-from mindsdb.utilities.fs import get_or_create_data_dir, create_dirs_recursive
-from mindsdb.utilities.functions import args_parse, is_notebook
-from mindsdb.utilities.telemetry import telemetry_file_exists, disable_telemetry
+from .__about__ import __version__
+from .__about__ import __version__ as mindsdb_version
+from .utilities.fs import get_or_create_data_dir, create_dirs_recursive
+from .utilities.functions import args_parse, is_notebook
+from .utilities.telemetry import telemetry_file_exists, disable_telemetry
```
StpMax, line 9: why imports are relative?

mindsdb/interfaces/storage/db.py
```diff
@@ -164,3 +164,8 @@ class Stream(Base):
 
 Base.metadata.create_all(engine)
 orm.configure_mappers()
+metadata = Base.metadata
+
+
+def get_engine():
+    return engine
```
StpMax, line 171: unused

mindsdb/make_migrations.py
```diff
@@ -0,0 +1,12 @@
+import os
+from flask import Flask
+from flask_migrate import Migrate
+
+import mindsdb.mindsdb.interfaces.storage.db as db
+
+app = Flask(__name__)
+app.config['SQLALCHEMY_DATABASE_URI'] = os.environ['MINDSDB_DB_CON']
```
StpMax, line 8: if you run that script manually, then how you can be sure MINDSDB_DB_CON is set up? this var could be setup from config file

StpMax (changes requested):

StpMax:
at this moment all actions executed manually. Shouldn't it be auto executed at mindsdb start?

mindsdb/interfaces/storage/db.py
ilia-tsyplenkov: No, it is in use. It is a part of flask_sqlalchemy.SQLAlchemy instance required interface

mindsdb/make_migrations.py
ilia-tsyplenkov: because mindsdb/__init__.py is executed anyway

ilia-tsyplenkov:
> at this moment all actions executed manually. Shouldn't it be auto executed at mindsdb start?

There is no needs to run migrations for each launch. Just need to add one new step `flask db upgrade` as a part of installation process. In case of development it is just a standard sequence: Make some db changes -> Create new migration -> Upgrade database

mindsdb/__init__.py
ilia-tsyplenkov: It is quite strange but with absolute imports any `flask db` command returns import error.

George3d6 (changes requested):
This seems nice in principle *but* will require testing and currently migrations can be done by just adding new columns manually to cloud.

Locally we can ask people to reinstall, since we don't support e.g. updates in local mode anwaya.

mindsdb/make_migrations.py
StpMax: when you run `flask db xxx` you will not have config in args, so it will not be read in _initi__.py

ilia-tsyplenkov:
> This seems nice in principle _but_ will require testing and currently migrations can be done by just adding new columns manually to cloud.
> 
> Locally we can ask people to reinstall, since we don't support e.g. updates in local mode anwaya.

In current implementation migration wouldn't be applied automatically in case of cloud

mindsdb/make_migrations.py
ilia-tsyplenkov: What is your suggestion here?

ilia-tsyplenkov: Well In case of development it is possible to setup `os.environ['MINDSDB_DB_CON']` manually. 
In case of customer launch with running migration as a part of mindsdb launch, everything will be ok. So I don't see here any serious issues

mindsdb/__main__.py
```diff
@@ -118,6 +119,11 @@ def close_api_gracefully(apis):
         } for api in api_arr
     }
 
+    if not is_cloud:
+        print('running migrations:')
+        p = subprocess.Popen(["python3 mindsdb/run_migrations.py"], shell=True)
```
StpMax, line 124: 1. you get access to DB before modifying it
2. this will work only if current work dir = mindsdb package dir.

StpMax (changes requested):

mindsdb/__main__.py
ilia-tsyplenkov: Thanks for the notes. Fixed

PR #1640, opened 2021-10-18
Replaced default value {} and [] as argument

Biki-das:
I Happen to found pieces of code using mutable as default arguement and we can evaluate a  None and them further do the conditioning


ZoranPandovski:
Closing this as discussed in the isssue.

PR #1709, opened 2021-10-27
Fixed errors in mindsdb-docs/docs/info.md which point to wrong file paths

ma1VAR3:
# Fixes
1. File links in https://github.com/mindsdb/mindsdb/blob/staging/docs/mindsdb-docs/docs/info.md were having incorrect file paths. Fixed those paths.
Example: 
```existing path```  : /deployment/docker
```fixed path```  : /deployment/docker.md
2. Removed ```extra``` alongside with ```disqus``` in https://github.com/mindsdb/mindsdb/blob/staging/docs/mindsdb-docs/mkdocs.yml#L25

### DIFF
--- a/docs/mindsdb-docs/docs/info.md
+++ b/docs/mindsdb-docs/docs/info.md
@@ -6,17 +6,16 @@ title: Getting started
 Follow the below steps to get up and running with MindsDB.
 
 ![Getting Started Image](assets/getting-started.png)
+
 #### Steps:
 
-1.  Deploy MindsDB: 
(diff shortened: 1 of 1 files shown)


ma1VAR3:
@ZoranPandovski but when you click on a link it takes you to a 404 page.
Example: https://github.com/mindsdb/mindsdb/blob/staging/deployment/docker
### Documentation:

mnds 1

### Link (for docker):

mnds 2

Actually just adding .md doesn't fix this behavior, the path needs to be modified a bit. Working on it, will open a pull request for the same.

PR #1757, opened 2021-11-06
docs(docker): add instructions for docker for mac

Buooy:
Fixes #

Currently the default docker for mac allocates 2GB of memory which is insufficient to run the mindsdb docker container. The process would fail with a cryptic "KILLED" message.

To rectify this, Docker for Mac users should bump their allocated memory to at least 4GB. 

This PR adds a note to users to do so before installation

mindsdb


ZoranPandovski (approved):
Thanks

PR #1760, opened 2021-11-09
For completeness sake add database parameter

OriPekelman:
The database name may not be the default `mindsdb` (and the parameter seems supported).


ZoranPandovski (approved):
Thanks

PR #1774, opened 2021-11-17
Fix 'rename predictor' endpoint

ilia-tsyplenkov:
Fixes 

`rename` endpoint returns next error:
```
"'ModelInterface' object has no attribute 'rename_model'"
```
Fix it by implementing this method

### DIFF
--- a/mindsdb/interfaces/model/model_controller.py
+++ b/mindsdb/interfaces/model/model_controller.py
@@ -364,6 +364,9 @@ class ModelController():
         db_p = db.session.query(db.Predictor).filter_by(company_id=company_id, name=old_name).first()
         db_p.name = new_name
         db.session.commit()
+        dbw = DatabaseWrapper(company_id)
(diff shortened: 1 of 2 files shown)


mindsdb/interfaces/model/model_controller.py
```diff
@@ -360,6 +360,11 @@ def delete_model(self, name, company_id: int):
 
         return 0
 
+    def rename_model(self, old_name, new_name, company_id: int):
+        db_p = db.session.query(db.Predictor).filter_by(company_id=company_id, name=old_name).first()
```
StpMax, line 364: you need 
```
dbwrapper..unregister_predictor(old_name)
dbwrapper..register_predictor(new_name)
```
and also rename model file in `fs_store`

StpMax (changes requested):

StpMax (approved):

PR #1777, opened 2021-11-18
fixed issue #1754 - Error when integration name contains space

### LINKED ISSUE #1754: Error when integration name contains space
If the user adds the space in the integration name we are not able to parse that for e.g `My SQL`

```sql
mysql> CREATE PREDICTOR home_rentals_model FROM `My SQL`(SELECT * from test_data.home_rentals)  PREDICT rental_price;
ERROR 1149 (HY000): Expected "PREDICT" (at char 45), (line:1, col:46)
mysql>
```

Yoda-Canada:
Fixes #1754 

  -update _get_integration function

---
- To see the specific tasks where the Asana app for GitHub is being used, see below:
  - https://app.asana.com/0/0/1201427993186043

### DIFF
--- a/mindsdb/interfaces/database/database.py
+++ b/mindsdb/interfaces/database/database.py
@@ -41,7 +41,8 @@ class DatabaseWrapper():
             logger.warning('Failed to integrate with database ' + db_alias + f', error: {e}')
 
     def _get_integration(self, db_alias):
-        integration = self.datasource_interface.get_db_integration(db_alias)
+        db_name = self.datasource_interface.get_db_integration(db_alias)
+        integration ={k: v.replace(" ", "") for k, v in db_name.items()}
         if integration:
             db_type = integration['type']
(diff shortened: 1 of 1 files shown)


StpMax:
Hi @Yoda-Canada , sorry for delay
Your changes affect only 'integration', ie when you work with MindsDB from your database (not via direct connection to MindsDB). In original issue error had raise in 'create predictor' query, which cant be used with such connection. I think this is parser error, probably it does not understand '`' symbols. If so, then fix should be somewhere here:
https://github.com/mindsdb/mindsdb/blob/staging/mindsdb/api/mysql/mysql_proxy/classes/sql_statement_parser.py#L336

Yoda-Canada:
@StpMax That's so kind of you. I just need a little more time. I'm almost there.

Yoda-Canada:
please refer to a new PR #1791.

PR #1887, opened 2022-01-21
Predictor parameters

George3d6:
When combined with the eponymous `mindsdb_sql` branch it allows for specifying arbitrary json-ai keys in the `USING` keyword of `CREATE PREDICTOR` as well as a special `models` key.

Example of allowed query:

```
CREATE PREDICTOR hrp32 FROM mydb (select * from test_data.home_rentals) PREDICT rental_price USING features='{"location": {"encoder": "CategoricalAutoEncoder(max_encoded_length=55)"}}', models='["LightGBM(stop_after=10, fit_on_dev=True)"]';
```


StpMax (approved):
Looks fine. I'll check how it works at monday.

StpMax (approved):

PR #1936, opened 2022-02-03
BYOM prototype

George3d6:
Solves #1941  and #1942


rajkumarraiyan:
1936

StpMax:
1.a is not a choice. We need columns names/type for proper mysql packets processing. For 'binary' mysql protocol we need that even before we can call 'model.predict'.
2. By the way: do wae want sometime in future add some kind of training progress bar? May be we need some hook in lightwood, which will return percent of progress in training (if it possible)?

StpMax (approved):
Looks well. Need only del `print` calls

PR #1939, opened 2022-02-04
[#1919] Update USING documentation

### LINKED ISSUE #1919: Update USING docs
The JSON syntax is not supported with USING statement. The [USING docs and example](https://github.com/mindsdb/mindsdb/blob/1c5c14de8617db550db313154e6ad42831145311/docs/mindsdb-docs/docs/sql/api/predictor.md#using-keyword) should be changed to follow the structure as

```sql
CREATE PREDICTOR predictor_name
FROM integration_name 
(SELECT column_name, column_name2 FROM table_name) as ds_name
PREDICT column_name as column_alias
USING ignore_columns="column_name";
```

tino097:
Fixes #1919

  - Use the correct syntax for the `USING` keyword in the examples


ZoranPandovski (approved):
Thanks @tino097

PR #2088, opened 2022-03-23
Migrate influxdb

pixpack:
Related to #1848  

  - Updated the influxDB python client for influxDB 2.0+
  - Migrated the influxDB datasource to mindsdb repository

This depends on #2060 so the PR will have to wait before the merge I think.


ZoranPandovski:
Thanks @pixpack, please check and follow the structure for MySQL handler https://github.com/mindsdb/mindsdb/tree/staging/mindsdb/integrations/mysql_handler. We are adding the documentation for this soon :raised_hands:

ZoranPandovski:
That looks like the wrong comment, this is still WIP so it will be improved for sure. Feel free to point out things like this so we can fix them and improve the documentation for this. Returning `{"OK": 200}` would be enough for now

pixpack:
I'm putting this on hold to finish work on the Couchbase integration which will probably be simpler.

The main issue that I'm currently facing with inlfux is that is using it's own query language called Flux. Since the new handler API is geared towards SQL with its funtions `get_tables()`, `join()` etc. I need some help to figure out a way to integrate this.

PR #2090, opened 2022-03-24
Modularized handlers v1

paxcema:
Modularized/encapsulated way of defining integrations to both databases and ML framworks. This is a working proof of concept, and includes the following inside `mindsdb/integrations`:

* Base classes: `BaseHandler`, `DatabaseHandler`, `PredictiveHandler`
* Example classes (one of each): `MySQLHandler`, `MlFlowHandler`
* Key-Value storage handlers: `StorageHandler`, `SqliteStorageHandler`, `RedisStorageHandler`

There are a lot of pending tasks, but this PR is going to be merged ASAP to get more contributors into writing other handlers. 

Some pending tasks:

* Dynamic dependency loader per each handler (e.g. redis only if the user has it setup)
* Better mlflow handler `check_status()` method
* MySQL handler to support more than just one DB
* Improve test coverage
* ~Adapt MySQL `join` to true signature (important as it is the reference example)~
* ~Tests are still not reformatted into actual unittests because we need to figure out how to add mlflow into the CI.~
* ~Agree on signature for `DBHandler.select_into`~

For reviewers: most of what's here has been agreed upon beforehand, so feel free to skim the PR.


mindsdb/integrations/mlflow_handler/requirements.txt
```diff
@@ -0,0 +1,2 @@
+mlflow
+mindsdb
```
ZoranPandovski, line 2: Maybe add `mindsdb` version >= to the current release to avoid using old versions.

mindsdb/integrations/mlflow_handler/mlflow/mlflow_handler.py
```diff
@@ -0,0 +1,273 @@
...
+
+        predictions = pd.DataFrame({'prediction': answer})
+        out = df.join(predictions)
+        return out
+
+
+if __name__ == '__main__':
+    # TODO: turn this into tests
+
+    if True:
```
ZoranPandovski, line 232: Do we need this?

mindsdb/integrations/mlflow_handler/mlflow/__about__.py
```diff
@@ -0,0 +1,9 @@
+__title__ = 'MindsDB MLflow connector'
+__package_name__ = 'mindsdb'
```
ZoranPandovski, line 2: Should we use `mindsdb` as the package name if this can be published as a separate package?

mindsdb/integrations/mysql_handler/mysql_handler/mysql_handler.py
```diff
@@ -0,0 +1,172 @@
...
+        if not self.check_status():
+            self.connect()
+        try:
+            with closing(self.connection) as con:
+                cur = con.cursor(dictionary=True, buffered=True)
+                cur.execute(f"USE {self.database};")
+                cur.execute(query_str)
+                res = True
+                try:
+                    res = cur.fetchall()
```
ZoranPandovski, line 66: Should we continue using `pandas.read_sql` per each datasource and always return dataframes? Or even maybe implement that in the `Database.run_native_query`  and just accept only connection? Note important, maybe we can do that in another PR(just worth mentioning)

ZoranPandovski (changes requested):

mindsdb/integrations/mlflow_handler/mlflow/mlflow_handler.py
paxcema: whoops, no, leftover

mindsdb/integrations/mysql_handler/mysql_handler/mysql_handler.py
paxcema: I'm don't think we've settled on the output type for `run_native_query` yet.  Also inside this method, any handler could use `pd.read_sql` internally, but the input should be a raw query string.

Do you think I should change it to use `pd.read_sql` in this case?

ZoranPandovski: We can start with the current impl and then decide :raised_hands:

ZoranPandovski (approved):

mindsdb/integrations/mlflow_handler/mlflow/mlflow_handler.py
```diff
@@ -0,0 +1,222 @@
...
+        self.mlflow_server_url = None
+        self.mlflow_server_path = None
+        self.connection = None
+        self.storage = None
+        self.parser = parse_sql
+        self.dialect = 'mindsdb'
+
+    def connect(self, **kwargs) -> Dict[str, int]:
+        """ Connect to the mlflow process using MlflowClient class. """  # noqa
+        print(kwargs)
```
StpMax, line 46: to del

mindsdb/integrations/mysql_handler/mysql_handler/mysql_handler.py
```diff
@@ -0,0 +1,116 @@
...
+        self.host = kwargs.get('host')
+        self.port = kwargs.get('port')
+        self.user = kwargs.get('user')
+        self.database = kwargs.get('database')  # todo: may want a method to change active DB
+        self.password = kwargs.get('password')
+        self.ssl = kwargs.get('ssl')
+        self.ssl_ca = kwargs.get('ssl_ca')
+        self.ssl_cert = kwargs.get('ssl_cert')
+        self.ssl_key = kwargs.get('ssl_key')
+        self.connect()
```
StpMax, line 29: connection must be closed at the end of usage

StpMax (changes requested):

PR #2109, opened 2022-04-04
Lightwood handler

paxcema:
This PR adds a handler for the Lightwood AutoML package.

Handler structure:
* `lightwood_handler.py`: the actual class plus (local) tests in the `if __name__ == '__main__'` section, which shall be moved to actual cloud CI post-merge.
* `utils.py`: general utility methods for packing/unpacking JsonAI objects, adding aliases to columns, and gathering data.
* `join_utils.py`: methods to build input `DataFrames` when predicting, via `mindsdb_sql` select statements.
* `ts_utils.py`: same as in `join_utils`, but this is logic specific to time series predictors.

Most of the last two files is logic incorporated from the `mindsdb_sql` package that we should remove from it once the refactor is live.

I've tried to leave comments with `TODO` tags on all things that are pending but we may want to figure out as we go about deploying the entire refactor, instead of trying it out in isolation.

<details>
  <summary>WIP notes:</summary>
  Pending for this PR:

  * ~~[ ] Implement non-KV storage handler and use that instead~~
  
  Future work:

  * ~~[ ] Base class to enable planner step execution (i.e. current `execute_query` method)~~
  * [x] Move lightwood specific logic from mindsdb_sql to this handler


mindsdbadmin:

PR #2116, opened 2022-04-06
Offline mode GUI

setohe0909:
Fixes #

  -

mindsdb


StpMax (approved):
All fine, but GUI itself is not ready: it require auth, but should not

PR #2150, opened 2022-04-18
Add PostgreSQL handler

ZoranPandovski:
This PR adds the initial version of the PostgreSQL handler with methods for connecting, checking status, running queries, and listing tables, views, and columns.

### DIFF
--- a/mindsdb/integrations/postgres_handler/postgres_handler.py
+++ b/mindsdb/integrations/postgres_handler/postgres_handler.py
@@ -12,19 +12,17 @@ class PostgresHandler(DatabaseHandler):
     def __init__(self, name, **kwargs):
         super().__init__(name)
         self.parser = parse_sql
+        self.connection_args = kwargs
         self.dialect = 'postgresql'
         self.database = kwargs.get('database')
-        self.host = kwargs.get('host')
-        self.port = kwargs.get('port')
-        self.user = kwargs.get('user')
-        self.password = kwargs.get('password')
 
     def __connect(self):
         """
         Handles the connection to a PostgreSQL database insance.
         """
         #TODO: Check psycopg_pool
-        connection = psycopg.connect(f'host={self.host} port={self.port} dbname={self.database} user={self.user} password={self.password}', connect_timeout=10)
+        connection_args['dbname'] = self.database
+        connection = psycopg.connect(**self.connection_args, connect_timeout=10)
         return connection
 
     def check_status(self):
--- a/mindsdb/integrations/postgres_handler/postgres_handler.py
+++ b/mindsdb/integrations/postgres_handler/postgres_handler.py
@@ -61,3 +61,30 @@ class PostgresHandler(DatabaseHandler):
                     pass
         return res
     
+    def get_tables(self):
+        """
+        List all tabels in PostgreSQL without the system tables information_schema and pg_catalog
+        """
+        query = f"SELECT * FROM information_schema.tables WHERE \
+                 table_schema NOT IN ('information_schema', 'pg_catalog')"
+        res = self.run_native_query(query)
+        return res
+    
+    def get_views(self):
+        """
+        List all views in PostgreSQL without the system views information_schema and pg_catalog
+        """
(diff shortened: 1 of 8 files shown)


mindsdbadmin:

mindsdb/integrations/postgres_handler/postgres_handler.py
```diff
@@ -0,0 +1,101 @@
...
+        self.port = kwargs.get('port')
+        self.user = kwargs.get('user')
+        self.password = kwargs.get('password')
+
+    def __connect(self):
+        """
+        Handles the connection to a PostgreSQL database insance.
+        """
+        #TODO: Check psycopg_pool
+        connection = psycopg.connect(f'host={self.host} port={self.port} dbname={self.database} user={self.user} password={self.password}', connect_timeout=10)
```
StpMax, line 27: May be we should save all argument to dict and use it as dict? So we will be able use any connection parameter (as i remember someone already asked about ssl mode https://www.postgresql.org/docs/current/libpq-connect.html#LIBPQ-PARAMKEYWORDS). 
```
def __init__(self, name, **kwargs):
    self.connection_args = kwargs
    self.connection_args['dbname'] = self.connection_args['database']
    del self.connection_args['database']    $ or map it in __connect ?
    
def __connect(self):
    connection = psycopg.connect(**self.connection_args, connect_timeout=10)
```
will be need map 'database' to 'dbname

mindsdb/integrations/postgres_handler/postgres_handler.py
```diff
@@ -0,0 +1,101 @@
...
+        try:   
+            con = self.__connect()
+            with closing(con) as con:
+                with con.cursor() as cur:
+                    cur.execute('select 1;')
+            connected = True
+        except Exception as e:
+            log.error(f'Error connecting to PostgreSQL {self.database} on {self.host}!')
+            pass
+        return connected
```
StpMax, line 44: will be good return here status and reason. So we will be able show to user why connection is not working, for example:
```
 {
     'success': True
 }
 {
     'success': False,
     'error': 'password is wrong'
 } 
```

mindsdb/integrations/postgres_handler/postgres_handler.py
```diff
@@ -0,0 +1,101 @@
...
+    
+    def run_native_query(self, query_str):
```
StpMax, line 46: may be call methods just 
`def native_query(self, query: str)`
and
`def query(self, query: ast)`
?

mindsdb/integrations/postgres_handler/postgres_handler.py
```diff
@@ -0,0 +1,101 @@
...
+        with closing(con) as con:
+            with con.cursor() as cur:
+                res = True
+                try:
+                    cur.execute(query_str)
+                    res = cur.fetchall()
+                except psycopg.Error as e:
+                    log.error(f'Error running {query_str} on {self.database}!')
+                    pass
+        return res
```
StpMax, line 62: May define response same as here: https://github.com/mindsdb/mindsdb/blob/staging/mindsdb/api/http/namespaces/sql.py#L12
So we can run queriess with and without result useing one method.
Also, will be good to output `column_types` for `table` response

mindsdb/integrations/postgres_handler/postgres_handler.py
```diff
@@ -0,0 +1,101 @@
...
+    def describe_table(self, table_name):
+        """
+        List names and data types about the table coulmns
+        """
+        query = f"SELECT table_name, column_name, data_type FROM \
+              information_schema.columns WHERE table_name='{table_name}';"
+        result = self.run_native_query(query)
+        return result
+
+    def select_query(self, targets, from_stmt, where_stmt):
```
StpMax, line 90: i think better will be have only one argument with ast. So you can do:
```
def select_query(self, query: ast):
    renderer = SqlalchemyRender('postgres')
    query_str = renderer.get_string(query, with_failback=True)
    return self.native_query(query_str)
````

StpMax (changes requested):

PR #2178, opened 2022-05-03
House sales forecasting tutorial

paxcema:
Adds a short tutorial for multivariate time series.


mindsdbadmin:

docs/mindsdb-docs/docs/sql/tutorials/house-sales-forecasting.md
```diff
@@ -0,0 +1,61 @@
...
+## Data and setup
+
+The dataset we will use is the pre-processed version of the "House Property Sales" [Kaggle competition](https://www.kaggle.com/datasets/htagholdings/property-sales), in particular, the "ma_lga_12345.csv" file, which tracks quarterly moving averages of house sales aggregated by type and amount of bedrooms in each listing.
+
+Make sure you have access to a working MindsDB installation (either local or via cloud.mindsdb.com), and either load it into a table in your database of choice or upload the file directly to the special `FILES` datasource (via SQL or GUI). Once you've done this, proceed to the next step. For the rest of the tutorial we'll assume you've opted for the latter option:
+
+```
+SHOW TABLES FROM files;
+
+SELECT * FROM files.HR_MA LIMIT 10;
```
ZoranPandovski, line 16: Maybe we can mention that HR_MA is the name of the uploaded dataset?

ZoranPandovski (changes requested):

docs/mindsdb-docs/docs/sql/tutorials/house-sales-forecasting.md
paxcema: Oh, good point yes, I will mention it!

ZoranPandovski:
@paxcema Can you please also add link in the navigation under tutorials/timeseries?

ZoranPandovski (changes requested):
@paxcema  Can you please also add link tot the tutorial under naviagation .yml file

ZoranPandovski (approved):

Dav1dF (commented):
@paxcema can we illustrate this by creating the single time-series that you mentioned 
```
Now, we can specify that we want to forecast the MA column, which is a moving average of the historical median price for house sales. However, looking at the data you can see several entries for the same date, depending on two factors: how many bedrooms the properties had, and whether properties were "houses" or "units". This means that we can have up to ten different groupings here (although, if you do some digging, you will find we only actually have seven of the possible ten combinations in practice).
MindsDB makes it simple so that we don't need to repeat the predictor creation process for every group there is. Instead, we can just group for both columns and the predictor will learn from all series and enable forecasts for all of them! The command for this is.
```

PR #2205, opened 2022-05-11
Proper columns names for results of internal queries

### LINKED ISSUE #1969: Wrong column names if `select null`
If do:
```
select null, null, null from information_schema.tables limit 1;
```
then response will be:
```
+------+--------+--------+
| None | None_2 | None_3 |
+------+--------+--------+
| nan  | nan    | nan    |
+------+--------+--------+

```
In real mysql all 3 column names will be 'NULL' (in upper case)

### LINKED ISSUE #1977: Result of query with same column names with return different column names
Issue exists for any 'internal' tables. If query multiple columns with same name:
```
select 1 as "a", 1 as "a" from information_schema.tables limit 1;
```
then result will be:
```
+------+------+
| a    | a_2  |
+------+------+
| 1    | 1    |
+------+------+
```

StpMax:
close #1969
close #1977


mindsdbadmin:

PR #2239, opened 2022-05-20
Updated Create Predictor Statement

Max-Miller03:
Fixes #

  -

mindsdb


ZoranPandovski (changes requested):
The change should be in a section 5. with create predictor

Max-Miller03:
> The change should be in a section 5. with create predictor

Got it. Will fix it now

ZoranPandovski:
Resolved in #2243

PR #2255, opened 2022-05-26
Fix/docs join

Ricram2:
Major refactor JOIN doc. 

https://www.loom.com/share/e6fbba77a778494a8a85ec38cb76cf24

image

Minor fixes:

image

image

image

  -

mindsdb

### DIFF
--- a/docs/mindsdb-docs/docs/sql/api/join.md
+++ b/docs/mindsdb-docs/docs/sql/api/join.md
@@ -1,19 +1,72 @@
-# JOIN clause
+# `#!sql JOIN` Statement
 
-The `JOIN` clause is used to combine rows from the database table and the model table on a related column. The basic syntax for joining from the data table and model is:
+## Description
+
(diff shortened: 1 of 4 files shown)


ZoranPandovski (approved):

docs/mindsdb-docs/docs/sql/api/join.md
```diff
@@ -1,19 +1,101 @@
...
+
+```sql
+CREATE PREDICTOR 
+  mindsdb.house_sales_model
+FROM example_db
+  (SELECT * FROM demo_data.house_sales)
+PREDICT MA
+ORDER BY saledate
+GROUP BY bedrooms, type
+-- as the target is quarterly, we will look back two years to forecast the next one
```
ZoranPandovski, line 86: should we remove this since it will be confusing?

ZoranPandovski (changes requested):

PR #1752, opened 2021-11-03
Use parametrized Dockerfile for building images

abitrolly:
This removes generation step and makes the setup more readable.

PR #2295, opened 2022-06-04
Improving readability of Docker files

abitrolly:
Another attempt to use Dockerfile ARG instead of generating docker files manually #1752.

This moves config files out of inline content for easier comparison, uses ARG VERSION argument to parameterize build, and documents build and release process.


ZoranPandovski:
recheckcla

ZoranPandovski (approved):
Thanks, looks great

ZoranPandovski:
recheckcla

PR #2465, opened 2022-07-10
fixed bug in arguments passed to log message

MinuraPunchihewa:
Made a quick fix to one of the arguments passed into a f string that makes up an error message.


ZoranPandovski (approved):
Thanks

PR #2525, opened 2022-07-19
Create table: raise exception if column is doubled

ea-rus:
Create table step:
- raise exception if column is doubled


ZoranPandovski:
@ea-rus Do we need this? If not, let's close the PR.

PR #2573, opened 2022-07-26
Lightwood handler improvements

StpMax:


mindsdb/api/mysql/mysql_proxy/datahub/datanodes/mindsdb_datanode.py
```diff
@@ -1,17 +1,12 @@
...
 import pandas as pd
 import numpy as np
 
 from mindsdb.api.mysql.mysql_proxy.datahub.datanodes.datanode import DataNode
 from mindsdb.api.mysql.mysql_proxy.utilities.sql import query_df
-from mindsdb.api.mysql.mysql_proxy.utilities.functions import get_column_in_case
-from mindsdb.utilities.functions import cast_row_types
 from mindsdb.utilities.config import Config
-from mindsdb.api.mysql.mysql_proxy.utilities import SqlApiException
+from mindsdb.api.mysql.mysql_proxy.utilities.lightwood_dtype import dtype
```
paxcema, line 9: This is still being used in `_get_model_columns`, but we should make it so that it calls the respective handler method instead, because having Lightwood `dtypes` here goes against mutual independence.

mindsdb/api/mysql/mysql_proxy/executor/executor_commands.py
```diff
@@ -1287,55 +1188,12 @@ def insert_predictor_answer(self, insert):
...
-        integration_id = integration_meta.get('id')
-        fetch_data_query = insert['select_data_query']
-
-        model_interface.learn(
-            insert['name'], ds_data_df, insert['predict'], integration_id=integration_id,
-            fetch_data_query=fetch_data_query, kwargs=kwargs, user_class=self.session.user_class
+        return ExecuteAnswer(
+            ANSWER_TYPE.ERROR,
+            error_code=0,
+            error_message='At the moment insert into predictors talbe is not supported'
```
paxcema, line 1194: typo: `talbe` -> `table`

mindsdb/api/mysql/mysql_proxy/mysql_proxy.py
```diff
@@ -35,7 +34,7 @@
 from mindsdb.api.mysql.mysql_proxy.classes.server_capabilities import server_capabilities
 from mindsdb.api.mysql.mysql_proxy.classes.sql_statement_parser import SqlStatementParser
 from mindsdb.api.mysql.mysql_proxy.utilities import log
-
+from mindsdb.api.mysql.mysql_proxy.utilities.lightwood_dtype import dtype
```
paxcema, line 37: As in `mindsdb_datanode`, we should remove this as a requirement. It's being used in `to_mysql_records`.

Maybe a solution is to enforce on all predictive handlers to always output `numpy` types.

mindsdb/interfaces/database/integrations.py
```diff
@@ -234,6 +235,16 @@ def create_handler(self, name: str = None, handler_type: str = None,
                 ViewController(),
                 company_id=company_id
             )
+        elif handler_type == 'lightwood':
```
paxcema, line 238: Probably better to do a generic "predictive handler" check and enforce all of them (lightwood included) to conform to the API defined by `ModelInterface()`. Not sure how to pass the class type instead of the handler name, though.

mindsdb/interfaces/database/integrations.py
```diff
@@ -246,7 +257,16 @@ def get_handler(self, name, company_id=None, case_sensitive=False):
                 & (func.lower(Integration.name) == func.lower(name))
             ).first()
         if integration_record is None:
-            raise Exception(f'Unknown integration: {name}')
+            if name == 'lightwood':
```
paxcema, line 260: Pending: make this check generic to support other ML backends (e.g. MLflow)

mindsdb/integrations/handlers/lightwood_handler/lightwood_handler/lightwood_handler.py
```diff
@@ -92,103 +129,506 @@ def check_connection(self) -> Dict[str, int]:
...
+        if predictor_record is None:
+            return Response(
+                RESPONSE_TYPE.ERROR,
+                error_message=f"Error: model '{table_name}' does not exists!"
+            )
+
+        data = []
+        if predictor_record.dtype_dict is not None:
+            for key, value in predictor_record.dtype_dict.items():
+                data.appen(key, value)
```
paxcema, line 160: typo: `append`

mindsdb/api/mysql/mysql_proxy/utilities/lightwood_dtype.py
```diff
@@ -0,0 +1,48 @@
+class dtype:
```
paxcema, line 1: See the above comments, I think this shouldn't be used. Else, we are forcing Lightwood philosophy on other handlers that may come after. 

The solution may require additional discussions, but I think it is something like guaranteeing predictive handlers will return data with the "correct"  data types depending on what data handler called them.

paxcema:
Another thing to consider @StpMax:

We need to decide whether the `db.Predictor` class should be Lightwood-specific or not. I'm inclined to make it specific because things like the JsonAI are unique to the framework.

Independent of the above, we should move that class to the `integrations` folder and, potentially, rewrite it into a `StorageHandler`. 

Finally, methods `_get_model` and `_get_model_info` are currently broken, but they should be fixed once we settle on the above.

PR #2602, opened 2022-07-29
DynamoDB Integration

MinuraPunchihewa:
I have implemented the new integration for DynamoDB. After some initial testing, it looks to be working as expected. Please do let me know if there is anything that I have missed or if there are any bugs.


ZoranPandovski (approved):
Looks great, thanks

PR #2551, opened 2022-07-21
[ENH] Lightwood handler: improved testing

paxcema:
WIP

PR #2652, opened 2022-08-08
[LW Handler] Add more TS tests

paxcema:
Supersedes #2551 

- Fixes couple of things in the handler to fix non-TS tests
- Adds more TS test coverage

### DIFF
--- a/mindsdb/integrations/handlers/lightwood_handler/setup.py
+++ b/mindsdb/integrations/handlers/lightwood_handler/setup.py
@@ -1,25 +1,32 @@
-from setuptools import setup, find_packages
+import setuptools
+import os
 
-pkgs, new_links = install_deps()
+about = {}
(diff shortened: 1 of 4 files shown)


mindsdb/integrations/handlers/lightwood_handler/setup.py
```diff
@@ -1,25 +1,32 @@
...
+about = {}
+with open("lightwood_handler/__about__.py") as fp:
+    exec(fp.read(), about)
 
-setup(
+with open('requirements.txt') as req_file:
+    requirements = [req.strip() for req in req_file.read().splitlines()]
+
+# TODO: relative import okay?
+with open('../../../../requirements_test.txt', 'r') as req_test_file:
```
StpMax, line 12: may be better will be hardcode that dependencies. Or avoid it, since it use only for tests

StpMax (approved):

PR #2664, opened 2022-08-11
route to analyse data

StpMax:
POST /analysis/data
Request body is JSON:
```
{
   "columns_names": ["a", "b"] // list of column names
   "data": [[1,2], [3,4]]  // array of rows
}
```
Response is also JSON:
```
{
   "analysis": {object},
   "timestamp": 12345
}
```


StpMax:
`column_names` instead of `columns_names`

PR #2670, opened 2022-08-12
Predictor cache

ea-rus:
Changes:
- Added module for caching
  -  Available cache types: file, redis
- cache predictor results in ApplyPredictorStep 

Docs:
 https://github.com/mindsdb/mindsdb/blob/predictor-cache/mindsdb/utilities/cache.py#L2

#2411


mindsdb/utilities/cache.py
```diff
@@ -1,121 +1,230 @@
...
+
 class BaseCache(ABC):
-    def __init__(self):
+    def __init__(self, max_size=None, serializer=None):
         self.config = Config()
+        if max_size is None:
+            max_size = self.config["cache"].get("max_size", 50)
+        self.max_size = max_size
+        if serializer is None:
+            self.serializer = self.config["cache"].get('serializer', dill)
```
StpMax, line 92: config is json file - there may be only text value 'dill' in it

StpMax:
Why not to use build in method `df.to_pickle` ? I did few tests, looks like it works at least x2 faster, then dump-to-json.
Also, may be for hash use another method? https://stackoverflow.com/a/62754084

StpMax (changes requested):

mindsdb/utilities/cache.py
ea-rus: Changed to string

ea-rus:
`Why not to use build in method df.to_pickle`
added set_df and get_df methods to cache dataframes

`https://stackoverflow.com/a/62754084`
pd.util.hash_pandas_object(df) unable to hash list and dict types.

StpMax (approved):

PR #2736, opened 2022-08-19
Add Mongo API docs

ZoranPandovski:
This PR adds the initial implementation of the Mongo API docs as:

- Add new Mongo API menu navigation
- Add #2730 
- Add #2731 
- Add #2732 
- Add #2733 
- Add #2734 
- Add #2735 

> Note: For the new version, we need to add docs for USING parameter, train TS model, query TS model, and connect to MongoAPI. I will open new issues for this.


ZoranPandovski:
As discussed with @Ricram2 we will merge this and then work on a version 2

PR #2871, opened 2022-09-05
Support Union / Union all

ea-rus:
- Added support UnionStep #2852
- no cache if predictor returns nothing
- unit tests fixes
- mongo insert predictor: don't use model_controller and integration_controller outside executor
- change message for drop non existing database
- enabled unit tests


mindsdb/api/mysql/mysql_proxy/classes/sql_query.py
```diff
@@ -770,11 +874,39 @@ def execute_step(self, step, steps_data):
...
+            # count of columns have to match
+            if len(left_result.columns) != len(right_result.columns):
+                raise ErSqlWrongArguments(
+                    f'UNION columns count mismatch: {len(left_result.columns)} != {len(right_result.columns)} ')
+
+            # types have to match
+            for i, left_col in enumerate(left_result.columns):
+                right_col = right_result.columns[i]
+                type1, type2 = left_col.type, right_col.type
+                if type1 is not None and type2 is not None:
```
paxcema, line 892: Potentially change to:

```python
if isinstance(type1, str) and isinstance(type2, str):
```

paxcema: But maybe this is a bug in the column type detection, if so, the above would be the wrong fix.

PR #2880, opened 2022-09-07
Union/all: cast datetime predictions

paxcema:


paxcema:
Closing, solved in #2871

PR #2927, opened 2022-09-15
Null rows in TS prediction result

ea-rus:
- Added filtration of prediction results. Filter is step.output_time_filter from planner
  - if date in filter is string it converts to datetime
- changed logic of JoinStep. 
  - if it is simple join and predictor is one of tables: keep only predictor rows and join table rows to in (left join if predictor is from left and right if it is from right)
  - else keep original type of join from select

[#2501](https://github.com/mindsdb/mindsdb/issues/2501)
[#2435](https://github.com/mindsdb/mindsdb/issues/2435)
[#2264](https://github.com/mindsdb/mindsdb/issues/2264)

### DIFF
--- a/mindsdb/api/mysql/mysql_proxy/classes/sql_query.py
+++ b/mindsdb/api/mysql/mysql_proxy/classes/sql_query.py
@@ -12,7 +12,9 @@
 import re
 from collections import OrderedDict, defaultdict
 import time
+import datetime as dt
 
+import dateinfer
 import duckdb
 import pandas as pd
 import numpy as np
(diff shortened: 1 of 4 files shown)


mindsdb/api/mysql/mysql_proxy/classes/sql_query.py
```diff
@@ -12,7 +12,9 @@
 import re
 from collections import OrderedDict, defaultdict
 import time
+import datetime as dt
 
+import dateinfer
```
StpMax, line 17: need add to to requirements.txt

StpMax (changes requested):

PR #2992, opened 2022-09-25
use trino dialect from pyhive, add  to database parameters and update…

bachng2017:
Fixes #

  1. use pyhive[trino] dialect for trino handler
  2. add parameter `with` to configure a global WITH clause (properties) for trino  tables (experimental)
  3. update related document

for 1 and 2, it worked in our lab but any recommend for implementations are appreciated


ZoranPandovski (approved):
Thanks looks good, one thing is that we will need to write a test for the substring part just to cover  more scenarios, but that can be done in separate PR

ZoranPandovski (approved):
Thanks looks good, one thing is that we will need to write a test for the substring part just to cover  more scenarios, but that can be done in separate PR

PR #3042, opened 2022-09-30
[i3039] Reconstruct RayServe Handler.

rfsiten:
Hey,  the RayServe in the latest version has been removed. I've discussed it with you team and know that you want to reconstruct it.

Now I've implemented it. In fact, I think there two several ways to implement it, I try to write a full ml handler at first, nearly finished, but I found it too complex, maybe result in some problems in stability. So I implement it in a light weight way, which only a few codes modified and it has been tested. 

Usage:

**create predictor**
```
CREATE PREDICTOR ray_serve_pred1
FROM test_mindsdb_proxy
   (SELECT dayofweek, hour, minute, cmp_mean_norm, anomaly_tag FROM trend_period_noisy_with_features_csv where column1 < 6000 and column1 > 60) 
PREDICT anomaly_tag
USING
    format='rayserve',
    url.train = 'http://127.0.0.1:8000/my_model/train',
    url.predict = 'http://127.0.0.1:8000/my_model/predict';
```
 
**use predictor**
```
SELECT c.dayofweek, c.hour, c.minute, c.cmp_mean_norm, p.anomaly_tag
FROM test_mindsdb_proxy.trend_period_noisy_with_features_csv c
join ray_serve_pred1 p
where c.column1 > 8000;   
```

### DIFF
--- a/mindsdb/integrations/handlers/rayserve_handler/rayserve_handler.py
+++ b/mindsdb/integrations/handlers/rayserve_handler/rayserve_handler.py
@@ -1,0 +1,101 @@
+import base64
+import json
+import pickle
+from typing import Optional, Dict
+
+import pandas as pd
+import requests
+
+from mindsdb.integrations.libs.base import BaseMLEngine
+from mindsdb.utilities import log
+from mindsdb.interfaces.storage import db
+
+
+class RayServeHandler(BaseMLEngine):
+    name = 'rayserve'
+
+    ARG_TARGET = "target"
+    ARG_COLUMN_SEQUENCE = "column_sequence"
+    ARG_TRAIN_URL = "train_url"
+    ARG_PREDICT_URL = "predict_url"
+    ARG_PREDICTOR_ID = "predictor_id"
+    ARG_PREDICTOR_NAME = "predictor_name"
+
+    KWARGS_DF = "df"
+
+    PERSIST_ARGS_KEY_IN_JSON_STORAGE = "rayserve_args"
+
+    def create(self, target, args=None, **kwargs):
+        # prepare arguments
+        using_args = args.get("using", dict())
+        train_url = using_args.get("train_url", None)
+        predict_url = using_args.get("predict_url", None)
+        if train_url is None:
+            raise Exception("Missing required argument: using.train_url")
+        if predict_url is None:
+            raise Exception("Missing required argument: using.predict_url")
+
+        predictor_id = self.model_storage.predictor_id
+        predictor_name = db.Predictor.query.get(predictor_id).name
+        # prepare args to serialize
+        serialize_args = dict()
+        serialize_args[self.ARG_TARGET] = target
+        serialize_args[self.ARG_TRAIN_URL] = train_url
+        serialize_args[self.ARG_PREDICT_URL] = predict_url
+        serialize_args[self.ARG_PREDICTOR_ID] = predictor_id
+        serialize_args[self.ARG_PREDICTOR_NAME] = predictor_name
+        # check df
+        df: pd.DataFrame = kwargs.get(self.KWARGS_DF, None)
+        if df is None:
+            raise Exception("missing required key in args: " + self.KWARGS_DF)
+        else:
+            column_sequence = sorted(list(df.columns.values))
+            df = df[column_sequence]
+            serialize_args[self.ARG_COLUMN_SEQUENCE] = column_sequence
+
+        try:
+            resp_status_code = ""
+            log.logger.info(F"Training model by RayServe, target: {target}, "
+                            F"train_url: {train_url}, predict_url: {predict_url}")
+            encoded_df = base64.encodebytes(pickle.dumps(df))
+            resp = requests.post(train_url, json={'df': encoded_df, 'target': target,
+                                                  "predictor_id": predictor_id, "predictor_name": predictor_name})
+            resp_status_code = resp.status_code
+            assert resp.status_code == 200
+            log.logger.info(F"Training model by RayServe success.")
+        except Exception as e:
+            raise Exception("Training failed: " + repr(e) + ", status_code: ", str(resp_status_code))
+        self.model_storage.json_set(self.PERSIST_ARGS_KEY_IN_JSON_STORAGE, serialize_args)
+        log.logger.info("Model args saved.")
+
+    def predict(self, df: pd.DataFrame, args: Optional[Dict] = None) -> pd.DataFrame:
+        # read model args from storage
+        deserialize_args = self.model_storage.json_get(self.PERSIST_ARGS_KEY_IN_JSON_STORAGE)
+
+        # resolve args
+        target = deserialize_args[self.ARG_TARGET]
+        feature_column_sequence = list(deserialize_args[self.ARG_COLUMN_SEQUENCE])
+        if target in feature_column_sequence:
+            feature_column_sequence.remove(target)
+        predict_url = deserialize_args[self.ARG_PREDICT_URL]
+        predictor_id = deserialize_args[self.ARG_PREDICTOR_ID]
+        predictor_name = deserialize_args[self.ARG_PREDICTOR_NAME]
+
+        df_send = df[feature_column_sequence]
+        encoded_df = base64.encodebytes(pickle.dumps(df_send))
+        resp = requests.post(predict_url, json={'df': encoded_df, 'target': target,
+                                                "predictor_id": predictor_id, "predictor_name": predictor_name})
+        pred_result = json.loads(resp.content)
+        if target not in pred_result:
+            raise Exception(F"Target column doesn't exist: {target}")
+        if len(pred_result[target]) != len(df_send):
+            raise Exception(F"Line number of pre diction doesn't equal with the one of input.")
+        rt = df.copy(deep=True)
+        pred_df = pd.DataFrame(index=rt.index, data=pred_result)
+        if target in set(rt.columns):
+            rt.drop(columns=[target], inplace=True)
+        rt = rt.join(pred_df, how="left")
+        return rt
+
+if __name__=="__main__":
+    print()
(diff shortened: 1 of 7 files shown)


paxcema:
Hi @rfsiten! This is great work and feedback. We agree writing full ML handlers is complex right now, so we're currently simplifying the interface (and abstracting a lot of the boilerplate). It should be ready by the end of the week, I think. Once that's done, we can revisit your implementation so that it complies with the streamlined new design.

paxcema:
We've already synced via Slack @rfsiten, but just to reiterate that #3155 has been merged into the `staging` branch, so you can now take a look at the new `BaseMLEngine`  abstraction in `integrations/libs/base` to refresh this great piece of work.

ZoranPandovski:
@rfsiten Let's resolve this :) I think we only need to do the changes mentioned from @paxcema

rfsiten:
> @rfsiten Let's resolve this :) I think we only need to do the changes mentioned from @paxcema

ok , no problem, I will try to finish it next week.

by the way,  another pr (https://github.com/mindsdb/mindsdb/pull/3650) of mine is also need to be review, I will fix some problem according to your recent change in staging and ask for your help of review... in about one~two days, thanks a lot.
@ZoranPandovski @paxcema

rfsiten:
@paxcema  @ZoranPandovski hey, the new implement of RayServeHandler has submitted, please review it, thanks.

mindsdb/integrations/handlers/rayserve_handler/requirements.txt
```diff
@@ -0,0 +1,5 @@
+mindsdb>=22.6.2.1
```
paxcema, line 1: Let's bring the first two requirements up to the latest versions, because we need the `BaseMLEngineExec` logic, or it won't work.

mindsdb/integrations/handlers/rayserve_handler/requirements.txt
```diff
@@ -0,0 +1,5 @@
...
+mindsdb_sql >= 0.4.0
+pandas <=1.4.3
+
+salesforce-merlion>=1.2.0,<=1.3.1
```
paxcema, line 5: This is a leftover, right? Please remove.

tests/unit/test_rayserve_handler.py
```diff
@@ -0,0 +1,149 @@
...
+    def setup_class(cls):
+        BaseExecutorTest.setup_class(cls)
+        processor = multiprocessing.Process(target=app.run, kwargs={"debug": True,
+                                                                    "host": "0.0.0.0", "port": mock_server_port})
+        processor.daemon = True
+        processor.start()
+        i = 0
+        while i < 30:
+            try:
+                resp = requests.post(f"http://127.0.0.1:{mock_server_port}/ready")
```
paxcema, line 72: May be better to increase the sleep() time and wait just once at the start to avoid sending that many requests? This approach seems a bit more intensive than it should, imho 😛

tests/unit/test_rayserve_handler.py
```diff
@@ -0,0 +1,149 @@
...
+        assert "col1" == ret.columns[0].name, "prediction result missing column col1"
+        assert "col2" == ret.columns[1].name, "prediction result missing column col2"
+        assert "label" == ret.columns[2].name, "prediction result missing column label"
+        assert [2, 3, 1] == ret.data[0], "wrong prediction result line 0"
+        assert [3, 2, 1] == ret.data[1], "wrong prediction result line 1"
+        assert [5, 6, 1] == ret.data[2], "wrong prediction result line 2"
+
+    def wait_training(self, model_name):
+        # wait
+        done = False
```
paxcema, line 134: Same as in setup_class(), maybe just wait once and for longer.

mindsdb/integrations/handlers/rayserve_handler/rayserve_handler.py
```diff
@@ -0,0 +1,101 @@
...
+        serialize_args[self.ARG_TRAIN_URL] = train_url
+        serialize_args[self.ARG_PREDICT_URL] = predict_url
+        serialize_args[self.ARG_PREDICTOR_ID] = predictor_id
+        serialize_args[self.ARG_PREDICTOR_NAME] = predictor_name
+        # check df
+        df: pd.DataFrame = kwargs.get(self.KWARGS_DF, None)
+        if df is None:
+            raise Exception("missing required key in args: " + self.KWARGS_DF)
+        else:
+            column_sequence = sorted(list(df.columns.values))
```
paxcema, line 52: Out of curiosity: why do we need to sort the columns here?

mindsdb/integrations/handlers/rayserve_handler/rayserve_handler.py
```diff
@@ -0,0 +1,101 @@
...
+            df = df[column_sequence]
+            serialize_args[self.ARG_COLUMN_SEQUENCE] = column_sequence
+
+        try:
+            resp_status_code = ""
+            log.logger.info(F"Training model by RayServe, target: {target}, "
```
paxcema, line 58: Change start of the string to "Started training a RayServe model..."

mindsdb/integrations/handlers/rayserve_handler/rayserve_handler.py
```diff
@@ -0,0 +1,101 @@
...
+                            F"train_url: {train_url}, predict_url: {predict_url}")
+            encoded_df = base64.encodebytes(pickle.dumps(df))
+            resp = requests.post(train_url, json={'df': encoded_df, 'target': target,
+                                                  "predictor_id": predictor_id, "predictor_name": predictor_name})
+            resp_status_code = resp.status_code
+            assert resp.status_code == 200
+            log.logger.info(F"Training model by RayServe success.")
```
paxcema, line 65: I'm a bit unsure this is correct. The model behind the `train_url` could potentially train very slowly, so I don't think the process is ever synchronous.

In fact, I think this should be changed to follow the Lightwood handler, where we detach the learning into a separate process that will commit training stats to the DB (e.g. accuracy, training time, etc.) once the learning process _does_ finish. Would you agree?

mindsdb/integrations/handlers/rayserve_handler/rayserve_handler.py
```diff
@@ -0,0 +1,101 @@
...
+        feature_column_sequence = list(deserialize_args[self.ARG_COLUMN_SEQUENCE])
+        if target in feature_column_sequence:
+            feature_column_sequence.remove(target)
+        predict_url = deserialize_args[self.ARG_PREDICT_URL]
+        predictor_id = deserialize_args[self.ARG_PREDICTOR_ID]
+        predictor_name = deserialize_args[self.ARG_PREDICTOR_NAME]
+
+        df_send = df[feature_column_sequence]
+        encoded_df = base64.encodebytes(pickle.dumps(df_send))
+        resp = requests.post(predict_url, json={'df': encoded_df, 'target': target,
```
paxcema, line 86: Can't quite remember, but both here and in `create()`: are `predictor_id` and `predictor_name` actual keys required by Ray Serve?

mindsdb/integrations/handlers/rayserve_handler/rayserve_handler.py
```diff
@@ -0,0 +1,101 @@
...
+        pred_result = json.loads(resp.content)
+        if target not in pred_result:
+            raise Exception(F"Target column doesn't exist: {target}")
+        if len(pred_result[target]) != len(df_send):
+            raise Exception(F"Line number of pre diction doesn't equal with the one of input.")
+        rt = df.copy(deep=True)
+        pred_df = pd.DataFrame(index=rt.index, data=pred_result)
+        if target in set(rt.columns):
+            rt.drop(columns=[target], inplace=True)
+        rt = rt.join(pred_df, how="left")
```
paxcema, line 97: Joining back to the original input df is not required, so let's remove this for simplicity's sake :)

mindsdb/integrations/handlers/rayserve_handler/rayserve_handler.py
```diff
@@ -0,0 +1,101 @@
...
+        return rt
+
+if __name__=="__main__":
```
paxcema, line 100: Remove this one, please.

paxcema (changes requested):

mindsdb/integrations/handlers/rayserve_handler/rayserve_handler.py
rfsiten: yes, I've just ref to the code of rayServe in Feb. 2022.
Here. I've got a problem, if we want to record the status, then we must require another url to query the status of RayServe. Would you mind if an "url.status" parameter to be added to using block?

mindsdb/integrations/handlers/rayserve_handler/rayserve_handler.py
rfsiten: > Can't quite remember, but both here and in `create()`: are `predictor_id` and `predictor_name` actual keys required by Ray Serve?

It's really a topic should be discussed intensively, do you think the url.train already contains enough information of the model?  If you want to build two models in rayServe, do you think we should use the different url or the same url with different arguments? Example:
Method 1:
model1:
url.train = http://.../model1
model2:
url.train = http://.../model2

or

Method 2:
model1:
url.train = http://.../models  with post info {"predictor_name": "model1"}
model2:
url.train = http://.../models  with post info {"predictor_name": "model2"}

Which is the better way or the way you designed to implement?

mindsdb/integrations/handlers/rayserve_handler/rayserve_handler.py
paxcema: Not at all! Feel free to add that parameter.

rfsiten: ok, I suppose I will commit my modification next week.

ZoranPandovski:
@rfsiten Please check the latest changes when you can so we can merge this

paxcema:
@rfsiten any chance you can/want to work on the above points? If not, we would be happy to address this in order to merge the PR.

paxcema:
Hi @rfsiten, we'll be closing this as it was resolved in #5556. I'd encourage you to take a look, try it out and -if you run into any areas for improvement- we can surely take inspiration from the nice work you've outlined here and collaborate again in the future.

PR #3214, opened 2022-10-08
Fix timeseries alignment

### LINKED ISSUE #3213: [Bug]:  wrong data join data in time-series
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

when add data field to JOIN query in house sale tutorial, the output is not correct.
For example, with window=4/horizon=2, the result was
1) no where condition, (*) is wrong result
```
mysql> SELECT t.saledate,m.saledate AS date, m.MA AS forecast FROM mindsdb.home_sales_pred AS m  JOIN trino_datasource.home_sales AS t WHERE t.type = 'house'  AND t.bedrooms = 2;
+------------+------------+--------------------+
| saledate   | date       | forecast           |
+------------+------------+--------------------+
| 2007-09-30 | 2007-09-30 | 441413.5849598734  |
...
| 2019-06-30 | 2019-06-30 | 498045.07026576105 |
| NULL(*)       | 2019-09-30 | 511941.7721996386  |
| NULL       | 2019-12-31 | 517506.31349071994 |
+------------+------------+--------------------+
50 rows in set (1.08 sec)
```

2) =LATEST, `saledate` is fine
```
mysql> SELECT t.saledate,m.saledate AS date, m.MA AS forecast FROM mindsdb.home_sales_pred AS m  JOIN trino_datasource.home_sales AS t WHERE t.type = 'house'  AND t.bedrooms = 2 and t.saledate = LATEST;
+------------+---------------------+-------------------+
| saledate   | date                | forecast          |
+------------+---------------------+-------------------+
| 2019-09-30 | 2019-09-30 00:00:00 | 441413.5849598734 |
+------------+---------------------+-------------------+
1 row in set (0.99 sec)

3) "> LATEST" , (*) is wrong result
mysql> SELECT t.saledate,m.saledate AS date, m.MA AS forecast FROM mindsdb.home_sales_pred AS m  JOIN trino_datasource.home_sales AS t WHERE t.type = 'house'  AND t.bedrooms = 2 and t.saledate > LATEST;
+------------+---------------------+--------------------+
| saledate   | date                | forecast           |
+------------+---------------------+--------------------+
| 2019-09-30 (*) | 2019-12-31 00:00:00 | 511941.7721996386  |
| NULL       | 2020-04-01 00:00:00 | 517506.31349071994 |
+------------+---------------------+--------------------+
2 rows in set (1.37 sec)
````

### Expected Behavior

expect result
1) without condition
```
mysql> SELECT t.saledate,m.saledate AS date, m.MA AS forecast FROM mindsdb.home_sales_pred AS m  JOIN trino_datasource.home_sales AS t WHERE t.type = 'house'  AND t.bedrooms = 2;
+------------+------------+--------------------+
| saledate   | date       | forecast           |
+------------+------------+--------------------+
| 2007-09-30(*) | 2007-09-30 | 441413.5849598734  |
...
| NULL       | 2019-12-31 | 517506.31349071994 |
+------------+------------+--------------------+
50 rows in set (1.10 sec)
```

2) = LATEST, no change
```
mysql> SELECT t.saledate,m.saledate AS date, m.MA AS forecast FROM mindsdb.home_sales_pred AS m  JOIN trino_datasource.home_sales AS t WHERE t.type = 'house'  AND t.bedrooms = 2 and t.saledate = LATEST;
+------------+---------------------+-------------------+
| saledate   | date                | forecast          |
+------------+---------------------+-------------------+
| 2019-09-30 | 2019-09-30 00:00:00 | 441413.5849598734 |
+------------+---------------------+-------------------+
1 row in set (0.99 sec)
```

3) "> LATEST"
```
mysql> SELECT t.saledate,m.saledate AS date, m.MA AS forecast FROM mindsdb.home_sales_pred AS m  JOIN trino_datasource.home_sales AS t WHERE t.type = 'house'  AND t.bedrooms = 2 and t.saledate > LATEST;
+----------+---------------------+--------------------+
| saledate | date                | forecast           |
+----------+---------------------+--------------------+
| NULL(*)     | 2019-12-31 00:00:00 | 511941.7721996386  |
| NULL     | 2020-04-01 00:00:00 | 517506.31349071994 |
+----------+---------------------+--------------------+
2 rows in set (1.00 sec)
```

### Steps To Reproduce

```markdown
1. Run the house sale tutorial,
2. Make the prediction
3. Query the result, with JOIN and add `t.saledate` to the select column
4. Check the result
```

### Anything else?

this issue only mention about the additional column. The value of the prediction for `2019-09-30` are also different, but we will mention in other issue.

bachng2017:
Fixes #3213

## change the logic of __no_forecast_offset

with current implementation, the predict row `__mindsdb_row_id` is not cleared correctly for the 1st row in the 2 cases
- no WHERE condition
- "> LATEST" is used

https://github.com/mindsdb/mindsdb/blob/staging/mindsdb/integrations/handlers/lightwood_handler/lightwood_handler/lightwood_handler.py#L622

So the wrong data is joined.

### DIFF
--- a/mindsdb/integrations/handlers/lightwood_handler/lightwood_handler/lightwood_handler.py
+++ b/mindsdb/integrations/handlers/lightwood_handler/lightwood_handler/lightwood_handler.py
@@ -554,7 +554,7 @@ class LightwoodHandler(PredictiveHandler):
         timeseries_settings = predictor_record.learn_args['timeseries_settings']
 
         if timeseries_settings['is_timeseries'] is True:
-            __no_forecast_offset = set([row.get('__mdb_forecast_offset', None) for row in pred_dicts]) == {None}
+            __no_forecast_offset = set([row.get('__mdb_forecast_offset', None) for row in pred_dicts]) == {1}
 
             predict = predictor_record.to_predict[0]
             group_by = timeseries_settings['group_by'] or []


bachng2017:
just wonder why the tests always fail

paxcema:
Tests can fail sometimes when servers involved in the CI time out. I've put in a request to repeat them, let's see if they pass.

The change itself _looks_ good, but I will take a closer look before approving.

paxcema (commented):
While this may solve the error, it is out of coincidence 😅. It is not correct to say that there is no forecast offset when it is actually being set to +1. I will close this PR to fix the underlying issue in another one.

bachng2017:
@paxcema thanks for the review

is it really a coincidence ?

according to these codes, 
https://github.com/mindsdb/mindsdb/blob/staging/mindsdb/api/mysql/mysql_proxy/classes/sql_query.py#L1142

_mdb_forecast_offset = 0  when "=LATEST"
_mdb_forecast_offset = 1 when ">LATEST"
_mdb_forecast_offset = None else

which make we think that only there is only real forecast when `_mdb_forecast_offset = 1 `

paxcema:
> there is only real forecast when `_mdb_forecast_offset = 1`

Yeah, this is correct! 

What I meant was that modifying the variable `__no_forecast_offset` is not the way to go because it would not be a valid name for it anymore. The solution is a bit more involved, which is what I will try to implement in a separate PR. But your interpretaion of the code above is correct 👍 .

For a bit more information on `__mdb_forecast_offset`, there is also [this docstring](https://github.com/mindsdb/lightwood/blob/40cd2b1542febc6a08a3f7d05a24b9375994e065/lightwood/data/timeseries_transform.py#L83).

bachng2017:
ic, although mindsdb does not support this syntax yet `WHERE T = LATEST - K` but the issue need a more general solution. As far as the issue is aware, we are good :)
Thanks for checking and reply

PR #3311, opened 2022-10-11
ML Engine docstrings

### LINKED ISSUE #3310: [Docs] Thoroughly document all new ML engine source code
First, as docstrings. In a later stage we can replicate those inside the documentation.

Relevant files for this are:
- `mindsdb/interfaces/storage/fs.py`
- `mindsdb/integrations/libs/ml_exec_base.py`
- `mindsdb/integrations/libs/base.py`
- `mindsdb/integrations/handlers/byom_handler/proc_wrapper.py`

paxcema:
Partially fixes #3310


mindsdb/integrations/handlers/byom_handler/proc_wrapper.py
```diff
@@ -1,6 +1,16 @@
...
-import pickle
+"""
+Utility functions used in the 'Bring Your Own Model' (BYOM) engine.
+
+These functions interact with:
+    - interfaces (stdin, stdout),
+    - python files,
+    - and the actual BYOM engine.
+"""
+
```
ea-rus, line 9: Doc to add/replace:
Wraps and run python code in separate python proceess. 
Communicates with parent process throughout stdin/out using pickle to serialize objects.
Flow:
 - Receives module code, method with params and stored attributed from parrent process
 - Creates python class object from code
 - Creates instance from class and filled with stored attributes
 - Calls choosen method of class with params
 - Packes response and sends to stdout
 - Exit

PR #3340, opened 2022-10-12
Use fully ready postgres db container for tests

ilia-tsyplenkov:
Use fully ready postgres db container for tests. Test covers ssl/nossl connection


mindsdb/integrations/handlers/postgres_handler/tests/docker/certs/server.key
```diff
@@ -0,0 +1,28 @@
+-----BEGIN PRIVATE KEY-----
```
paxcema, line 1: Is this OK? 😮

ilia-tsyplenkov: yes, it is a self signed certificate which I upload to the docker-container to have ssl connection with test postgres database container

paxcema: Aaah got it 👌

PR #3377, opened 2022-10-14
ML handler supbrocess

ea-rus:
Changes: 

1 Dividing into components:

- app initialize
    parse args
    set env.MINDSDB_CONFIG_PATH

- config
    requiers env.MINDSDB_CONFIG_PATH
    sets
        env.MINDSDB_DB_CON
    Config() - makes initialization

- log
    uses config
    initialize_log - makes initialization

- database
    uses env.MINDSDB_DB_CON
    have init() method

- file storage
    uses config

2 Changes in access to database and logging

3 created two modes to run predict method: 
- in subprocess
- in running subprocess

Mode is the setting in ml handler

4 added unit test for Lightwood handler

5 fixes in HF handler


ilia-tsyplenkov:
@ea-rus 89 files in one PR. Are you serious thinking that it is reviewable?

ea-rus:
> @ea-rus 89 files in one PR. Are you serious thinking that it is reviewable?

Most of them are repeatable changes of usage database and logger. It is need for component decoupling and make it able to import independently

ilia-tsyplenkov:
> > @ea-rus 89 files in one PR. Are you serious thinking that it is reviewable?
> 
> Most of them are repeatable changes of usage database and logger. It is need for component decoupling and make it able to import independently

independent database usage might be in one PR
independent logger usage in the second PR
config in the third
ML related changes in another


That is how sensible decomposition might look like.
But instead of doing it you decided to move all eggs into one bucket. Again!

ilia-tsyplenkov:
95 files have been changed in this PR while the whole mindsdb repository contains 529 files of python code
```
itsyplen@it-laptop:~/repos/work/MindsDB/mindsdb$ find . -type f -name "*.py" | wc -l
529
```
Thus this PR affects 18% of the whole repository!!!!! it would be funny if it wasn't so sad

paxcema:
I agree @ilia-tsyplenkov, we need to improve separation of concerns within PRs. Massive changes in line count is fine if the underlying reason is unique and simple to grasp. We should avoid massive changes where, conceptually, there is a mix of multiple things going on as in here.

mindsdb/utilities/log.py
```diff
@@ -47,100 +35,80 @@ def isatty(self):
...
+#         log = Log(log_type=str(log_type), source=source, payload=str(payload), company_id=self.company_id)
+#         session.add(log)
+#         session.commit()
+
+# default logger
+logger = logging.getLogger('dummy')
+
+
+def initialize_log(config=None, logger_name='main', wrap_print=False):
+    global logger
```
paxcema, line 98: Maybe a name with less chance of collision?

mindsdb/utilities/config.py
```diff
@@ -22,15 +23,45 @@ def _merge_configs(original_config, override_config):
     return original_config
 
 
+config = None
+
 class Config():
     def __init__(self):
+        # initialize once
+        global config
```
paxcema, line 31: Maybe a name with less chance of collision?

mindsdb/integrations/libs/ml_handler_proc.py
```diff
@@ -0,0 +1,227 @@
...
+    def init_handler(self, class_path, company_id, integration_id, predictor_id):
+        # mdb initialization
+        import mindsdb.interfaces.storage.db as db
+        db.init()
+
+        from mindsdb.interfaces.storage.model_fs import ModelStorage, HandlerStorage
+
+        module_name, class_name = class_path
+        module = importlib.import_module(module_name)
+        klass = getattr(module, class_name)
```
paxcema, line 97: Just a nitpick: I thought the standard alternate naming for "class" in python was "cls", though I can't find a source right now.

paxcema (approved):
Generally looks okay, but I focused on the ML subprocess -related changes. I would recommend that @StpMax reviews the rest of the changes (renaming log->logger notwithstanding) before merging.

mindsdb/__init__.py
```diff
@@ -1,95 +1,2 @@
...
-
-    if telemetry_file_exists(mindsdb_config['storage_dir']):
-        os.environ['CHECK_FOR_UPDATES'] = '0'
-        print('\n x telemetry disabled! \n')
-    elif os.getenv('CHECK_FOR_UPDATES', '1').lower() in ['0', 'false', 'False'] or mindsdb_config.get('cloud', False):
-        disable_telemetry(mindsdb_config['storage_dir'])
-        print('\n x telemetry disabled \n')
-    else:
-        print('\n ✓ telemetry enabled \n')
+from mindsdb.__about__ import __version__  # noqa: F401
```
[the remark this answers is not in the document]
ea-rus: I think better to have version in root package like other modules do
```
import json
json.__version__
```
Also some of mindsdb code is using this version from root

mindsdb/__main__.py
```diff
@@ -19,14 +20,26 @@
...
 from mindsdb.interfaces.model.model_controller import ModelController
 from mindsdb.interfaces.database.integrations import IntegrationController
 import mindsdb.interfaces.storage.db as db
 from mindsdb.integrations.utilities.install import install_dependencies
 
+from mindsdb.utilities.fs import create_dirs_recursive
+from mindsdb.utilities.functions import args_parse
+from mindsdb.utilities.telemetry import telemetry_file_exists, disable_telemetry
+
+# is_ray_worker = False
```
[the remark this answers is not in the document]
ea-rus: When there is probability that some block of code is removed by mistake and something wrong without it, I prefer to keep it commented. And will delete it in time of next task for this file.
It is convinient to see what code was there before in case of debuging

ea-rus:
> 95 files have been changed
Yes but I was not sure that these changes will help to split ml handler to different process so I didn't push them and keep as a big experiment)

mindsdb/integrations/libs/ml_handler_proc.py
ea-rus: Max renamed it in ml_exec_base.py
I will rename it the same

PR #3539, opened 2022-10-20
Cleaned up handler tests

Temidayo32:
Fixes #

## Please describe what changes you made, in as much detail as possible:
This constitute a clean up of the unit tests written for the database handlers. This involves fixing wrong import paths, and missing dictionary keys (most of the test didn't include the `connection_data` key even though it was written in the handler!).


Temidayo32:
@ZoranPandovski please review this

Temidayo32:
Next up will be to fix the undefined functions. Would suggest removing them? Or defining those functions in the handlers?

Temidayo32:
By undefined methods, I mean methods called in the test modules by are not defined in their corresponding handler modules

Temidayo32:
@ZoranPandovski I am particular about the tests because I assume the contributors writing integrations would also be responsible for the tests too. So the tutorial would then cover not just the integration itself, but I also on writing the tests. Or what do you say?

PR #3552, opened 2022-10-21
Bug Fix: removed undefined tests

MinuraPunchihewa:
I have removed some tests that use undefined functions by going through all of the handlers.

I have just cleaned up the code a little bit as well in some of the test modules.

Further, I have noticed that no tests have been defined for the following handlers,

- byom
- file
- mongodb
- view
- huggingface

Maybe these can be taken up as separate issues?

@ZoranPandovski, a question for you, is it possible to use the `.query()` method within the test modules and then run it using Python locally? I was under the impression that the `native_query()` method needs to be used, unless we are running the code through the MindsDB SQL Editor.


ZoranPandovski (approved):

ZoranPandovski:
@MinuraPunchihewa it depends, the idea with query was that it gets AST syntax and uses dialect to convert it to SQL. In the case of `native_query` where it accepts the raw SQL query.

PR #3602, opened 2022-10-22
Update db_handler.Dockerfile

PramodBobade:
Fixes #

## Added higher python version
  -

mindsdb


ZoranPandovski (changes requested):
MindsDB is not compatible with python 3.10 because some of our dependencies are not supporting it

PR #3653, opened 2022-10-24
Сheck target of predictor 

ea-rus:
Changes: 
- check if training dataframe contents column to predict and show error to user before run learn in subproccess.


paxcema (approved):

ea-rus:
#3268

PR #3765, opened 2022-10-28
[fix] json_ai override and problem definition generation

paxcema:


mindsdb/integrations/libs/ml_exec_base.py
```diff
@@ -311,16 +311,27 @@ def learn(self, statement):
...
-        if problem_definition is None:
-            problem_definition = {}
+        problem_definition = {'target': target}
 
-        problem_definition['target'] = target
+        for attr in ['using', 'horizon', 'window']:
+            if hasattr(statement, attr) and getattr(statement, attr) is not None:
+                problem_definition[attr] = getattr(statement, attr)
+
+        if hasattr(statement, 'order_by') and statement.order_by is not None:
```
ea-rus, line 320: You can remove 'hasattr(statement, 'order_by')' and the similar because statement is CreatePredictor and already has these attributes

tests/integration_tests/flows/test_mysql_api.py
```diff
@@ -247,16 +247,16 @@ def test_select_from_files(self):
 
     def test_ts_train_and_predict(self, subtests):
         train_df = pd.DataFrame({
-            'group': ["A" for _ in range(100, 210)] + ["B" for _ in range(100, 210)],
```
ea-rus, line 250: had this test failed without these renamings ?

mindsdb/integrations/libs/ml_exec_base.py
```diff
@@ -311,16 +311,27 @@ def learn(self, statement):
         if target not in training_data_df.columns:
             raise Exception(f'Prediction target "{target}" not found in training dataframe: {list(training_data_df.columns)}')
 
-        problem_definition = statement.using
-        if problem_definition is None:
-            problem_definition = {}
+        problem_definition = {'target': target}
```
paxcema, line 314: remove argument, already passed in `create()` signature

PR #4028, opened 2022-11-11
remove view handler

StpMax:


mindsdb/integrations/libs/ml_exec_base.py
```diff
@@ -282,6 +283,21 @@ def learn(
...
+                data_integration_ref = {
+                    'type': 'view'
+                }
+            else:
+                data_integration_ref = {
+                    'type': 'integration',
+                    'id': data_integration_meta['id']
+                }
+
+        problem_definition = {'target': target}
```
ea-rus, line 297: this clears  problem_definition (except of target)

StpMax: fixed

PR #4067, opened 2022-11-17
[fix] hh:mm:ss support in get_date_format

### LINKED ISSUE #4047: [Bug][CUSTOMER]: Timeseries ORDER BY column removes HH:MM:SS in output
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior
image

https://www.loom.com/share/a8899ba512214d178e3f0ad3c043275b

### Expected Behavior

The model's ORDER BY column should be the same as the input.
Alternatively, it could be a tidied version e.g. for an hour interval: 12:00:01 tidied to 12:00:00

### Steps To Reproduce

```markdown
Queries below:
https://docs.google.com/document/d/1OCxOeTN-TFED65m3BWf2ycQxkPH44SmFUWftlnxGhzo/edit?usp=sharing
```

### Anything else?

_No response_

paxcema:
Fixes #4047


paxcema:
Note: this doesn't fully solve the problem, as the GUI still shows the timestamp object returned by the model in YYYY:MM:DD format. From what I can tell, `mindsdb` is doing the right thing as it returns a string for the reference datetime from the table (which is correctly showed in the GUI) and it passes a `Timestamp` object for the one inferred by the model. I assume that this may be some casting or formatting issue in the GUI, but I'd like @setohe0909 or @alejandrovillegas to confirm before we merge this.

paxcema:
@setohe0909 has confirmed that the GUI does not alter the data in any way. I'm tempted to change our model's output type to string as well, because I remembered this behavior brought a few problems when interacting with BI tools like Metabase or Superset.

ea-rus:
Added test for PR. Is I understanding problem right?

StpMax:
@paxcema if change model output to string, then we will need to parse it at mindsdb side. We use this field to filter result, to do it we need timestamps. @ea-rus right?

paxcema:
@ea-rus indeed, that was part of the problem, thanks for the test!

The only remaining issue is as follows: given the types in the screenshot above, the GUI is still reporting the timestamp _without_ HH:MM:SS information, even though it exists when the data is output from MindsDB.

@StpMax What I propose as a potential fix is to cast timestamps into a string after we filter. However, there may be some other better fix 🤔.

StpMax:
@paxcema @ea-rus time is missing in here https://github.com/mindsdb/mindsdb/blob/staging/mindsdb/api/mysql/mysql_proxy/classes/sql_query.py#L1016

paxcema:
@StpMax do you mean it getting filtered out? Or it is already arriving without time? That's strange, I'm pretty sure I checked yesterday and fixed the internals where it was being stripped.

StpMax:
@paxcema it is going with time to `apply_ts_filter`, but result is without time

paxcema:
Just to confirm @StpMax, did you observe that behavior using commit `669267d` of this branch? I can replicate what you describe in `staging`, however, it is solved (at least locally for me) with the fixes in this PR. 

FYI I have implemented my proposed fix in the last commit.

PR #4183, opened 2022-12-06
[ENH] Support for adjusting models

paxcema:
This PR adds support for adjusting / fine-tuning a model with new data. The fundamental difference when compared to the retrain command is that we're not starting from scratch. Instead, we copy the latest `COMPLETE` model version and use it as a starting point for the ML handler to `update` it, provided the backend supports this option, of course.

Note, it requires a new `mindsdb_sql` release as it uses the syntax for `ADJUST`.


mindsdb/__main__.py
```diff
@@ -9,7 +9,7 @@
 import json
 
 import torch.multiprocessing as mp
-mp.set_start_method('spawn')
+# mp.set_start_method('spawn')
```
paxcema, line 12: uncomment

mindsdb/integrations/handlers/lightwood_handler/lightwood_handler/functions.py
```diff
@@ -150,41 +149,32 @@ def run_learn(df: DataFrame, args: dict, model_storage) -> None:
...
-def run_update(predictor_id: int, df: DataFrame, company_id: int):
+@mark_process(name='adjust')
+def run_adjust(df: DataFrame, args: dict, model_storage):
     try:
-        predictor_record = db.Predictor.query.filter_by(id=predictor_id).first()
+        base_predictor_id = args['base_model_id']
+        base_predictor_record = db.Predictor.query.filter_by(
+            id=base_predictor_id,
+            status=PREDICTOR_STATUS.COMPLETE
+        ).first()  # TODO should this be last instead?
```
paxcema, line 159: should this be last instead?

mindsdb/integrations/libs/ml_exec_base.py
```diff
@@ -147,6 +148,100 @@ def learn_process(class_path, context_dump, integration_id,
     db.session.commit()
 
 
+@mark_process(name='adjust')
+def adjust_process(class_path, context_dump, integration_id,
```
paxcema, line 152: Maybe worth refactoring the common bits between this process and learn into helper(s) function(s)?

mindsdb/integrations/handlers/lightwood_handler/lightwood_handler/functions.py
ea-rus: no, we are filtering by id, there should be only one record

tests/unit/test_executor.py
```diff
@@ -5,7 +5,7 @@
```
ea-rus, line 5: it whould be great to have a test for adjust

PR #4384, opened 2023-02-01
Jobs manage

ea-rus:
## Description

[DOCS](https://github.com/mindsdb/mindsdb/blob/jobs-manage/mindsdb/interfaces/jobs/README.md)

https://www.loom.com/share/fda7d7ada2a14a4da83f380b956805cb

## Type of change

### What is the solution?

## Checklist:


mindsdb/interfaces/jobs/README.md
```diff
@@ -0,0 +1,146 @@
...
+)
+START { now | <date> }
+END { now | <date> }
+EVERY [<numer>] <period> 
+```
+Parameters:
+- START - optional, first date when job have to start. Default is now
+- END - optional, after this date job shouldn't run anymore. If not set - repeat forever
+- EVERY - optional job repeat rule.
+  - If not set - run once (no mather what was set in END)
```
paxcema, line 17: typo - ~mather~ -> matter

mindsdb/interfaces/jobs/README.md
```diff
@@ -0,0 +1,146 @@
...
+```
+
+This example creates job in 'proj1' project and runs it every 2 days till 1st April. 
+
+This job have to
+- retrain predictor 
+- use trained predictor on records from table 'int2.table1'
+- append results of prediction to int1.rentals table
+
+join_learn_process=true - this parameter is used for waiting to retrain predictor
```
paxcema, line 55: Thinking at a higher level, maybe it's worth introducing a blocking mechanism of some kind for this? Because other jobs like adjust or create would also be async and hence we need to wait. Or maybe just using `join_learn_process` on all three of those is good enough for now.

mindsdb/interfaces/jobs/jobs_controller.py
```diff
@@ -0,0 +1,351 @@
...
+import sqlalchemy as sa
+
+from mindsdb_sql import parse_sql, ParsingException
+
+from mindsdb.interfaces.storage import db
+from mindsdb.interfaces.database.projects import ProjectController
+from mindsdb.utilities.context import context as ctx
+from mindsdb.utilities import log
+
+mp_ctx = mp.get_context('spawn')
```
paxcema, line 15: Shouldn't this be inside a try-except? I think the main entry point to mindsdb is already setting this.

mindsdb/interfaces/jobs/scheduler.py
```diff
@@ -0,0 +1,98 @@
...
+import torch.multiprocessing as mp
+
+from mindsdb.utilities.config import Config
+from mindsdb.utilities.log import initialize_log
+from mindsdb.utilities import log
+from mindsdb.interfaces.storage import db
+
+from mindsdb.interfaces.jobs.jobs_controller import JobsExecutor
+
+mp_ctx = mp.get_context('spawn')
```
paxcema, line 13: Same as the comment above.

mindsdb/interfaces/jobs/scheduler.py
```diff
@@ -0,0 +1,98 @@
...
+        logger.debug('Scheduler check timetable')
+        try:
+            check_timetable(config)
+        except (SystemExit, KeyboardInterrupt):
+            raise
+        except Exception as e:
+            logger.error(e)
+
+        # different instances should start in not the same time
+        time.sleep(20 + random.randint(1, 10))
```
paxcema, line 30: Shouldn't this also be configurable? I can envision local deployments where the user wants monitoring at the per-second level (or maybe fractions of a second!).

mindsdb/interfaces/jobs/scheduler.py
```diff
@@ -0,0 +1,98 @@
...
+def execute_sync(record, config):
+    logger.info(f'Job execute: {record.name}({record.id})')
+
+    exec_method = config.get('jobs', {}).get('executor', 'local')
+
+    task_process(record.id, exec_method)
+
+
+def execute_async(record, config):
+    # if async is used with training in background then training is not happening
```
paxcema, line 49: Should we check this?

For example, when parsing the query if would be invalid to ask for a train/retrain/adjust job if there is no `USING` clause with `join_learn_process` set to True.

paxcema (changes requested):
Overall looks good to me! Left a couple of suggestions but maybe those are best left for another PR.

mindsdb/interfaces/jobs/README.md
ea-rus: I think join_learn_process is enough for now. Because it should work for retrain and adjust too

mindsdb/interfaces/jobs/jobs_controller.py
ea-rus: I looked at the other examples of mp.get_context, they are not try-catched

mindsdb/interfaces/jobs/scheduler.py
ea-rus: I will add to config. But for now we can't run job more often than every minute. So precision is not about seconds now.

mindsdb/interfaces/jobs/README.md
paxcema: Agreed 👍

mindsdb/interfaces/jobs/jobs_controller.py
paxcema: My bad, I was thinking of `set_start_method`, ignore this)

mindsdb/interfaces/jobs/scheduler.py
paxcema: Makes sense

paxcema (approved):

PR #4395, opened 2023-02-03
Apache Phoenix/HBase Integration

MinuraPunchihewa:
I have implemented the new integration for Databend. After some initial testing, it looks to be working as expected. Please do let me know if there is anything that I have missed or if there are any bugs.

@ZoranPandovski I believe this integration completes both the Apache Phoenix and Apache HBase tasks. This is because Apache Phoenix is essentially used to query HBase using SQL. It is described as a "SQL skin over HBase" and is not exactly a data source on its own. This is what I have discovered in my research. What do you think?


ZoranPandovski (approved):
Looks great

ZoranPandovski:
@MinuraPunchihewa So, can HBASE be implemented by extending this one? If so, that's great. We can implement that too. One confusing thing for me is how the authentication is handled. Does the `readonly` flag always authorize everyone to `SELECT`?

MinuraPunchihewa:
@ZoranPandovski Regarding authentication, apologies that is something that did not occur to me. I think I can fix it. Let me make a few changes.

MinuraPunchihewa:
@ZoranPandovski About HBase, basically Apache Phoenix is the established way to query HBase using SQL (since it is a NoSQL database). Phoenix is not exactly a storage system on its own, but rather SQL skin that sits on top of HBase.

The following architecture diagram might make it a little more clear,
image

However, this does not mean you can query HBase directly with this integration. You'd still need what is known as a Phoenix Query Server.

From what I have gathered, there is no way to query HBase with SQL apart from this. Do you want me to investigate this further?

ZoranPandovski:
Great, let's discuss this for future improvements/changes.

PR #4447, opened 2023-02-10
Amazon Aurora Integration

MinuraPunchihewa:
I have implemented the new integration for Amazon Aurora. I have not been able to perform any testing as I don't really have access to an AWS account.

Amazon Aurora comes in two flavors: a MySQL compatible version and a PostgreSQL compatible version. This integration makes use of the relevant handlers that have already been implemented for these RDBMSs.

It uses `boto3` to identify which database engine that the database cluster is running and then instantiates the relevant handler. For this purpose, users will be required to pass in their AWS access key and secret access key. However, this can be simplified by getting the users to pass in the name of the database engine as a parameter. 
@ZoranPandovski What are your thoughts on this?

I am somewhat confident that this integration will work, however, it would be beneficial to conduct some testing.

### DIFF
--- a/mindsdb/integrations/handlers/aurora_handler/__about__.py
+++ b/mindsdb/integrations/handlers/aurora_handler/__about__.py
@@ -1,0 +1,9 @@
+__title__ = 'MindsDB Amazon Aurora handler'
+__package_name__ = 'mindsdb_aurora_handler'
+__version__ = '0.0.1'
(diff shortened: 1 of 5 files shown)


ZoranPandovski (approved):

ZoranPandovski:
I will share the instance info and we can discuss the improvements :raised_hands:

PR #4554, opened 2023-02-20
Show training progress

ilia-tsyplenkov:
## Description
Do required changes in db, mode_storage, queries handling to show training progress

### What is the solution?

## Checklist:


mindsdb/interfaces/storage/model_fs.py
```diff
@@ -33,6 +33,18 @@ def status_set(self, status, status_info=None):
             rec.data = status_info
         db.session.commit()
 
+    def training_state_set(self, current_state_num, total_states, state_name):
```
ea-rus, line 36: Maybe is better not having all parameters required, for example I want to do:
`training_state_set(1, 10)`
then
`training_state_set(2)`
or want to use only state names `training_state_set(name="almost done")`
@paxcema, what do you think?

ea-rus (changes requested):

mindsdb/integrations/handlers/lightwood_handler/lightwood_handler/lightwood_handler.py
```diff
@@ -409,6 +409,10 @@ def _get_ensemble_data(self):
 
         return pd.DataFrame([[ai_info_str]], columns=['ensemble'])
 
+    def _get_progress_data(self):
```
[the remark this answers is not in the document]
paxcema: I agree. Having it in describe could also be a nice to have, but showing it in `select * from models` makes the most sense to me.

mindsdb/interfaces/storage/model_fs.py
paxcema: I guess this way it could be more flexible. For example maybe a handler would like to give "sub-updates" within a phase, so with this they could simply do `training_state_set(state_name='now doing X')`.

Seems like it's not a huge change (we can partially update the record with the subset of provided values), so I would be in favor just to see how implementers use it.

ilia-tsyplenkov: Done

mindsdb/integrations/handlers/lightwood_handler/lightwood_handler/lightwood_handler.py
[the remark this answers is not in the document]
ilia-tsyplenkov: Done

PR #4637, opened 2023-02-23
Dremio Integration

MinuraPunchihewa:
I have implemented the new integration for Dremio. After some initial testing, it looks to be working as expected. Please do let me know if there is anything that I have missed or if there are any bugs.


mindsdb/integrations/handlers/dremio_handler/dremio_handler.py
```diff
@@ -0,0 +1,236 @@
...
+            HandlerStatusResponse
+        """
+
+        headers = {
+            'Content-Type': 'application/json',
+        }
+
+        data = '{' + f'"userName": "{self.connection_data["username"]}","password": "{self.connection_data["password"]}"' + '}'
+
+        response = requests.post(self.base_url + '/apiv2/login', headers=headers, data=data, verify=False)
```
ZoranPandovski, line 69: For other calls we used apiv3, is it required to use v2 for login?

mindsdb/integrations/handlers/dremio_handler/requirements.txt
```diff
@@ -0,0 +1,2 @@
+pandas
```
ZoranPandovski, line 1: This is already depenedency of mindsdb, so we can exclude it from here

mindsdb/integrations/handlers/dremio_handler/dremio_handler.py
```diff
@@ -0,0 +1,236 @@
...
+                log.logger.info('Job creation successful. Job id is: ' + job_id)
+            else:
+                log.logger.info('Job creation failed.')
+
+            log.logger.info('Waiting for the job to complete...')
+
+            job_status = requests.request("GET", self.base_url + "/api/v3/job/" + job_id, headers=auth_headers).json()[
+                'jobState']
+
+            while job_status != 'COMPLETED':
```
ZoranPandovski, line 137: Why we need this?

ZoranPandovski (changes requested):
Thanks @MinuraPunchihewa. Please check the above comments

mindsdb/integrations/handlers/dremio_handler/requirements.txt
MinuraPunchihewa: Got it.

mindsdb/integrations/handlers/dremio_handler/dremio_handler.py
MinuraPunchihewa: We need this because when we submit a job to the Dremio server, we need to wait until it is completed for us to extract the results. If we try to get the results while the job is still running, it will return an error.

mindsdb/integrations/handlers/dremio_handler/dremio_handler.py
MinuraPunchihewa: Yes, it looks like the authentication still uses an older version of the API,
https://docs.dremio.com/software/rest-api/#authentication

ZoranPandovski (approved):

PR #4712, opened 2023-02-28
HF: token limit message

ea-rus:
## Description

Show error message in value_error column if input size exceed model limit

**Fixes** #4491 

Added truncation_policy param, values: left, right, strict(default)

## Type of change

### What is the solution?

## Checklist:


mindsdb/integrations/handlers/huggingface_handler/huggingface_handler.py
```diff
@@ -188,7 +188,17 @@ def tidy_output_summarization(args, result):
...
-        input_list_str = [str(x) for x in input_list]
+
+        max_tokens = pipeline.tokenizer.model_max_length
+        input_list_str = []
+        errors = []
+        for i, line in enumerate(input_list):
+            if max_tokens is not None:
+                tokens = len(pipeline.tokenizer(line)['input_ids'])
+                if tokens > max_tokens:
+                    errors.append([i, f'Tokens count exceed model limit: {tokens} > {max_tokens}'])
```
paxcema, line 199: I would recommend adding a user-controlable "policy" with three options, where we can specify whether to keep the `first` or `last` `max_tokens` of rows where it's bigger than the maximum. This way, the failure mode may still be helpful if the user so chooses. By default, we can use this `strict` behavior.

paxcema (changes requested):

mindsdb/integrations/handlers/huggingface_handler/huggingface_handler.py
ea-rus: added truncation_policy param, values: left, right, strict(default)

paxcema (approved):

PR #4810, opened 2023-03-04
Google Cloud SQL Integration

MinuraPunchihewa:
I have implemented the new integration for Google Cloud SQL. After some initial testing, it looks to be working as expected. Please do let me know if there is anything that I have missed or if there are any bugs.


mindsdb/integrations/handlers/cloud_sql_handler/requirements.txt
```diff
@@ -0,0 +1,4 @@
+psycopg[binary] >= 1.15.3
+docker >= 5.0.3
```
ZoranPandovski, line 2: Do we need docker as dependency?

ZoranPandovski (changes requested):

mindsdb/integrations/handlers/cloud_sql_handler/requirements.txt
MinuraPunchihewa: I added this because the MYSQL handler is using it, Zoran. This makes use of the MYSQL handler.

ZoranPandovski: Hmm, strange. I am not sure why we need it, maybe for tests. I will check that. Thanks

ZoranPandovski (approved):

PR #4933, opened 2023-03-12
Apache Ignite Integration

MinuraPunchihewa:
I have implemented the new integration for Apache Ignite. After some initial testing, it looks to be working as expected. Please do let me know if there is anything that I have missed or if there are any bugs.


mindsdb/integrations/handlers/ignite_handler/ignite_handler.py
```diff
@@ -0,0 +1,238 @@
...
+            HandlerResponse
+        """
+
+        # if isinstance(query, ASTNode):
+        #     query_str = query.to_string()
+        # else:
+        #     query_str = str(query)
+        #
+        # return self.native_query(query_str)
+        renderer = SqlalchemyRender('ignite')
```
ZoranPandovski, line 170: @MinuraPunchihewa Are you sure this will work? Ignite is not supported dialect in the Renderer https://github.com/mindsdb/mindsdb_sql/blob/cac792efdc13762d8ab73e93006abd7b6dd5bafc/mindsdb_sql/render/sqlalchemy_render.py#L19

ZoranPandovski (changes requested):

mindsdb/integrations/handlers/ignite_handler/ignite_handler.py
MinuraPunchihewa: @ZoranPandovski Apologies, it looks like I have uncommented the wrong block of code when pushing this through. I have pushed a couple more commits, it should be working fine now. I just tested it again,
image

ZoranPandovski (approved):

PR #4936, opened 2023-03-12
Fixed Dependencies in MySQL and Postgres Handlers

MinuraPunchihewa:
The following dependency requirements have been resolved below,

- The `docker` dependency has been removed from the MySQL as it has not been used anywhere. The same requirement has been removed from two other handlers that use this handler: Amazon Aurora and Google Cloud SQL
- The `psycopg[binary]` dependency in the Postgres handler and all other handlers that inherit from it has been updated as the version specified does not seem to exist as mentioned here: https://github.com/mindsdb/mindsdb/issues/4555. I have tested out the Postgres handler with the new version of the dependency by spinning up a Docker container and it seems to be working fine.

### DIFF
--- a/mindsdb/integrations/handlers/aurora_handler/requirements.txt
+++ b/mindsdb/integrations/handlers/aurora_handler/requirements.txt
@@ -1,4 +1,3 @@
 psycopg[binary] >= 1.15.3
-docker >= 5.0.3
 mysql-connector-python
 boto3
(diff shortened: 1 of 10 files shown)


ZoranPandovski:
Looks great. Let's check with @ilia-tsyplenkov about the Docker dependency in case it is somehow needed for the new microservices architecture

MinuraPunchihewa:
Sounds good @ZoranPandovski.

ZoranPandovski (approved):

PR #5113, opened 2023-03-21
Statsforecast V4 - Implement describe() method

sammiller06:
## Description

Implements the `describe()` method for the StatsForecast handler.

## Type of change

### What is the solution?

New code and tests in the existing handler files.
Accuracy is based on in-sample cross-validation results.
See the specification for the method at https://docs.google.com/spreadsheets/d/1vm3N5f8IZZmuCyXhgP_-V1ZFDY8aCug8WfyxcQLpl20/edit#gid=0

## Checklist:

### Describe() method for default model (AutoARIMA)
image

### With auto selection - gives separate accuracy for each model
image

### With model attribute
image

### With feature attribute
image

### With ensemble attribute
image


tests/unit/ml_handlers/test_statsforecast.py
```diff
@@ -152,6 +164,27 @@ def test_model_choice(self, mock_handler):
         assert len(mindsdb_result) == prediction_horizon * n_groups
         assert np.allclose(mindsdb_result, package_predictions)
 
+        # test describe() method
```
paxcema, line 167: Just for future maintenance, would be useful to comment the expected data structure for nested fields (in particular, `accuracies` seems tricky to interpret just based on these assertions).

tests/unit/ml_handlers/test_statsforecast.py
```diff
@@ -198,3 +231,8 @@ def test_auto_model_selection(self, mock_handler):
         # check against ground truth
         mindsdb_result = result_df.iloc[:, -1]
         assert np.allclose(mindsdb_result, package_predictions)
+
+        describe_result = self.run_sql("describe proj.modelx")
+        assert len(describe_result["accuracies"][0]) > 1
```
paxcema, line 236: Same as in the comment above, would be nice to have a short explanation of the expected data structure.

tests/unit/ml_handlers/test_statsforecast.py
paxcema: Okay, debatable if it's truly necessary, going through the actual source of `describe` is enough. Definitely not critical, would be a nice to have.

paxcema (approved):
Looks good! Just two nitpicks, nothing critical.

tests/unit/ml_handlers/test_statsforecast.py
sammiller06: Think you're right, it's not much work to add.

PR #5396, opened 2023-04-03
Integration test for invalid openai name parameter

Fayezkhan1:
## Description

This is a test function that checks for an error when an invalid OpenAI model name is provided to the MindsDB Predictor. The function aims to ensure that the Predictor raises a ValueError exception with the message "Invalid model name" when the provided model name is not available in the OpenAI API.

**Fixes** #5112 

## Type of change

### What is the solution?
This test checks whether an error is raised when an invalid model name is provided while creating a MindsDB predictor that uses OpenAI engine for NLP.

The test first sets the OpenAI API key and fetches a list of available models. It then tries to create a MindsDB predictor with an invalid model name in the query parameter. The expected behavior is that a ValueError should be raised with a message containing the string "Invalid model name". The pytest.raises context manager is used to check if the expected error is raised.

This test is a part of the MindsDB test suite, which helps ensure that the MindsDB software works as expected and to prevent regressions when new changes are made to the code.
## Checklist:


Fayezkhan1:
> Can you please include this in https://github.com/mindsdb/mindsdb/blob/staging/tests/handler_tests/test_openai_handler.py
yeah sure

ZoranPandovski:
@Fayezkhan1, please re open this once you have time to resolve the suggestions

PR #5454, opened 2023-04-05
[tests] openai: add main completion flows  

paxcema:


mindsdb/interfaces/database/integrations.py
```diff
@@ -417,6 +418,12 @@ def _get_handler_meta(self, module):
     def _load_handler_modules(self):
         mindsdb_path = Path(importlib.util.find_spec('mindsdb').origin).parent
         handlers_path = mindsdb_path.joinpath('integrations/handlers')
+
```
ea-rus, line 421: when I remove this code test_openai.py  also works for me

paxcema: That is strange, what directory do you run them from, and how? I tried a few variations but it fails without this. Perhaps it's related to OS differences?

ea-rus: how do you run test and what error do you get?

paxcema: `python -m pytest -v ./unit/ml_handlers/test_openai.py` from the `tests` folder. It gives me this error without the code above:

```
___________________________________ ERROR at setup of TestOpenAI.test_prompt_template ____________________________________

self = <mindsdb.tests.unit.ml_handlers.test_openai.TestOpenAI object at 0x13ce439a0>

    def setup_method(self):
        super().setup_method()
>       self.set_executor()

unit/executor_test_base.py:143:
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
unit/executor_test_base.py:158: in set_executor
    integration_controller = IntegrationController()
../mindsdb/interfaces/database/integrations.py:33: in __init__
    self._load_handler_modules()
../mindsdb/interfaces/database/integrations.py:428: in _load_handler_modules
    for handler_dir in handlers_path.iterdir():
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

self = PosixPath('/MindsDB/mindsdb/integrations/handlers')

    def iterdir(self):
        """Iterate over the files in this directory.  Does not yield any
        result for the special paths '.' and '..'.
        """
>       for name in self._accessor.listdir(self):
E       FileNotFoundError: [Errno 2] No such file or directory: '/MindsDB/mindsdb/integrations/handlers'

/usr/local/Cellar/python@3.9/3.9.16/Frameworks/Python.framework/Versions/3.9/lib/python3.9/pathlib.py:1160: FileNotFoundError
```

In particular, the path above `/MindsDB/mindsdb/integrations/handlers` indeed does not exist. The right path would be: `/MindsDB/mindsdb/mindsdb/integrations/handlers`, which is what the fix above ensures.

PR #5476, opened 2023-04-05
SurrealDB integration

manosdell:
## Description

Implementation of SurrealDB integration

**Fixes** #4806

## Type of change

### What is the solution?
This handler is implemented using the `pysurrealdb` library which utilizes the `websocket` library for the database connectivity. Moreover, we have created a utils directory for a cleaner and easier to understand implementation of the get_tables() and get_columns() methods.

## Checklist:


mindsdb/integrations/handlers/surrealdb_handler/surrealdb_handler.py
```diff
@@ -0,0 +1,231 @@
...
+        Args:
+            query (ASTNode): sql query represented as AST. It may be any kind
+                of query: SELECT, INSERT, DELETE, etc
+        Returns:
+            HandlerResponse
+        """
+        query_string = query.to_string()
+
+        # ensure the correct query is passed
+        last_word = query_string.split()[-1]
```
ZoranPandovski, line 155: Thanks @manosdell. Can you please clarify why we need these two lines? Is there some existing surrealdb dialect that we can re-use for SQLAlchemy?

ZoranPandovski (changes requested):

mindsdb/integrations/handlers/surrealdb_handler/surrealdb_handler.py
ZoranPandovski: Apart from this looks great and it is ready for merge

manosdell: Unfortunately such dialect doesn't seem to exist yet. SurrealDB is still in Beta after all. I added the second line because otherwise the select query wouldn't return only the specified columns from the database.

ZoranPandovski: Ok, let's start with what we have. We can improve it after this. Thanks great job

PR #5479, opened 2023-04-05
[langchain] grounding via executor access

paxcema:
## Description

This PR modifies the Langchain behavior to use MindsDB's executor in one of two ways: 
1. As tools in the agent's toolkit (one for normal queries, one for metadata-related queries).
2. As a custom Langchain SQLDatabase abstraction

## Type of change

### What is the solution?

Main underlying change: we now pass the executor to the `handler.predict()` method and expose it to the "right" engines (for now, only Langchain, but other agent-based engines could be included eventually).

#### For #1 above:

Created a couple of Langchain tools that expose the executor's `query_execute()` method for valid SQL strings.

It is a good step towards a better integration, though the main drawback is that retrieval only really works when output tables are small and answers are in fact within the first (`< max_tokens`) records. Model/agent accuracy here is key.  

#### For #2 above:
We implement a custom class that inherits from `langchain.sql_database.SQLDatabase`, thus leveraging all the groundwork that Langchain offers for this type of agent.

## Example

```sql
CREATE MODEL langchain_model
PREDICT answer
USING
engine = 'langchain',
prompt_template='Answer the users question in a helpful way. Whatever intermediate steps you perform, observations should not be bigger than roughly one book page, otherwise you will run into errors. Here is the question: {{question}}. ',
modal_dispatch='sql_agent'; -- or 'default' for tool-based approach

SELECT answer
FROM langchain_model
WHERE question = 'What tables are available in my mindsdb connection?'
USING
verbose=True,
max_iterations=3;

SELECT answer
FROM langchain_model
WHERE question = 'What neighborhoods are found in the "example_db.demo_data.home_rentals" mindsdb table?'
USING
verbose=True,
max_iterations=3;

SELECT answer
FROM langchain_model
WHERE question = 'Could you describe the "example_db.demo_data.house_sales" mindsdb table? It is fine to look at the first few dozen records instead of the entire table.'
USING
verbose=True,
max_iterations=3;
```

More sample questions:

> 'Could you describe the table `example_db.arrivals` please?'

> 'Could you retrieve a single row from the table files.arrivals, specifying the name of each column please?'

> 'Could you give me the mean observed traffic from the table files.arrivals please?'

> 'I want to know the average number of beds in the downtown neighborhood as per the table `example_db.HomeRentals`'

> 'There is a property in the south_side neighborhood with an initial price of 2543 in the table `example_db.HomeRentals`. What are some other details of this listing?'

## Checklist:


mindsdb/integrations/libs/ml_exec_base.py
```diff
@@ -369,6 +369,9 @@ def predict(self, model_name: str, data: list, pred_format: str = 'dict',
             args['dtype_dict'] = predictor_record.dtype_dict
             args['learn_args'] = predictor_record.learn_args
 
+        if self.handler_class.__name__ in ('LangChainHandler',):
```
ea-rus, line 372: Instead I would put this code into _get_ml_handler
```
    command_executor = None
    if it is LangChainHandler:
         sql_session = SessionController()
         command_executor = ExecuteCommands(sql_session, None)
    ml_handler = self.handler_class(
                engine_storage=handlerStorage,
                model_storage=modelStorage,
                executor=command_executor,
     )
```

paxcema: That would require changing the signature of this method to return both the handler and its arguments, right? I'm not sure if that is desirable, perhaps an `_override_handler_args()` method would be better instead?

ea-rus: Do you mean signature of handler_class? If so it is encapsulated from developer and no need to change ml handlers

paxcema: Indeed, that's the one I meant. My concern is mostly semantics: if the method is called `handler_class`, isn't it confusing that it returns handler arguments as well?

I see above this code snippet there is similar logic for the lightwood handler, so my proposal is to take both of these and move them into a new method that would solely focus on overriding certain handler arguments depending on their name/type.

mindsdb/integrations/handlers/langchain_handler/mindsdb_database_agent.py
```diff
@@ -0,0 +1,150 @@
...
+            tables.append(table_info)
+
+        final_str = "\n\n".join(tables)
+        return final_str
+
+    def _get_single_table_info(self, table_str: str) -> str:
+        controller = self._metadata
+        integration, table_name = table_str.split('.')
+
+        tbl_name, n_rows, tbl_type = controller.get_handler(integration).get_tables().data_frame.iloc[0].to_list()
```
paxcema, line 100: Bug spotted by @ea-rus, we should be passing the table here.

PR #5605, opened 2023-04-12
Hierarchical forecast v2 - integration with Neuralforecast handler

sammiller06:
## Description

Implement hierarchical reconciliation for Nixtla's Neuralforecast handler.

## Type of change

### What is the solution?

We added hierarchical reconciliation using a similar method to the Statsforecast handler.

Also added a new unit test and updated the describe method.

## Checklist:

### Model creation
```
CREATE MODEL 
  mindsdb.nf_hier
FROM example_db
  (SELECT * FROM demo_data.house_sales)
PREDICT ma
ORDER BY saledate
GROUP BY type, bedrooms
-- as the target is quarterly, we will look back two years to forecast the next one
WINDOW 8
HORIZON 4
USING engine='neuralforecast', train_time=0.01, hierarchy=['type', 'bedrooms']
;
```
image

### Prediction
```
 SELECT m.saledate as date, m.ma as forecast
  FROM mindsdb.nf_hier as m 
  JOIN example_db.demo_data.house_sales as t
  WHERE t.saledate > LATEST and t.type='house' and t.bedrooms=4
  LIMIT 4;
```
image

### Describe
```DESCRIBE mindsdb.nf_hier.model;```
image

### DIFF
--- a/tests/unit/ml_handlers/test_neuralforecast.py
+++ b/tests/unit/ml_handlers/test_neuralforecast.py
@@ -140,3 +140,42 @@ class TestNeuralForecast(BaseExecutorTest):
 
         describe_features = self.run_sql("describe proj.model_exog_var.features")
         assert describe_features["exog_vars"][0] == ["exog_var_1"]
+
+    @patch("mindsdb.integrations.handlers.postgres_handler.Handler")
+    def test_hierarchical(self, mock_handler):
+        # create project
+        self.run_sql("create database proj")
+        df = pd.read_csv("tests/unit/ml_handlers/data/house_sales.csv")  # comes mindsdb docs forecast example
+        self.set_handler(mock_handler, name="pg", tables={"df": df})
+
+        self.run_sql(
+            """
+           create model proj.model_1_group
+           from pg (select * from df)
+           predict ma
+           order by saledate
+           group by type, bedrooms
+           horizon 4
+           window 8
+           using
+             engine='neuralforecast',
+             hierarchy=['type', 'bedrooms'],
+             train_time=0.01
+        """
+        )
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

File: tests/unit/ml_handlers/test_neuralforecast.py [hunk 1]
```diff
@@ -140,3 +140,41 @@ def test_with_exog_vars(self, mock_handler):
...
+           using
+             engine='neuralforecast',
+             hierarchy=['type', 'bedrooms'],
+             train_time=0.01
+        """
+        )
+        self.wait_predictor("proj", "model_1_group")
+
+        # run predict
+        mindsdb_result_hier = self.run_sql(
```
paxcema, line 169: I'm wondering whether it makes sense to train a non-hierarchical model in this test, and check that results are different between them? Presumably there is _some_ impact in the forecast, at least. 

The big question in my mind would be whether NF is deterministic (i.e. are seeds fixed) so that we can guarantee two runs in a row for the same model config would yield equal results? If not, then this addition doesn't make sense. If yes, then I'd ask that you include this so we know there is an impact.

paxcema (changes requested):
On the implementation itself: nicely done, straightforward. Builds out of components written in the PR for hierarchical statsforecast, so all good there.

Just one (potential) ask in terms of tests. Otherwise LGTM!

tests/unit/ml_handlers/test_neuralforecast.py
sammiller06: Indeed - the issue is that NF is not deterministic (unlike SF, which is why the test for diffs works there).

paxcema (approved):

PR #5631, opened 2023-04-14
Cache - Lightwood

rsjesse:
## Description

1. Lightwood may always pull/copy a model even though it may already exists. Can be improved by checking if already exists before copying.
2. Lightwood may repeatedly load the same predictor. Can be improved by caching. 

We are currently using these changes in our production servers. It has cut down predictor times from ~4 seconds down to ~0.03 seconds on the model load and prediction. A massive improvement.

**Fixes** #(issue)

## Type of change

- [X] Enhancement

### What is the solution?

1. Add caching when lightwood grabs the predictor.
2. Perform a rough file size check before attempting to copy

## Checklist:

- [X] I have commented my code, particularly in hard-to-understand areas.
- [X] I have updated the documentation, or created issues to update them.
- [X] I fixed|updated|added unit tests and integration tests for each feature (if applicable).
- [X] I have shared a short loom video or screenshots demonstrating any new functionality.

### DIFF
--- a/mindsdb/interfaces/storage/fs.py
+++ b/mindsdb/interfaces/storage/fs.py
@@ -92,10 +92,10 @@ class LocalFSStore(BaseFSStore):
 
     def get(self, local_name, base_dir):
         remote_name = local_name
-        copy(
-            os.path.join(self.storage, remote_name),
-            os.path.join(base_dir, local_name)
-        )
+        src = os.path.join(self.storage, remote_name)
+        dest = os.path.join(base_dir, local_name)
+        if not os.path.exists(dest) or os.path.getsize(src) != os.path.getsize(dest):
+            copy(src, dest)
 
     def put(self, local_name, base_dir):
         remote_name = local_name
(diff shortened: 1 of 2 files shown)


paxcema (approved):
This is _awesome_ @rsjesse! That's a huge improvement. Let's check if the CI tests pass, but otherwise code looks good to me. @ea-rus I suggest you take a look as well, just in case.

mindsdb/interfaces/storage/fs.py
```diff
@@ -92,10 +92,10 @@ def __init__(self):
 
     def get(self, local_name, base_dir):
```
ea-rus, line 93: It is a common function, not used only by predict method
I would update predict method and instead line `self.model_storage.fileStorage.pull()`:
- use `self.model_storage.fileStorage.pull_path(update=False)` if it will work
- or update pull method and use `self.model_storage.fileStorage.pull(update=False)`

StpMax: I agree with Andrey, `fs.get` should be improved. But for now i think that can be merged as is

StpMax (approved):

PR #5704, opened 2023-04-18
StatsForecast - minor frequency patch

sammiller06:
## Description

Pandas sometimes calls monthly frequency "MS" (month start). 

This was omitted in the first implementation, so it assumed season length of 1 rather than 12 for MS data.

## Type of change

### What is the solution?

Added row for the new frequency.

## Checklist:

### DIFF
--- a/mindsdb/integrations/handlers/statsforecast_handler/statsforecast_handler.py
+++ b/mindsdb/integrations/handlers/statsforecast_handler/statsforecast_handler.py
@@ -34,6 +34,7 @@ def get_season_length(frequency):
     season_dict = {  # https://pandas.pydata.org/docs/user_guide/timeseries.html#timeseries-offset-aliases
         "H": 24,
         "M": 12,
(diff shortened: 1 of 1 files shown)


mindsdb/integrations/handlers/statsforecast_handler/statsforecast_handler.py
```diff
@@ -34,6 +34,7 @@ def get_season_length(frequency):
     season_dict = {  # https://pandas.pydata.org/docs/user_guide/timeseries.html#timeseries-offset-aliases
         "H": 24,
         "M": 12,
+        "MS": 12,
```
paxcema, line 37: Good catch, I've ran into this previously. I recall there's also the option for doing "[X]F", with `X` an integer and `F` the frequency. That would definitely be a new feature though.

paxcema (approved):
Looks good! Let's just wait on the one test that timed-out before merging.

PR #5785, opened 2023-04-25
Google Books Integration

panos-span:
## Description

**Fixes** #5416 

## Type of change

### What is the solution?

I have created a class for the google books handler and for the Bookshelves and Volumes API Table and the connection is achieved using the libraries `google-auth-oauthlib`, `google-auth-httplib2` and `google-api-python-client`.

## Checklist:


mindsdb/integrations/handlers/google_books_handler/README.md
```diff
@@ -0,0 +1,83 @@
+# Google Books API Integration
+
+This handler integrates with the [Google Books API](https://developers.google.com/books/docs/overview) to allow you to
+make book and bookshelf data available to use for model training and predictions.
+
+## Example: Automate your book recommendations
+
+To see how the Google Calendar handler is used, let's walk through a simple example to create a model to predict
```
ZoranPandovski, line 8: Please change the Google Clanedar

ZoranPandovski (changes requested):
@panos-span Wouldn't be better if we use the https://github.com/googleapis/google-api-python-client ?

mindsdb/integrations/handlers/google_books_handler/README.md
panos-span: Yes my bad

mindsdb/integrations/handlers/google_books_handler/google_books_handler.py
```diff
@@ -0,0 +1,185 @@
...
+                self.credentials = Credentials.from_authorized_user_file('token_books.json', self.scopes)
+            if not self.credentials or not self.credentials.valid:
+                if self.credentials and self.credentials.expired and self.credentials.refresh_token:
+                    self.credentials.refresh(Request())
+                else:
+                    flow = InstalledAppFlow.from_client_secrets_file(
+                        self.credentials_file, self.scopes)
+                    self.credentials = flow.run_local_server(port=0)
+            # Save the credentials for the next run
+            with open('token_books.json', 'w') as token:
```
ZoranPandovski, line 70: Should we check for permission here? I am thinking about use cases when mindsdb is deployed on cloud instances?

ZoranPandovski (changes requested):

mindsdb/integrations/handlers/google_books_handler/google_books_handler.py
panos-span: A potential simple and easy fix , since the credentials file size is very small, could be that the user uploads the wanted file and then gets the file content through a simple query (for example `SELECT * from file.credentials`) and then give the contents of the file as input.

ZoranPandovski (approved):
Amazing, thanks @panos-span

PR #5996, opened 2023-05-04
Use args for BYOM

### LINKED ISSUE #5747: Support for USING arguments when training and calling BYOM models

Current signatures expect only a dataframe as input, but it may be desirable to modify the behavior of these calls with USING-level arguments as we do in non-BYOM models right now.

### Video or screenshots

_No response_

_No response_

### Anything else?

_No response_

ea-rus:
Using:

1. Add args argument to train and predict methods:
```
class CustomPredictor():
    def train(self, df, target_col, args=None):
       ...
       
    def predict(self, df, args=None):
       ...
```

2. Passing args to model training
``` 
create predictor pred 
from files (select * from byom)
predict Time
using engine='uploaded_model',
  param1=1, param2='2'
```

3. Passing args at predict
```
select * from files.byom t 
join pred3 p
using param1=1, param2='2'
```

fixes #5747

## Description

**Fixes** #(issue)

## Type of change

### What is the solution?

## Checklist:

### DIFF
--- a/mindsdb/integrations/handlers/byom_handler/proc_wrapper.py
+++ b/mindsdb/integrations/handlers/byom_handler/proc_wrapper.py
@@ -109,8 +109,13 @@ def main():
     if method == 'train':
         df = pd_decode(params['df'])
         to_predict = params['to_predict']
+        args = params['args']
         model = model_class()
-        model.train(df, to_predict)
+
+        call_args = [df, to_predict]
+        if args:
+            call_args.append(args)
+        model.train(*call_args)
 
         # return model
         data = model.__dict__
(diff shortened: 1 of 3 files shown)


paxcema (approved):

mindsdb/integrations/handlers/byom_handler/proc_wrapper.py
```diff
@@ -121,11 +126,15 @@ def main():
...
         df = pd_decode(params['df'])
+        args = params['args']
 
         model = model_class()
         model.__dict__ = decode(model_state)
 
-        res = model.predict(df)
+        call_args = [df]
+        if args:
+            call_args.append(args)
```
paxcema, line 136:
```python
else:
    call_args.append({})
```

Alternatively, replace the entire `if` with a one-liner:

```python
call_args.append(args if args else {})
```

PR #6038, opened 2023-05-09
Shopify App Integration

MinuraPunchihewa:
I have implemented the new integration for Shopify. For now, I have added support for querying Products (products), Customers (customers) and Orders (orders) using the handler.

After some initial testing, it looks to be working as expected. Please do let me know if there is anything that I have missed or if there are any bugs.

I will be adding a README outlining how this handler has been implemented first thing in the morning.


MinuraPunchihewa:
A README has also been added.

ZoranPandovski:
Can you please make sure this is up to date with staging so we can avoid torch errors

MinuraPunchihewa:
@ZoranPandovski I just merged the code from staging, but the status does not seemed to have changed?

MinuraPunchihewa:
@ZoranPandovski I just merged the code from staging, but the status does not seemed to have changed?

ZoranPandovski (approved):

PR #6166, opened 2023-05-17
[Bug]: fix tests structure so that users don't need to manually add i…

SebastianJames55:
## Description

- Added init file in tests folder

**Fixes** #5688

## Type of change

### What is the solution?

Init file was missing in tests folder


tests/__init__.py
paxcema, line 1: Hmm, tests seem to be broken with this. @ZoranPandovski any ideas?

SebastianJames55: @paxcema was this change the expected fix?

paxcema: Code wise, personally, yes. However, if tests are not passing we can't merge this. Paging @ZoranPandovski again in case he missed this.

SebastianJames55: @paxcema thank you for the update. I also tried the following but to no effect:
configuring pyproject.toml like
```
[tool.pytest.ini_options]
python_paths = ["./mindsdb/utilities"]
```
also added a pytest.ini for pytest to read from pyproject.toml
```
[pytest]
python_paths = ./mindsdb/utilities
```
I'm still getting the same error

ZoranPandovski:
Closing for now, if we find a way to make tests working, we can re open

SebastianJames55:
Understood, thank you for the update.

PR #6374, opened 2023-05-30
added explanation for persisting predictions

martyna-mindsdb:
## Description

added explanation for persisting predictions

**Fixes** #-

## Type of change

### What is the solution?

## Checklist:

### DIFF
--- a/docs/faqs/persist-predictions.mdx
+++ b/docs/faqs/persist-predictions.mdx
@@ -16,9 +16,9 @@ FROM mysql_demo_db.amazon_reviews AS input
 JOIN sentiment_classifier AS output;
 ```
 
-But there are ways to save predictions, so you don't need to run this query every time.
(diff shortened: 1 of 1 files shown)


docs/faqs/persist-predictions.mdx
```diff
@@ -5,6 +5,22 @@ sidebarTitle: Persisting Predictions
 
 MindsDB provides a range of options for persisting predictions and forecasts. Let's explore all possibilities to save the prediction results.
 
+<Note>
+**Reasons to Save Predictions**
+
+Every time you want to get predictions, you need to query the model, usually joined with an input data table, like this:
```
ZoranPandovski, line 11: Maybe we can explain that querying the model by default returns the result set which is not persistent. For future use, it is better to persist the result-set instead of querying the model again with the same data. What do you think?

martyna-mindsdb: Good point. I'll add it up.

PR #6614, opened 2023-06-15
Patch Describe syntax bug

parthiv11:
## Description

- while using describe predictor mindsdb.aiforever.features attribute was always None, I made some changes in mindsdb\api\mysql\mysql_proxy\executor\executor_commands.py then it works fine.

### DIFF
--- a/mindsdb/api/mysql/mysql_proxy/executor/executor_commands.py
+++ b/mindsdb/api/mysql/mysql_proxy/executor/executor_commands.py
@@ -694,12 +694,11 @@ class ExecuteCommands:
         # try full name
         attribute = None
         model_info = self._get_model_info(statement.value)
+        parts = statement.value.parts.copy()
+        attribute = parts.pop(-1)
+        model_info = self._get_model_info(Identifier(parts=parts)) if not model_info  else model_info
         if model_info is None:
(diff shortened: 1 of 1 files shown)


ea-rus:
Is 'aiforever' the lightwood model?

parthiv11:
No it is model from new handler I am working on Replicate Handler.

#6795

mindsdb/api/mysql/mysql_proxy/executor/executor_commands.py
```diff
@@ -694,12 +694,11 @@ def answer_describe_predictor(self, statement):
         # try full name
         attribute = None
         model_info = self._get_model_info(statement.value)
+        parts = statement.value.parts.copy()
+        attribute = parts.pop(-1)
+        model_info = self._get_model_info(Identifier(parts=parts)) if not model_info  else model_info
```
ea-rus, line 699: here two _get_model_info are called in sequence. 
The right way is: to call once,  and call once again with different name only if model wasn't found.
(As it was before)
Try this PR #6686 (or update from staging, because it was merged long ago) - it should help you

ea-rus (changes requested):

PR #6685, opened 2023-06-27
INSERT Support for Shopify Integration

MinuraPunchihewa:
This Pull Request adds support for `INSERT` statements to be executed on the Shopify Customers table. This is facilitated by a new query utility that has been added for parsing `INSERT` queries called `INSERTQueryParser`.

After some initial testing, it looks to be working as expected. Please do let me know if there is anything that I have missed or if there are any bugs.

Given below is an example of how INSERT queries can be run
image

In this initial implementation, only a subset of the Customer attributes are supported.
,


mindsdb/integrations/handlers/shopify_handler/shopify_tables.py
```diff
@@ -99,6 +104,32 @@ def select(self, query: ast.Select) -> pd.DataFrame:
...
+        None
+
+        Raises
+        ------
+        ValueError
+            If the query contains an unsupported condition
+        """
+        insert_statement_parser = INSERTQueryParser(
+            query,
+            supported_columns=['first_name', 'last_name', 'email', 'phone', 'tags', 'currency'],
```
ZoranPandovski, line 126: Can we add this to the README. New section for INSERT and the required values?

ZoranPandovski (changes requested):

ZoranPandovski (approved):

PR #6781, opened 2023-07-05
add optional cpu huggingface requirements file

### LINKED ISSUE #6538: Using cpu-only torch

As far as I see, we have to use the version of the torch using Nvidia GPU and its packages. I think it's a good idea to let the user specify whether they want to use a CPU-only version if possible.

### Video or screenshots

_No response_

We need to have:

```
--index-url https://download.pytorch.org/whl/cpu
torch==2.0.1 
```

in our **requirements.txt**. Unfortunately, our setup.py gets stuck when the package is introduced like that. Furthermore, in other dependencies GPU version of the torch is already in use. So, we need to refactor multiple dependencies.

### Anything else?

_No response_

hamishfagg:
Also modifies setup.py to handle extra requirements files in the form `requirements<x>.txt` inside a handler folder. This allows us multiple setups of a handler (like `huggingface_cpu`, which installs the cpu version of torch instead of the gpu version).

fixes #6538

### DIFF
--- a/mindsdb/integrations/handlers/huggingface_handler/requirements_cpu.txt
+++ b/mindsdb/integrations/handlers/huggingface_handler/requirements_cpu.txt
@@ -1,0 +1,9 @@
+mindsdb>=22.6.2.1
+mindsdb_sql >= 0.4.0
+pandas <=1.4.3
+transformers==4.21.0
+sentencepiece==0.1.97
+
+tensorflow
+tqdm
+torch==2.0.1+cpu --index-url https://download.pytorch.org/whl/cpu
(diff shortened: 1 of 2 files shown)


mindsdb/integrations/handlers/huggingface_handler/requirements_cpu.txt
```diff
@@ -0,0 +1,10 @@
+mindsdb>=22.6.2.1
```
ea-rus, line 1: why mindsdb dependency here? handlers don't use mindsdb

ea-rus (changes requested):

ea-rus:
So for huggingface with cpu command is going to be like this, right?
`pip install mindsdb[huggingface_cpu]`

hamishfagg:
@ea-rus ive just updated this so take another look.

unfortunately it has to be installed with `pip install --extra-index-url https://download.pytorch.org/whl/ .[huggingface_cpu]`, there's no way around having to specify the extra URL.

hamishfagg:
fixed by #8395

PR #6807, opened 2023-07-11
More consistent/useful logging

hamishfagg:
## Description

Updates/simplifies logging configuration and adds log messages to make debugging issues easier.

- Adds process name and module name to log messages to show where messages are coming from
- Convert 'print' statements to proper logging
- Unify logging format and invocation everywhere
- Remove import error logs from the few handlers that had them
- Allow logging of HTTP requests via a generic flask handler

Configuration of logging is now implicit as long as devs import and use logging like so:

```python
from mindsdb.utilities import log

logger = log.getLogger(__name__)

logger.info("Hi! :)")
```

I would prefer to use the `logging` library itself in general code, but it's very hard to track down where MindsDB spawns new processes and configure logging for each process. With the approach here, devs don't have to worry about it - `getLogger()` will do everything needed and return a logger object (this function is also compatible with `logging.getLogger()` on purpose to make later changes easier).

Log level is configurable via the `MINDSDB_LOG_LEVEL` env var.

Since there are a lot of changes in this PR the actual configuration of logging is [here](https://github.com/mindsdb/mindsdb/blob/552f9e3665f76e76e04c0b4e04e5c37f60607d30/mindsdb/utilities/log.py)

## TODO
- [ ] Reimplement logging to file

## Checklist:


mindsdb/utilities/log.py
```diff
@@ -1,151 +1,78 @@
...
+def getLogger(name=None):
+    """
+    Get a new logger, configuring logging first if it hasn't been done yet.
+    """
+    global logging_initialized
+    if not logging_initialized:
+        configure_logging()
+        logging_initialized = True
+
+    return logging.getLogger(name)
```
QuantumPlumber, line 78: This line will create a child logger of the root logger. It should be something like `logging.getLogger("mindsdb").getChild(name)`

QuantumPlumber (approved):
Passes all tests and is a big improvement to logging. Cheers.

PR #6813, opened 2023-07-11
Chromadb handler

dusvyat:
## Description

Handler for [ChromaDB](https://www.trychroma.com/)

In summary, it currently supports the following:

- Count of number of docs in a collection
- Select n docs from a collection
- Create a new collection using insert with select. This will embed the source data from the select query and add to a new collection
- Filter a collection based on a query text
- Filter a collection on metadata

[Docs](https://docs.google.com/document/d/1SZEPC1xgRLg22Hd97QmMvMRy2haoxreCKXj_AEnkTLM/edit#heading=h.yvz44edfc1i5)

[Demo](https://www.loom.com/share/2d2a38a731114a54b41e3486ab809c26?sid=5d7e1250-ddb7-472d-963f-cfc74067212b)

**Fixes** #(issue)

## Type of change

### What is the solution?

## Checklist:


mindsdb/integrations/libs/base.py
```diff
@@ -105,6 +106,15 @@ def __init__(self, name: str):
         super().__init__(name)
 
 
+class VectorStoreHandler(BaseHandler):
```
dusvyat, line 109: this is the only change done, the rest are just linting

mindsdb/integrations/handlers/chromadb_handler/requirements.txt
```diff
@@ -0,0 +1,7 @@
+langchain==0.0.186
+SQLAlchemy
+duckdb==0.7.1
+chromadb==0.3.25
+requests==2.28.0
+sentence_transformers
```
paxcema, line 6: Perhaps add a lower bound if the default embedding model needs any?

mindsdb/integrations/libs/base.py
```diff
@@ -105,6 +105,15 @@ def __init__(self, name: str):
...
+    """
+    Base class for handlers associated to vector databases.
+    """
+
+    def __init__(self, name: str):
```
paxcema, line 113: @tmichaeldb FYI ^

mindsdb/integrations/handlers/chromadb_handler/chromadb_handler.py
```diff
@@ -0,0 +1,368 @@
...
+
+        Args:
+            query (ASTNode): The query to filter.
+
+        Returns:
+            Dict: filter for collection.
+        """
+
+        where = {}
+        if query.where.op == "and":
```
paxcema, line 136: I'm having a bit of trouble reading this 😅 . Would it work for an `AND` chain? e.g. `WHERE x=x1 AND y=y1 AND z=z1`? Not a must have, just curious (and would be good to document in the handler's `README.md` perhaps).

mindsdb/integrations/handlers/chromadb_handler/chromadb_handler.py
```diff
@@ -0,0 +1,368 @@
...
+            else:
+                # if no 'search_query' provided, run a regular query and filter on metadata if provided
+                collection_data = self._collection.get(
+                    where=where.get("meta_data_filter")
+                )
+                result = pd.DataFrame(collection_data)[
+                    ["ids", "documents", "metadatas"]
+                ]
+
+        if query.limit:
```
paxcema, line 231: Nitpick: given that `query.limit` affects the value of `k` if it exists, shouldn't this `if` go inside the `else` clause above?

mindsdb/integrations/handlers/chromadb_handler/chromadb_handler.py
```diff
@@ -0,0 +1,368 @@
...
+    split_documents,
+)
+from mindsdb.integrations.libs.base import VectorStoreHandler
+from mindsdb.integrations.libs.const import HANDLER_CONNECTION_ARG_TYPE as ARG_TYPE
+from mindsdb.integrations.libs.response import RESPONSE_TYPE
+from mindsdb.integrations.libs.response import HandlerResponse as Response
+from mindsdb.integrations.libs.response import HandlerStatusResponse as StatusResponse
+from mindsdb.utilities import log
+
+# todo create separate util dir for vectorstore
```
paxcema, line 26: From today's discussion: we'll wait for the PgVector handler before refactoring anything out.

mindsdb/integrations/handlers/chromadb_handler/chromadb_handler.py
```diff
@@ -0,0 +1,368 @@
...
+                resp_type=RESPONSE_TYPE.ERROR,
+                error_message=f"Error executing query on ChromaDB client, {e}!",
+            )
+
+    def native_query(self, query: str) -> Response:
+        """
+        Execute a native query on the ChromaDB database.
+        """
+
+        # todo fix this - not sure if we need this?
```
paxcema, line 299: Depends on whether we need to support this:

```sql
SELECT * FROM (
    SELECT * FROM ...
)
```

Which I don't think we do? At least not with the scope we've initially agreed on. Perhaps let's replace with a `NotImplementedError` for now, assuming it doesn't break the rest of the handler.

mindsdb/integrations/handlers/chromadb_handler/chromadb_handler.py
```diff
@@ -0,0 +1,368 @@
...
+                error_message=f'Error getting tables from ChromaDB client {self._connection_data["chroma_server_host"]}, {e}!',
+            )
+
+        return Response(
+            resp_type=RESPONSE_TYPE.TABLE,
+            data_frame=collections_name,
+        )
+
+
+connection_args = OrderedDict(
```
paxcema, line 344: Should these (args and example) be here? I can't find usages apart from the import checks.

Perhaps let's move them to some file (e.g. `chromadb_handler/constants.py`?) and mention them in the `README.md` for any interested contributors to be aware that these are the definitions.

mindsdb/integrations/handlers/chromadb_handler/helpers.py
```diff
@@ -0,0 +1,145 @@
...
+import torch
+from langchain.document_loaders import DataFrameLoader
+from langchain.embeddings import HuggingFaceEmbeddings
+from langchain.schema import Document
+from langchain.text_splitter import RecursiveCharacterTextSplitter
+
+from mindsdb.utilities import log
+
+
+def extract_collection_name(sql_query):
```
paxcema, line 15: Ideally add unit tests for this.

mindsdb/integrations/handlers/chromadb_handler/chromadb_handler.py
dusvyat: I copied this from DuckDB handler, its used as a data input validation via UI. 

If param is entered that is not listed here it will raise an error via UI

PR #6819, opened 2023-07-12
Triggers & Mongodb Embeddings

ea-rus:
### New features:

**Embeddings mode for openai**

To create model that will be returning embeddings on output using openai api:
```sql
CREATE MODEL <name>
    PREDICT embedding
    USING
    engine = 'openai',
    mode='embedding',
    model_name='text-embedding-ada-002', -- name of embedding model
    question_column='summary';  -- input column name
```

**New update syntax**

```sql
 update <database>.<table>
    on <key column> [, <key column>]
    from
    (
       <query>
    )
```
It can be used for update data in integration table using result of execution of other query. All input will be translated to series of UPDATE TABLE ... SET col = ... WHERE key_col = ...
'key_col' in WHERE will be used from  <key column> list. All the rest columns from result of subselect will be used in SET clause.

Example:
```sql
 update db1.tbl1
    on id
    from
    (
        select t.id, m.prediction from db2.tbl2 t
        join model m
    )
```
It will update prediction column in db1.tbl1 using id column  

### Triggers

**Create trigger syntax:**
```sql
create trigger <name> 
on <integration name>.<table name>
[columns <column name> [, <column name>]  ]
 (
    <sql>
) 
```

Trigger will use 'subscribe' method of the data handler.
At the moment it is implemented only for mongodb handler and works only on cluster-set. 
Inside of trigger it is possible to use special variable TABLE_DELTA. It will content table with one row of data that was received from data hander (updated or inserted row)

Overall example.
```sql
  create trigger tr_embed 
   on db1.listingsAndReviews
   columns summary
    (
        update db1.listingsAndReviews
        on _id
        from
        (
        select t._id, m.embedding from TABLE_DELTA t
            join openai_emb m
        )
    )
```

**List of triggers**

```sql
    select * from mdb_triggers
```

**Delete trigger**

```sql
    drop trigger tr_embed 
```

### Other changes:

- created tasks engine
  - replaces chatbot process
  - runs different types of tasks in thread
- created triggers engine:
   - manage triggers create/drop/select
   - run trigger in task

https://www.loom.com/share/da9ae1cf8b7640f8b2ea867de57ffc3b

## Type of change

### What is the solution?

## Checklist:


mindsdb/interfaces/storage/db.py
```diff
@@ -334,3 +334,38 @@ class ChatBotsHistory(Base):
...
+    __tablename__ = 'tasks'
+    id = Column(Integer, primary_key=True)
+    company_id = Column(Integer)
+    user_class = Column(Integer, nullable=True)
+
+    # trigger, chatbot
+    object_type = Column(String, nullable=False)
+    object_id = Column(Integer, nullable=False)
+
+    last_error = Column(String)
```
StpMax, line 364: Looks like data in table could be updated. In that case let add `updated_at` field

StpMax: may be also add to `triggers`

mindsdb/interfaces/tasks/task_monitor.py
```diff
@@ -0,0 +1,139 @@
...
+
+                db.session.rollback()  # disable cache
+                time.sleep(self.MONITOR_INTERVAL_SECONDS)
+
+            except (SystemExit, KeyboardInterrupt):
+                self.stop_all_tasks()
+                return
+
+            except Exception as e:
+                log.logger.error(e)
```
StpMax, line 43: may be add `rollback()` here?

StpMax (approved):

mindsdb/interfaces/storage/db.py
ea-rus: done in chatbot-unify branch

mindsdb/interfaces/tasks/task_monitor.py
ea-rus: added in chatbot-unify branch

PR #6912, opened 2023-07-21
Decouple Langchain from OpenAI and add Claude Support

tmichaeldb:
## Description

Since `langchain_handler` and `openai_handler` don't share much common functionality, it makes sense to decouple them completely. This PR refactors `langchain_handler` and adds support to use [Anthropic Claude and Claude Instant](https://www.anthropic.com/index/introducing-claude) models.

One thing to note is that Anthropic models (and several others) don't always conform to the output format Langchain expects (when not using tools). These models, even when being prompted to return output in a specific format, will still sometimes return the final answer as text. As a workaround for this, we handle parsing errors ourselves, and look out for this special case.

In the future, it would benefit us to write our own `langchain.schema.BaseOutputParser` to be more robust and flexible.

Note: This feature requires `langchain==0.0.238` and `tokenizers` pinned to `0.12.1` (for `anthropic` package)

**Closes** #6822 
**Closes** #6823 

## Type of change

## Checklist:

### DIFF
--- a/mindsdb/integrations/handlers/langchain_handler/langchain_handler.py
+++ b/mindsdb/integrations/handlers/langchain_handler/langchain_handler.py
@@ -50,7 +50,10 @@ class LangChainHandler(BaseMLEngine):
         self.generative = True
         self.stops = []
         self.default_mode = 'default'  # can also be 'conversational' or 'conversational-full'
-        self.supported_modes = ['default', 'conversational', 'conversational-full', 'image']
+        self.engine_to_supported_modes = {
+            'openai': ['default', 'conversational', 'conversational-full', 'image'],
+            'anthropic': ['default', 'conversational', 'conversational-full']
+        }
         self.default_model = _DEFAULT_MODEL
         self.default_max_tokens = _DEFAULT_MAX_TOKENS
         self.default_agent_model = _DEFAULT_AGENT_MODEL
@@ -88,8 +91,12 @@ class LangChainHandler(BaseMLEngine):
 
         if not args.get('mode'):
             args['mode'] = self.default_mode
-        elif args['mode'] not in self.supported_modes:
-            raise Exception(f"Invalid operation mode. Please use one of {self.supported_modes}")
+
+        supported_modes = self.engine_to_supported_modes['openai']
+        if args['model_name'] in _ANTHROPIC_CHAT_MODELS:
+            supported_modes = self.engine_to_supported_modes['anthropic']
+        if args['mode'] not in supported_modes:
+            raise Exception(f"Invalid operation mode. Please use one of {supported_modes}")
 
         self.model_storage.json_set('args', args)
 


mindsdb/integrations/handlers/langchain_handler/langchain_handler.py
```diff
@@ -302,6 +329,20 @@ def _completion(agent, prompts):
...
+                    if not response.startswith(_PARSING_ERROR_PREFIX):
+                        completions.append(f'agent failed with error:\n{str(e)[:50]}...')
+                    else:
+                        # By far the most common error is a Langchain parsing error. Some OpenAI models
+                        # always output a response formatted correctly. Anthropic, and others, sometimes just output
+                        # the answer as text (when not using tools), even when prompted to output in a specific format.
+                        # As a somewhat dirty workaround, we accept the output formatted incorrectly and use it as a response.
+                        #
+                        # Ideally, in the future, we would write a parser that is more robust and flexible than the one Langchain uses.
+                        response = response.removeprefix(_PARSING_ERROR_PREFIX).removesuffix('`')
```
paxcema, line 344: Nice workaround!

mindsdb/integrations/handlers/openai_handler/openai_handler.py
```diff
@@ -16,10 +16,10 @@
 
 from mindsdb.utilities.hooks import before_openai_query, after_openai_query
 from mindsdb.utilities import log
-from mindsdb.utilities.config import Config
 from mindsdb.integrations.libs.base import BaseMLEngine
 from mindsdb.integrations.handlers.openai_handler.helpers import retry_with_exponential_backoff, \
     truncate_msgs_for_token_limit
+from mindsdb.integrations.utilities.handler_utils import get_api_key
 
 CHAT_MODELS = ('gpt-3.5-turbo', 'gpt-3.5-turbo-0301', 'gpt-4', 'gpt-4-0314', 'gpt-4-32k', 'gpt-4-32k-0314')
```
paxcema, line 24: Not necessarily in the scope of this PR, but would be great to update the `-0314` and `-0301` models with the most recent counterparts listed [here](https://platform.openai.com/docs/models/moderation).

mindsdb/integrations/utilities/handler_utils.py
```diff
@@ -0,0 +1,55 @@
+import os
+from typing import Dict
+
+from mindsdb.interfaces.storage.model_fs import HandlerStorage
+from mindsdb.utilities.config import Config
+
+"""Contains utilities to be used by handlers."""
+
+
+def get_api_key(
```
paxcema, line 10: Yes! This was sorely needed. @ZoranPandovski I'll open up a community issue once this is merged so that all other existing ML handlers that have copy-pasted this bit of code will now use the util instead 👍

paxcema (approved):
Looks good to me, I've left a couple nitpicks but we could merge as-is. May also be good to open an issue to update docs where needed.

mindsdb/integrations/handlers/openai_handler/openai_handler.py
paxcema: Agreed, opened up #6925 for tracking

PR #6931, opened 2023-07-25
Email Integration

MinuraPunchihewa:
Hey @ZoranPandovski, @torrmal,
I have updated the implementation of the Email integration and I believe that I have it working now. I was looking at this in order to rewrite the Gamil handler and I thought that fixing this would be a good starting point.

Given below are some queries that I ran using this implementation and they seem to be working fine,
image

Please note that in order to get this working for Gmail I had to use app passwords,
https://support.google.com/accounts/answer/185833?hl=en

Please let me know what you guys think. I have left a few TODOs and I also need to updated the README, I believe. If you guys think that I am on the right track, I can finish up those as well.

@torrmal I also had a quick question regarding the parameter called `since_emailid` here,
```
def search_email(self, mailbox="INBOX", subject=None, to=None, from_=None, since_date=None, until_date=None,
                 since_emailid=None):
```
Does this refer to the individual IDs of the emails or is it something else?


mindsdb/integrations/handlers/email_handler/email_helpers.py
```diff
@@ -0,0 +1,83 @@
...
+    def __init__(self, email, password, smtp_server='smtp.gmail.com', smtp_port=587, imap_server="imap.gmail.com"):
+        self.email = email
+        self.password = password
+        self.smtp_server = smtplib.SMTP(smtp_server, smtp_port)
+        self.imap_server = imaplib.IMAP4_SSL(imap_server)
+
+    def send_email(self, to_addr, subject, body):
+        msg = MIMEMultipart()
+        msg['From'] = self.email
+        msg['To'] = to_addr
```
ZoranPandovski, line 19: Is this working with multiple recipients?

mindsdb/integrations/handlers/email_handler/email_helpers.py
```diff
@@ -0,0 +1,83 @@
...
+
+        query = ' '.join(query_parts)
+
+        ret = []
+
+        resp, items = self.imap_server.uid('search', None, query)
+        items = items[0].split()
+        for emailid in items[::-1]:
+            resp, data = self.imap_server.uid('fetch', emailid, "(BODY[HEADER.FIELDS (SUBJECT TO FROM DATE)])")
+            raw_email = data[0][1].decode("utf-8")
```
ZoranPandovski, line 59: Maybe we can check for `UnicodeDecodeError` exception here in case the text contains other chars

mindsdb/integrations/handlers/email_handler/email_helpers.py
```diff
@@ -0,0 +1,83 @@
...
+
+            ret.append(email_line)
+
+        self.imap_server.logout()
+
+        return pd.DataFrame(ret)
+
+
+if __name__ == "__main__":
+    email_client = EmailClient()
```
ZoranPandovski, line 79: Do we need the email and pass here?

ZoranPandovski (changes requested):
Great progress. Please take a look at the comments above. Also, can we please implement a login/logout function to ensure the SMTP connection is properly closed?

mindsdb/integrations/handlers/email_handler/email_helpers.py
MinuraPunchihewa: I believe this is unnecessary, Zoran. I will remove it.

mindsdb/integrations/handlers/email_handler/email_helpers.py
MinuraPunchihewa: Yes, it works with a comma separated string. This is an example,
```
INSERT INTO email_datasource.emails(to, subject, body)
VALUES
("minurapunchihewa17@gmail.com, natashamp97@gmail.com", "TEST", "TEST")
```

mindsdb/integrations/handlers/email_handler/email_helpers.py
MinuraPunchihewa: Great idea! I've added it.

MinuraPunchihewa:
I have included an updated README as well.

PR #6964, opened 2023-07-26
Unknown argument validation + fix openai handler test

yuhuishi-convect:
## Description

When creating the openAI handler, we validate the arguments provided by a user against a known list of arguments.
We throw an error an unexpected argument was found.

This solution is a bit hacky, in the sense we need to keep maintaining and updating the right argument list on the handler level. I am proposing a [design doc](https://docs.google.com/document/d/1mE64JEXmn0Kz60c0_qVAWhjvNKCpiNvg5wVLxHPoKjY/edit?usp=sharing) to ask handler developers to self declare and document
the argument specifications in the handler implementation.

**Fixes** #6845

Error message if wrong arguments are provided:
image

## Type of change

### What is the solution?

We maintain a list of known arguments that can be accepted.
Anything beyond this list will be regarded as invalid. The handler throws an error when creating it with unknown arguments.

## Checklist:

### DIFF
--- a/tests/unit/ml_handlers/test_openai.py
+++ b/tests/unit/ml_handlers/test_openai.py
@@ -69,6 +69,21 @@ class TestOpenAI(BaseExecutorTest):
         with pytest.raises(Exception):
             self.wait_predictor("proj", "test_openai_nonexistant_model")
 
+    def test_unknown_arguments(self):
+        self.run_sql("create database proj")
+        with pytest.raises(Exception):
+            self.run_sql(
+                f"""
+                create model proj.test_openai_unknown_arguments
+                predict answer
+                using
+                    engine='openai',
+                    question_column='question',
+                    openai_api_key='{OPEN_AI_API_KEY}',
+                    model='unknown_argument';  --- this is a wrong argument name
+            """
+            )
+
     @patch("mindsdb.integrations.handlers.postgres_handler.Handler")
     def test_qa_no_context(self, mock_handler):
         # create project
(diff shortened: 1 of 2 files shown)


paxcema:
👍 Strongly recommend we continue developing what's proposed in the design doc, it is sorely needed and would be a massive improvement over the current situation. Perhaps ideal if we can design it in such a way that "customizable" bits in other commands (e.g. connection strings to DBs) can also leverage this new mechanism.

tests/unit/ml_handlers/test_openai.py
```diff
@@ -69,6 +68,21 @@ def test_invalid_openai_name_parameter(self):
...
+        with pytest.raises(Exception):
+            self.run_sql(
+                f"""
+                create model proj.test_openai_unknown_arguments
+                predict answer
+                using
+                    engine='openai',
+                    question_column='question',
+                    openai_api_key='{OPEN_AI_API_KEY}',
+                    model='unknown_argument';  --- this is a wrong argument name
```
paxcema, line 82: Minor nitpick: maybe rename the argument to something that has less probability of being a legit one in the future, e.g. `evidently_wrong_argument`

tests/unit/ml_handlers/test_openai.py
```diff
@@ -222,13 +236,13 @@ def test_bulk_normal_completion(self, mock_handler):
              openai_api_key='{OPEN_AI_API_KEY}';
         """  # noqa
         )
-        self.wait_predictor("proj", "test_openai_prompt_template")
+        self.wait_predictor("proj", "test_openai_bulk_normal_completion")
```
paxcema, line 239: Whoops, nice catch 😄

tests/unit/ml_handlers/test_openai.py
```diff
@@ -205,11 +219,11 @@ def test_prompt_template(self, mock_handler):
 
     @patch("mindsdb.integrations.handlers.postgres_handler.Handler")
     def test_bulk_normal_completion(self, mock_handler):
-        """ Tests normal completions (e.g. text-davinci-003) with bulk joins that are larger than the max batch_size """
+        """Tests normal completions (e.g. text-davinci-003) with bulk joins that are larger than the max batch_size"""
         # create project
         self.run_sql("create database proj")
-        handler_instance = OpenAIHandler()
-        N = 1 + handler_instance.max_batch_size  # get N larger than default batch size
+        max_batch_size = 10
```
paxcema, line 225: How would the test control that this number is actually bigger than what the handler is able to, ahm, handle? Say for instance OpenAI changes the maximum batch size from its current value (`N=20` I believe), this test should ideally test for `N=21`.

I guess we don't control this at the handler itself, but it seems like this change would introduce an additional parameter to control in order for the test to accomplish its objective.

paxcema (changes requested):
One nitpick and one requested change

paxcema (approved):

PR #7023, opened 2023-08-04
Bridge handler for langchain.embedding

yuhuishi-convect:
## Description

This PR builds a bridging class from `langchain.embeddings` module to MindsDB, bringing in 
embeddings models that are supported in `langchain` as supported models in MindsDB.

Design doc: https://docs.google.com/document/d/1ZoiTl3NM1MOYi0ZbVdIp1_zvoHece8dkvb8qU4YsvB8/edit
Usage: https://www.loom.com/share/a3f3365080f14ece80be823b5c9eb6ab?sid=493eb33e-296b-4a54-90a3-ad560f96d78e

## Type of change

### What is the solution?

Mapping from the base class of `langchain.embeddings.base.BaseEmbeddings` to MindsDB's `BaseMLEngine` class.

## Checklist:


yuhuishi-convect:
This new set of embedding models satisfy the interface as defined in https://docs.google.com/document/d/1_-Gc5yIeuXDeXPeBqhyY5y-yk6FJmNyCe1Qr7jtCVh0/edit#heading=h.5ut8jdqnay0s

Need to figure out a way to bring other models in other ml engines to this format.

mindsdb/integrations/handlers/langchain_embedding/langchain_embedding_handler.py
```diff
@@ -0,0 +1,201 @@
...
+        values = row.values.tolist()
+        document = "\n".join(
+            [f"{field}: {value}" for field, value in zip(fields, values)]
+        )
+        return document
+
+    def finetune(
+        self, df: Union[DataFrame, None] = None, args: Union[Dict, None] = None
+    ) -> None:
+        # re-save the model to the model storage
```
paxcema, line 183: It feels like calling this method on an embedding engine like this one should actually finetune the underlying embedding generator rather than just reindexing based on a new `df`, no?

If that were the case, then for this particular integration it may mean we have to deactivate this method entirely.

tests/unit/ml_handlers/test_langchain_embedding.py
```diff
@@ -0,0 +1,365 @@
...
+        ret = self.run_sql(
+            """
+            SELECT * FROM proj.test_dummy_embedding
+            WHERE content='hello'
+            """
+        )
+        assert "content" in ret.columns
+        assert "embeddings_output_column" in ret.columns
+        # the embeddings should be a list of 512 floats
+        assert len(ret["embeddings_output_column"][0]) == 512
```
paxcema, line 66: Curious, how is this test actually passing? Shouldn't the `FakeEmbedding` class not be found and subsequently failing to produce an embedding?

paxcema (approved):
Generally OK, I just have a couple minor questions for further discussion whenever time allows, but they are not blockers

tests/unit/ml_handlers/test_langchain_embedding.py
yuhuishi-convect: FakeEmbeddings is an available embedding model in langchain. It just spits out a random vector given a piece of text. https://python.langchain.com/docs/integrations/text_embedding/fake

mindsdb/integrations/handlers/langchain_embedding/langchain_embedding_handler.py
yuhuishi-convect: Yes, it does nothing. I will return a "NotImplementedError" instead

PR #7026, opened 2023-08-06
PyPI handler

lnxpy:
## Description

This handler adds some pypi-related statistics functionalities to the dashboard.

⚠️  This PR is _just_ created. This handler may not pass the tests at the current stage. It's in the development process.

**Fixes** #6867 

## Type of change

- [x] 📢 Breaking change (fix or feature that would cause existing functionality to not work as expected)

### What is the solution?

_Will be edited.._

## Checklist:


lnxpy:
🚫  Do NOT merge this PR at this stage.
⚠️  This PR is _just_ created. This handler may not pass the tests at the current stage. It's in the development process.

mindsdb/integrations/handlers/pypi_handler/pypi_handler.py
```diff
@@ -0,0 +1,30 @@
...
+        connection_data = kwargs.get("connection_data", {})
+        self.package = connection_data.get("package", None)
+
+        pypi_recent_data = PyPIRecentTable(self)
+        self._register_table("recent", pypi_recent_data)
+
+    def check_connection(self):
+        return StatusResponse(True)
+
+    def connect(self):
```
MinuraPunchihewa, line 26: Hey @lnxpy,
I thought I would reply here since you asked. Since you don't seem to need to instantiate an object or anything here, I believe you can just return the module itself in your `connect()` method.
```
def connect(self):
        return pypistats
```

You can now use this within your `Table` classes similar to the following example:
```
connection = self.handler.connect()
df = connection.recent('<some-package>', total=True, format="pandas")
```

mindsdb/integrations/handlers/pypi_handler/pypi_tables.py
```diff
@@ -0,0 +1,16 @@
...
+
+from mindsdb.integrations.libs.api_handler import APITable
+from mindsdb.utilities.log import get_log
+
+logger = get_log("integrations.pypi_handler")
+
+
+class PyPIRecentTable(APITable):
+    def select(self, query: ast.Select) -> pd.DataFrame:
+        df = pypistats.recent(self.handler.package, total=True, format="pandas")
```
MinuraPunchihewa, line 15: Maybe you were planning on doing this, but maybe the package to be viewed can be provided as part of the SQL query, i.e. the WHERE statement. For example:
```
SELECT *
FROM pypi_datasource.recent
WHERE package = 'pillow'
```

MinuraPunchihewa (changes requested):

mindsdb/integrations/handlers/pypi_handler/pypi_handler.py
lnxpy: Hi @MinuraPunchihewa,
Thanks.

MinuraPunchihewa:
Hey @lnxpy,
Is this done? I see that you've implemented all the tables.

lnxpy:
@MinuraPunchihewa Hi there!
It actually is but needs some cleaning as well. I'll work on it and clean it a little bit. I'll inform you once it's ready.

lnxpy:
@MinuraPunchihewa I think we're good to go. 🍻

MinuraPunchihewa (commented):
Hey @lnxpy,
This looks great! Personally, I feel that it would be better if we can allow users to query these statistics for all of the packages using a single instance of a DB. Since we are passing a package when executing a `CREATE DATABASE` statement, we will need multiple instances to query multiple packages.
On the other hand though, if we do it the way I've suggested, a `package` parameter will need to be made mandatory in the `WHERE` clause.
What are your thoughts on this? @ZoranPandovski what do you think?

lnxpy:
> Hey @lnxpy, This looks great! Personally, I feel that it would be better if we can allow users to query these statistics for all of the packages using a single instance of a DB..

There's a trade-off I guess. Lots of users would take an instance and use such a datasource for retrieving specific statistics of a package I guess. Like a company that needs to showcase the information and download rates of its Python package probably. That's kinda hard to imagine that someone needs to query for multiple packages. I mean, respectfully, that might not be as useful as the one that I implemented. Overall, that's fine. I'm pretty sure that there is a point in your thoughts that I can't see clearly. Let's see what @ZoranPandovski thinks about our speeches then we'll decide. :))

ZoranPandovski:
If this is ready we can now go with a package per connection. Then we can work on the imoprovment

mindsdb/integrations/handlers/pypi_handler/pypi_tables.py
lnxpy: @MinuraPunchihewa Then how's the implementation would be when there is no tables and `Handler.__init__()`?!

MinuraPunchihewa: Hey @lnxpy,
Apologies, I missed this. 

You will still need the handler and the tables.

The difference here would be that your `CREATE DATABASE` statements would be executed without passing any parameters. It will look something like this,
```
CREATE DATABASE pypi_datasource
WITH ENGINE = 'pypi'
```
Within the handler code, you will still have the `__init__()` method, but the difference is that we won't need to interact with the `connection_data` parameter in anyway.

I think this will be a good reference if you want to implement it this way,
https://github.com/mindsdb/mindsdb/blob/staging/mindsdb/integrations/handlers/mediawiki_handler/mediawiki_handler.py

lnxpy:
@MinuraPunchihewa @ZoranPandovski 
All the required changes are made. 👍🏻

ZoranPandovski (approved):

PR #7027, opened 2023-08-06
add support for poetry install

dusvyat:
## Description

This PR adds the ability to install mindsdb with poetry

**Fixes** #(issue)

## Type of change

### What is the solution?

## Checklist:


pyproject.toml
```diff
@@ -1,7 +1,88 @@
+[tool.poetry]
+name = "mindsdb"
+version = "0.0.0"  # leave as 0, it will auto update when pushed to artifact store based on latest git
```
paxcema, line 3: To understand a bit bettter - this means we should never update this line? Or does it mean that poetry will automatically change this line to match the latest git?

pyproject.toml
```diff
@@ -1,7 +1,88 @@
...
+pydateinfer = "0.3.0"
+pyarrow = ">=10.0.1,<10.1.0"
+dataprep-ml = "^0.0.15"
+python-magic = ">=0.4.27"
+openpyxl = ">=3.1.1"
+dill = "0.3.6"
+slack-sdk = "^3.21.3"
+mysql-connector-python = "^8.0.29"
+
+[tool.poetry.extras]
```
paxcema, line 81: Open question: any way we can read ML handler deps into this section as new groups?

For example, say OpenAI's handler requires `pandas, openai, tiktoken` and is defined under `requirements.txt` in its respective folder. Could we refer to that file here, somehow?

paxcema (approved):
Overall this looks good to me. Would be good to have the DevOps team's blessing (cc: @AndrewFarley, @hamishfagg, @tomhuds) just to ensure this won't interfere with the existing roadmap.

paxcema:
Ok just had a good chat with @AndrewFarley, we should look at a few things prior to merging this (if at all):

1. Can we simplify our current handling of `extra` dependencies with Poetry?
2. How does a working dockerfile look like with this approach?
3. What steps should we take to ensure users won't be negatively affected if we choose to switch?

pyproject.toml
dusvyat: yes correct, it would in place look for a git tag. This can be set up in CI/CD. I already have some example actions from past projects I've worked on

pyproject.toml
dusvyat: Yes I was thinking that could be a cool way to massively streamline current setup of having a requirements.txt per handler

dusvyat:
1. Can we simplify our current handling of `extra` dependencies with Poetry?

Not sure I follow, just to clarify how it currently is working in current implementation:

- To install all main requirements along with an extra e.g. grpc: 

poetry install -E  grpc 

- To install more than one extra:

poetry install -E  grpc  -E  telemetry

- To install test and dev along with main requirements:

poetry install --with test, dev

I find it simpler than pip where you would need to create a bunch of separate requirements.txt files


2. How does a working dockerfile look like with this approach?

I will propose something, will look to dig up an example from a past project

3. What steps should we take to ensure users won't be negatively affected if we choose to switch?

I propose a soft switch, its fairly easy to support both and I have seen repos with both options. If its working, over time I propose we faze the requirements.txt out.


Poetry also supports publishing, I have used in the past it works well for that too



> Ok just had a good chat with @AndrewFarley, we should look at a few things prior to merging this (if at all):
> 
> 1. Can we simplify our current handling of `extra` dependencies with Poetry?
> 2. How does a working dockerfile look like with this approach?
> 3. What steps should we take to ensure users won't be negatively affected if we choose to switch?

dusvyat:
Am I right in thinking that this the [release docker file](https://github.com/mindsdb/mindsdb/blob/staging/docker/release)? 

It seems rather than installing requirements from requirements.txt we are specifying them individually. What is the reason for this? 

> @paxcema @AndrewFarley
> 
> **1. Can we simplify our current handling of `extra` dependencies with Poetry?**
> 
> Not sure I follow, just to clarify how it currently is working in current implementation:
> 
> * To install all main requirements along with an extra e.g. grpc:
> 
> `poetry install -E grpc `
> 
> * To install more than one extra:
> 
> `poetry install -E grpc -E telemetry`
> 
> * To install test and dev along with main requirements:
> 
> `poetry install --with test, dev`
> 
> I find it simpler than pip where you would need to create a bunch of separate requirements.txt files
> 
> **2. How does a working dockerfile look like with this approach?**
> 
> I will propose something, will look to dig up an example from a past project
> 
> **3. What steps should we take to ensure users won't be negatively affected if we choose to switch?**
> 
> I propose a soft switch, its fairly easy to support both and I have seen repos with both options. If its working, over time I propose we faze the requirements.txt out.
> 
> Poetry also supports publishing, I have used in the past it works well for that too
> 
> > Ok just had a good chat with @AndrewFarley, we should look at a few things prior to merging this (if at all):
> > 
> > 1. Can we simplify our current handling of `extra` dependencies with Poetry?
> > 2. How does a working dockerfile look like with this approach?
> > 3. What steps should we take to ensure users won't be negatively affected if we choose to switch?

dusvyat:
@paxcema any issue with me merging this?

paxcema:
Deferred to @tomhuds on next steps.

dusvyat:
@tomhuds Any news? Are we good to merge this?

tomhuds:
Nope no news - not a priority for this sprint

ANIKKARAINE (approved):

ZoranPandovski:
@dusvyat Should we close this?

dusvyat:
@ZoranPandovski yes I think we can for now

PR #7028, opened 2023-08-06
Handler for SentenceTransformers

dusvyat:
## Description

Given a table, get embeddings and store as a separate table

**Fixes** #(issue)

## Type of change

### What is the solution?

## Checklist:


mindsdb/integrations/handlers/sentence_transformer_handler/helpers.py
```diff
@@ -0,0 +1,10 @@
+from integrations.handlers.sentence_transformer_handler.settings import (
+    load_embeddings_model,
```
paxcema, line 2: Nit: maybe `load_embeddings_model` should be moved here, instead?

mindsdb/integrations/handlers/sentence_transformer_handler/requirements.txt
```diff
@@ -0,0 +1,2 @@
+sentence-transformers
+torch
```
paxcema, line 2: `pydantic` is missing

paxcema: Same with `pandas`, but OK as it's required by MindsDB core.

mindsdb/integrations/handlers/sentence_transformer_handler/settings.py
```diff
@@ -0,0 +1,79 @@
...
+
+class DfLoader:
+    def __init__(self, data_frame: pd.DataFrame, page_content_column: str):
+        self._data_frame = data_frame
+        self._page_content_column = page_content_column
+
+    def load(self) -> List[Document]:
+        """takes a Dataframe column and loads it as a list of documents"""
+        documents = []
+        for n_row, frame in self._data_frame[self._page_content_column].iteritems():
```
paxcema, line 41: Potentially slow for large frames, but fine for v1.0. Just highlighting for future reference.

mindsdb/integrations/handlers/sentence_transformer_handler/sentence_transformer_handler.py
```diff
@@ -0,0 +1,71 @@
...
+            columns = df.columns
+
+        documents = df_to_documents(df=df, page_content_columns=columns)
+
+        model = dill.loads(self.model_storage.file_get("model"))
+
+        embeddings = []
+        metadata = []
+
+        for _, document in enumerate(documents):
```
paxcema, line 49: Had a meeting recently with Yoshi and Tom, we've agreed that the output of an embedding model such as the ones created by this handler should always output a single embedding, i.e. a dataframe with a single column.

The aggregation mechanism is up to the underlying integration, as long as it is consistent it will be fine.

paxcema (changes requested):
Looking good! I may suggest additional changes once the base `EmbeddingModel` class is finished. Will be fairly minor though, I'd be OK with merging as-is.

mindsdb/integrations/handlers/sentence_transformer_handler/requirements.txt
dusvyat: yes I left out pandas deliberately, will add pydantic tho!

mindsdb/integrations/handlers/sentence_transformer_handler/sentence_transformer_handler.py
dusvyat: Ok good to know, apologies I wasn't aware of this

mindsdb/integrations/handlers/sentence_transformer_handler/helpers.py
dusvyat: I like to follow pattern where any loading and config happens in `settings.py`, ofc this is not set in stone so happy to move it if you prefer it that way

mindsdb/integrations/handlers/sentence_transformer_handler/settings.py
dusvyat: tbh only reason for this was I thought we wanted to log metadata, if we don't then can remove and make faster

mindsdb/integrations/handlers/sentence_transformer_handler/helpers.py
paxcema: Ok fair, let's do that, no strong preference on my end 👌

mindsdb/integrations/handlers/sentence_transformer_handler/settings.py
paxcema: Ah I see, let's keep like this for now then

mindsdb/integrations/handlers/sentence_transformer_handler/sentence_transformer_handler.py
paxcema: No worries, we actually dropped the ball on communicating this properly I'd say.

re: metadata yes, I think so!

PR #7174, opened 2023-08-24
Don't Cut Off Langchain Error Message in Langchain Handler

tmichaeldb:
## Description

Currently we only display the first 50 characters of a Langchain error message. This change returns the whole error in a completion. Is there any reason we want to cut it off short? Even if it's super long it shouldn't be an issue.

**Fixes** #6878 

## Type of change


paxcema (approved):
Agreed, no real reason we're doing this.

PR #7177, opened 2023-08-25
handle case of group_by column not present in the dataframe.

pedrofluxa:
## Description

Based on the discussion on the Slack channel, I believe the problem comes from the way the group_by is being handled when no group_by argument is specified.

**Fixes** #(issue)

## Type of change

### What is the solution?

https://github.com/mindsdb/lightwood/issues/1180#issuecomment-1693278552

## Checklist:

- [ ] I have updated the documentation, or created issues to update them. -> no need
- [ ] I fixed|updated|added unit tests and integration tests for each feature (if applicable). -> no need
- [ ] I have shared a short loom video or screenshots demonstrating any new functionality. -> no need

### DIFF
--- a/mindsdb/integrations/utilities/time_series_utils.py
+++ b/mindsdb/integrations/utilities/time_series_utils.py
@@ -121,6 +121,9 @@ def get_hierarchy_from_df(df, model_args):
     nixtla_df = df.rename({model_args["order_by"]: "ds", model_args["target"]: "y"}, axis=1)
     nixtla_df["ds"] = pd.to_datetime(nixtla_df["ds"])
     for col in model_args["group_by"]:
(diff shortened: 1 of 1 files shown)


paxcema (changes requested):
EDIT: The command below is incorrect, disregard.

* * * 
Tried manually and it seems there is another error down the line. 

Here's the command for trying it out:


```sql
CREATE MODEL 
  mindsdb.house_sales_model25sfh
FROM example_db
  (SELECT * FROM demo_data.house_sales WHERE type = 'house' AND bedrooms = 3)
PREDICT ma
ORDER BY saledate
WINDOW 8
HORIZON 4
USING 
ENGINE = 'statsforecast',
hierarchy = True;
```

PR #7205, opened 2023-08-30
[ENH] Typed statement for anomaly detection models

paxcema:
## Description

_Note_: requires a new `mindsdb_sql` release ([PR#294](https://github.com/mindsdb/mindsdb_sql/pull/294), in particular, part of [this](https://github.com/mindsdb/mindsdb_sql/pull/293) release).

This PR adds support for `CREATE ANOMALY DETECTION MODEL` syntax and models. The only real difference in the backend is that models are now task-specific, and a specific type of SQL statement can be used to detect between these.

For now, logic in the executioner remains unchanged, but as we further develop the UX of each use case, these may diverge as we see fit.

The PR additionally changes the handler for TimeGPT so that it can support normal forecasting _and_ anomaly detection. This is a first working example (although not necessarily final) of how ML handlers should deal with multi-task support, moving forward. 

## Type of change

### What is the solution?

## Checklist:


mindsdb/api/mysql/mysql_proxy/executor/executor_commands.py
```diff
@@ -1147,13 +1154,9 @@ def answer_drop_view(self, statement):
     @mark_process('learn')
     def answer_create_predictor(self, statement):
         integration_name = self.session.database
-        if len(statement.name.parts) > 1:
```
ea-rus, line 1150: this piece of code allows to create model in other project. for example active project is mindnsdb but we can call 'create mode proj1.mindsdb'

mindsdb/api/mysql/mysql_proxy/executor/executor_commands.py
```diff
@@ -550,7 +554,10 @@ def execute_command(self, statement):
             db_name = statement.value.parts[-1]
             self.change_default_db(db_name)
             return ExecuteAnswer(ANSWER_TYPE.OK)
-        elif type(statement) == CreatePredictor:
+        elif type(statement) in (
```
ea-rus, line 557: maybe isinstance(..., CreatePredictorBase) ?
and put 'elif type(statement) == FinetunePredictor' upper
it is not important, just a proposal

mindsdb/api/mysql/mysql_proxy/executor/executor_commands.py
paxcema: Ahhh I see, will restore!

paxcema: Hmm, looks like it is not being used, according to the linter 🤔

mindsdb/api/mysql/mysql_proxy/executor/executor_commands.py
paxcema: Will keep as is for now, otherwise there is a small risk that the `if` order may be switched around mistakenly and so breaking either finetuning or creation

paxcema:
@ea-rus feedback has been addressed and the new version of mindsdb_sql has been released, can you take another look and hopefully approve if there's nothing else?

ea-rus (approved):

.github/workflows/mindsdb.yml
```diff
@@ -68,10 +68,9 @@ jobs:
     - name: Install staging mdb libraries if it's not mindsdb/stable
       if: ${{ github.ref != 'refs/heads/stable' }}
       run: |
-          pip install git+https://github.com/mindsdb/mindsdb_sql.git@staging --upgrade --no-cache-dir
           pip install git+https://github.com/mindsdb/lightwood.git@staging --upgrade --no-cache-dir
           pip uninstall pycodestyle -y && pip install pycodestyle==2.9.1  # temp fix while exist conflict in last version with flake8
-          pip install gluonts==0.11.4  # temp fix until next lightwood release (>23.1.2.1)
+          pip install git+https://github.com/mindsdb/mindsdb_sql.git@staging --upgrade --no-cache-dir  # install post lightwood so it does not get overwritten
```
ea-rus, line 73: I think moving mindsdb_sql here won't help
we need to bump version in staging branch of mindsdb_sql

PR #7217, opened 2023-08-31
Allow delete active model verison

### LINKED ISSUE #7121: [Bug]: Cannot DELETE FROM mindsdb.models_versions table

**Here is the model that has two versions:**

image

**Deleting one (inactive) version results in success:**

image

**But when querying the `models_versions` table, the deleted version is still there:**

image

### Video or screenshots

_No response_

### Expected behavior

_No response_

_No response_

### Anything else?

This bug is on cloud - alpha is ok.

StpMax:
## Description

Allow to delete active model version, if all models version deleting in operation:
```
delete from models_version where name = 'model';
```

**Fixes** 
close #7121

## Type of change

### What is the solution?

## Checklist:

### DIFF
--- a/mindsdb/interfaces/model/model_controller.py
+++ b/mindsdb/interfaces/model/model_controller.py
@@ -5,6 +5,7 @@ import base64
 import datetime as dt
 from copy import deepcopy
 from multiprocessing.pool import ThreadPool
+from collections import defaultdict
 
 import pandas as pd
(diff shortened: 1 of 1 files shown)


ea-rus:
I don't thing we should allow to delete all versions:
`delete from models_versions where name = 'model';`
 
It is the same as deleling of model, let they use delete from models
`delete from models where name = 'model';`

PR #7228, opened 2023-09-01
fixed langchain parameter name

martyna-mindsdb:
## Description

fixed langchain parameter name

- using `prompt_template` locally works fine - but the model's answer is null

image

- using `prompt` on cloud works fine (after deploying this fix, use `prompt_template` on cloud as well)

image

**Fixes** #(issue)

## Type of change

### What is the solution?

## Checklist:


paxcema (approved):

paxcema:
The code diff itself looks good to me. I am looking into the behavior above, it may be we're missing something here.

paxcema:
Ok, all good. The prompt template is missing the actual input column. If we change it to:

`prompt_template = 'Answer the users input in a helpful way: {{question}}'` then this PR works OK.

There was an additional bug where the `prompt_template` variable was not overridden if provided at prediction time, so I've added a fix for that.

martyna-mindsdb:
Thanks @paxcema 
So this RP can be merged now.

PR #7267, opened 2023-09-06
MySQL: remove show tables

federico-razzoli:
## Description

Previously `get_tables()` run `SHOW TABLE` against MySQL. With this change, it now runs a `SELECT` against `information_schema.TABLES`. It returns the same columns as the PostgreSQL integration and it has an `ORDER BY`.

This change guarantees that `SHOW TABLES` (when run in a MySQL integration) shows the tables in an alphabetical order.

Including more columns means that we also allows MindsDB to distinguish tables from views. Not sure if this information is currently used.

**Fixes** #7232

## Type of change

## Checklist:


federico-razzoli:
Please wait. I'll comment again when I think that the PR is ready.

federico-razzoli:
Should be ok. Please process the pull request.

PR #7317, opened 2023-09-12
Fix Update Functionality in Chatbot Controller

tmichaeldb:
## Description

Related to #7306.

This PR fixes the update functionality, since `get_chatbot` returns a dict. So now we get the actual DB record from that.

Also, the REST API needed to be updated because we get a `ValueError` instead of `NoResultFound` when a project doesn't exist now.

Finally, all the tests are updated with the correct behavior.

### DIFF
--- a/mindsdb/interfaces/chatbot/chatbot_controller.py
+++ b/mindsdb/interfaces/chatbot/chatbot_controller.py
@@ -247,7 +247,7 @@ class ChatBotController:
 
         if params is not None:
             # Merge params on update
-            existing_params = {} if not existing_chatbot_rec.params else existing_chatbot_rec.params
+            existing_params = existing_chatbot_rec.params or {}
             params.update(existing_params)
             existing_chatbot_rec.params = params
         db.session.commit()


mindsdb/interfaces/chatbot/chatbot_controller.py
```diff
@@ -216,23 +216,25 @@ def update_chatbot(
         if existing_chatbot is None:
             raise Exception(f'Chat bot not found: {chatbot_name}')
 
+        existing_chatbot_rec = db.ChatBots.query.get(existing_chatbot['id'])
```
yuhuishi-convect, line 219: NIT: what type of exception does it throw if id is invalid

mindsdb/interfaces/chatbot/chatbot_controller.py
```diff
@@ -245,12 +247,12 @@ def update_chatbot(
 
         if params is not None:
             # Merge params on update
-            existing_params = {} if not existing_chatbot.params else existing_chatbot.params
+            existing_params = {} if not existing_chatbot_rec.params else existing_chatbot_rec.params
```
yuhuishi-convect, line 250:
```suggestion
            existing_params = existing_chatbot_rec.params or {}
```
NIT

yuhuishi-convect (approved):
LGTM

PR #7333, opened 2023-09-14
Allow providing an empty API list arg

hamishfagg:
This change is required for the jobs microservice [here](https://github.com/mindsdb/INTERNAL-mindsdb-build-deploy-to-kubernetes/pull/13)

This allows supplying an empty api list with `--api=`, which will then only run the jobs and tasks services (depending on the user's config).


mindsdb-devops (approved):
LGTM

PR #7334, opened 2023-09-14
[fix] fix infer freq statsforecast

paxcema:
## Description

**Fixes** #7193.

Underlying issue: taking `isinstance(x, dt.date)` is also true for `dt.time` objects, so timestamp resolution was being truncated. Additionally, the current infer freq method was failing in this case (milisecond level).

## Type of change

## Verification Process

 - [x]   Test Location: local
 - [x]   Verification Steps: run StatsForecast learning hub example, and provided dataset in issue. *Note:* due to CI acting up, I could not verify these tests were/are passing.

## Checklist:


mindsdb/api/mysql/mysql_proxy/classes/sql_query.py
```diff
@@ -1556,7 +1556,7 @@ def get_date_format(samples):
                     if isinstance(arg, Constant) and isinstance(arg.value, str):
                         arg.value = fnc(arg.value)
 
-        if self.model_types.get(order_col) in ('date', 'datetime'):
+        if self.model_types.get(order_col) in ('date', 'datetime') or isinstance(predictor_data[0][order_col], pd.Timestamp):  # noqa
```
paxcema, line 1559: imho we should simplify this time filter urgently, but realistically that's another issue (part of the TS refactor).

As of now, this is a good compromise to avoid breaking anything.

mindsdb/integrations/handlers/statsforecast_handler/statsforecast_handler.py
```diff
@@ -159,10 +159,6 @@ def predict(self, df, args={}):
             results_df = forecast_df[forecast_df.index.isin(groups_to_keep)]
 
         result = get_results_from_nixtla_df(results_df, model_args)
-
-        ts_col = model_args["order_by"]
-        if len(result) > 0 and isinstance(result.iloc[0][ts_col], dt.date):
-            result[ts_col] = result[ts_col].dt.date
```
paxcema, line 165: Obs: Not good. Time component was being truncated.

mindsdb/integrations/utilities/time_series_utils.py
```diff
@@ -65,10 +66,10 @@ def get_results_from_nixtla_df(nixtla_df, model_args):
 
 def infer_frequency(df, time_column, default=DEFAULT_FREQUENCY):
     try:  # infer frequency from time column
-        date_series = pd.to_datetime(df[time_column]).unique()
-        if isinstance(date_series, np.ndarray):
-            date_series.sort()
-        inferred_freq = pd.infer_freq(date_series)
+        date_series = pd.to_datetime(df.sort_values(by=time_column)[time_column]).unique()
```
paxcema, line 69: Obs: this is not robust. E.g. given a monthly dataset, it will yield a specific time distance (most common), `nD` instead of the correct `1M`. We should improve this, @pedrofluxa any ideas? If not, we can try later and unblock the microsecond use case for now with what we have.

paxcema:
More test-driven and double check the trade off.

paxcema:
Alternative: move mechanism out of Lightwood into dataprep_ml, then use this for all engines.

paxcema:
Draft review by @pedrofluxa:
1. interpolate by [s]
2. Resample to most popular freq -> changes data but that's fine. (assumes no gaps)

Add to that:
1. Heuristic for

pedrofluxa (changes requested):
Changes will not break current capabilities, but I strongly suggest to include a re-sampling step in the `transform_to_nixtla_df()` routine to handle non-periodic (or pseudo-periodic) inferred frequencies.

mindsdb/integrations/utilities/time_series_utils.py
paxcema: Solution: first try `pd.infer_freq`, and if it does not work follow up with this approach.

paxcema:
Okay, feedback has been addressed. @pedrofluxa please have a look, and if nothing is amiss approve + merge :)

mindsdb/integrations/utilities/time_series_utils.py
```diff
@@ -22,6 +23,20 @@ def transform_to_nixtla_df(df, settings_dict, exog_vars=[]):
...
     """
     nixtla_df = df.copy()
+
+    # Resample every group
+    freq = settings_dict['frequency']
+    resampled_df = pd.DataFrame(columns=nixtla_df.columns)
+    if settings_dict["group_by"] and settings_dict["group_by"] != ['__group_by']:
+        for group, groupdf in nixtla_df.groupby(by=settings_dict["group_by"]):
+            groupdf.index = pd.to_datetime(groupdf.pop(settings_dict["order_by"]))
+            resampled_groupdf = pd.DataFrame(groupdf[settings_dict['target']].resample(freq).mean())
```
pedrofluxa, line 33: This will re-sample using a moving average, which might break some of the statistical properties of the data.

Not the best-of-the-best re-sampling but should work just fine for most cases, I approve it.

pedrofluxa (approved):
Looks good, I approve the changes.

PR #7366, opened 2023-09-19
Unsafe version of BYOM

StpMax:
## Description

 - added route /readiness to get instance 'readiness' status (at the moment just ceck if instance do any task)
 - byom handler can work in two models: safe (old way via venv) and unsafe (run task in current process). How BYOM will 
 -  work defined by MINDSDB_BYOM_TYPE variable
 - added handler cache for predict/learn process
 - lock files moved to separate folder
 - locks now can be not only exclusive, but also shared

**Fixes** #(issue)

## Type of change

## Verification Process

## Additional Media:

## Checklist:


mindsdb/interfaces/storage/model_fs.py
```diff
@@ -69,12 +69,12 @@ def file_get(self, name):
     def file_set(self, name, content):
         self.fileStorage.file_set(name, content)
 
-    def folder_get(self, name, update=True):
```
ea-rus, line 72: update=False is used for big HF models. It allows to not download if it is already downloaded to instance

mindsdb/interfaces/storage/fs.py
```diff
@@ -393,53 +421,44 @@ def __init__(self, resource_group: str, resource_id: int,
...
-                # not pull from source if object is exist
-                if os.path.exists(self.resource_group_path / self.folder_name / path):
-                    return
-            try:
-                # TODO not sync if not changed?
-                self.fs_store.get(os.path.join(self.folder_name, path), str(self.resource_group_path))
-            except Exception:
-                pass
+    def pull_path(self, path):
+        # TODO implement pull per element
```
ea-rus, line 453: will it be implemented before merge? because path is used now

mindsdb/interfaces/storage/model_fs.py
```diff
@@ -167,28 +179,12 @@ def file_del(self, name):
 
     # folder
 
-    def folder_get(self, name, update=True, not_empty=False):
-        '''
```
ea-rus, line 171: lets keep doc-string, it is path of our api for handlers

mindsdb/interfaces/storage/model_fs.py
StpMax: - storage already has `sync` parameter to control copy behavior
 - local and s3 storages checks changes in files before upload/download even if sync is True
 - `folder_get` does not work for s3 storage

mindsdb/interfaces/storage/fs.py
StpMax: no, this method should not exists, because we may copy only whole base folder if use s3 as storage

ea-rus (approved):

PR #7419, opened 2023-09-26
Fix timegpt queries

ea-rus:
## Description

Fixes:
1. Don't infer types globally. Otherwise view shows different result

```sql
CREATE VIEW v (
   SELECT strftime(month, '%Y-%m-%d') 
    FROM mysql_demo_db (select * from historical_expenditures)
);

-- this query shows query without  strftime
select  * from v;
```
Video:
https://www.loom.com/share/2a403f650d5a42b199b0da309116ef3f

2. Return date column from timeGPT as date type. It is needed for the last query from the script worked:
```sql
--------------- data

CREATE DATABASE mysql_demo_db
WITH ENGINE = "mysql",
PARAMETERS = {
    "user": "user",
    "password": "MindsDBUser123!",
    "host": "db-demo-data.cwoyhfn6bzs0.us-east-1.rds.amazonaws.com",
    "port": "3306",
    "database": "public"
    };

CREATE VIEW training_data (
   SELECT * FROM mysql_demo_db.historical_expenditures
   WHERE month NOT IN ('2017-07-01', '2017-08-01', '2017-09-01')
);

--------------- models

CREATE MODEL statsforecast_model
FROM mindsdb
    (SELECT * FROM training_data)
PREDICT expenditure
ORDER BY month
GROUP BY category
WINDOW 36
HORIZON 3
USING ENGINE = 'statsforecast';

CREATE MODEL timegpt_model
FROM mindsdb
    (SELECT * FROM training_data)
PREDICT expenditure
ORDER BY month
GROUP BY category
WINDOW 36
HORIZON 3
USING ENGINE = 'timegpt';

SELECT strftime(realvalues.month, '%Y-%m-%d') AS month, realvalues.category, realvalues.expenditure AS true_value,
       statsforecast.expenditure AS statsforecast_value,
       abs(cast(realvalues.expenditure as double)-cast(statsforecast.expenditure as double))/cast(realvalues.expenditure as double)*100 AS percentage,
       timegptforecast.expenditure AS timegpt_value
FROM (SELECT *
        FROM mysql_demo_db.historical_expenditures
        WHERE category = 'industry'
        AND (month = '2017-07-01'
            OR month = '2017-08-01'
            OR month = '2017-09-01')) AS realvalues
JOIN (SELECT m.month, m.category, d.expenditure
        FROM training_data AS d
        JOIN statsforecast_model AS m
        WHERE d.category = 'industry'
        AND d.month > LATEST
        LIMIT 3) AS statsforecast
ON realvalues.month = statsforecast.month
JOIN (SELECT m.month, m.category, d.expenditure
        FROM training_data AS d
        JOIN timegpt_model AS m
        WHERE d.category = 'industry'
        AND d.month > LATEST
        LIMIT 3) AS timegptforecast
ON realvalues.month = timegptforecast.month;
```

**Fixes** #(issue)

## Type of change

## Verification Process

## Additional Media:

## Checklist:


mindsdb/integrations/utilities/infer_types.py
```diff
@@ -0,0 +1,54 @@
...
+
+    # Check if the inferred type is the same as the original type
+    if inferred_column.dtype != sample_data.dtype:
+        # If not, apply the inferred type to the entire column
+        df[column] = infer_column_type(df[column], type=type)
+
+    return df
+
+
+def infer_and_convert_types(df):
```
paxcema, line 52: This method is not being used anymore, perhaps drop it?

mindsdb/integrations/utilities/infer_types.py
```diff
@@ -0,0 +1,54 @@
+import pandas as pd
+
+
+def infer_column_type(column, type=None):
```
paxcema, line 4: I feel like we should drop these methods, and use `type_infer` instead. If it does not work correctly, then frame it as "what can/should we add from this method into that package", as anything having to do with type inference for ML (like this) should go into that package.

paxcema (changes requested):
I would approve as-is, but perhaps this is an opportunity to streamline this bit of code even further. Let me know your thoughts @ea-rus, if you think that's best tackled as a separate issue I will approve this.

mindsdb/integrations/utilities/infer_types.py
ea-rus: What functionality is required now in timegpt handler:
- convert one column in datetime if it possible.

I could replace it with using of type_infer module like this:
```python

sample_data = df[column].sample(min(100, len(df)))
if type_infer.type_check_date(sample_data):
    df[column] = pd.to_datetime(df[column])
```

Would it be better? Or there could be another more preferable solution?

mindsdb/integrations/utilities/infer_types.py
ea-rus: dropped full file

mindsdb/integrations/utilities/infer_types.py
ea-rus: done

paxcema: I think there's a better way (also, the learning hub example does not work as-is). 

Here is a set of changes that work, we may want to merge that [here](https://github.com/mindsdb/mindsdb/compare/fix-timegpt...fix_timegpt_7419_v2).

paxcema (changes requested):
Learning hub solution does not work as-is. If we slightly change the function call, it works. See [here](https://github.com/mindsdb/mindsdb/compare/fix-timegpt...fix_timegpt_7419_v2).

paxcema (approved):

mindsdb/integrations/handlers/timegpt_handler/timegpt_handler.py
```diff
@@ -104,7 +108,12 @@ def predict(self, df, args={}):
         else:
             raise Exception(f'Unsupported prediction mode: {model_args["mode"]}')
 
+        # infer date
+        ds_col = model_args["order_by"]
```
paxcema, line 112: Try this instead and we can do without the extra file and methods:

```python
# cast back to datetime if applicable
sample_data = forecast_df['ds'].sample(min(100, len(forecast_df)))
dtypes = count_data_types_in_column(sample_data)
if dtype.date in dtypes or dtype.datetime in dtypes:
    forecast_df['ds'] = pd.to_datetime(forecast_df['ds'])
````

**Review by paxcema — CHANGES_REQUESTED** [review]

File: mindsdb/integrations/utilities/infer_types.py [hunk 4]
```diff
@@ -0,0 +1,54 @@
+import pandas as pd
```
paxcema, line 1: remove file, not in use

mindsdb/integrations/handlers/timegpt_handler/timegpt_handler.py
```diff
@@ -88,6 +91,7 @@ def predict(self, df, args={}):
 
             # TODO: enable this post-refactor
             # X_df=None,  # exogenous variables
+            # add_history=False,  # insample
```
paxcema, line 94: nitpick: remove as it is added in line 48 already

paxcema (changes requested):
Two minor changes on code formatting. Apart from that, this looks ready to go.

mindsdb/integrations/utilities/infer_types.py
ea-rus: done

mindsdb/integrations/handlers/timegpt_handler/timegpt_handler.py
ea-rus: removed 'add_history=False' line

PR #7463, opened 2023-09-30
Pytorch_tabular Integration

pranavvp16:
model_storage for trained model is remaining to add . Is there any documentation regarding how to save a pytorch model using the self.model_storage #6387


paxcema:
> Is there any documentation regarding how to save a pytorch model using the self.model_storage

It is quite similar to what is mentioned in step 4 of the [step-by-step guide here](https://docs.mindsdb.com/contribute/ml-handlers): `engine_storage` saves anything that is common to all models created with this engine for the user, and `model_storage` can be used to save anything that is model-specific.

In terms of what methods you can use, check its source [here](https://github.com/mindsdb/mindsdb/blob/f7516a6252f26f4e437189195e1f1042486ac221/mindsdb/interfaces/storage/model_fs.py#L124). I don't think we have API docs for it yet.

Finally, I recommend checking other engines for example usage.

pranavvp16:
pytorch tabular is working now @paxcema @ZoranPandovski 
image
image

mindsdb/integrations/handlers/pytorch_tabular_handler/pytorch_tabular_handler.py
```diff
@@ -23,21 +25,26 @@ def create_validation(self,args: Optional[dict] = None) -> None:
...
+        if "task" in args:
             if args["task"] not in task_supported:
                 raise Exception(f"Please specify task parameter supported : {task_supported}")
         if "initialization" not in args or args["initialization"] not in initialization_supported:
             raise Exception(f"Initialization scheme choices are : {initialization_supported}")
-        #if not target:
-            #raise Exception("Please provide the target column")
+        if "layers" in args:
+            if not re.match(pattern,args["layers"]):
+                raise Exception(f"Please specify layers in format : '128-64-32'")
```
paxcema, line 36: Out of ignorance (no experience with pytorch tabular yet), why is a formatted string preferred to a list of integers?

Our SQL parser does support something like this:

```sql
CREATE MODEL
...
USING layers = ['128', '64', '32']  -- integers also work: [128, 64, 32] 
```

mindsdb/integrations/handlers/pytorch_tabular_handler/README.md
```diff
@@ -0,0 +1,92 @@
...
+AND population=500
+AND households=200
+AND median income=8.5
+AND ocean_proximity= 'NEAR BAY';
+```
+
+Get batch predictions from the handler
+```
+SELECT m.*
+FROM files. California Housing as t
```
paxcema, line 90: These spaces will likely raise a parsing error as well

paxcema: I noticed it is fixed on the screenshots you've provided, which is great. Let's just make sure this example is also updated.

paxcema (changes requested):

pranavvp16:
@paxcema any updates on this ??

tests/unit/executor_test_base.py
```diff
@@ -120,6 +118,9 @@ def clear_db(self, db):
             name="langchain_embedding", data={}, engine="langchain_embedding"
         )
         db.session.add(r)
+        r = db.Integration(name="rag", data={}, engine="rag")
```
paxcema, line 121: This change seems unnecessary, please remove this.

paxcema (changes requested):
Nice, there are two pending items before we can merge this:
1. A working set of unit tests for the handler
2. Fix all linting errors that `flake8` shows. You can see them in the Github actions for the last commit.

pranavvp16:
@ZoranPandovski I have pull the staging and all the tests pass now

ZoranPandovski (changes requested):
Thanks, let's solve the conflicts before we merge this

paxcema:
Hey @pranavvp16 there is still one file with a merge conflict, can you check please?

pranavvp16:
@paxcema I have fixed the merge conflicts

paxcema:
Nice! Seems like there is still a problem in the requirements check though 🔍, related to `setuptools`. Can you please fix it?

PR #7711, opened 2023-10-11
[ENH] Integration to Anyscale endpoints

paxcema:
## Description

Fixes #issue_number

## Type of change

## Verification Process

## Additional Media:

## Checklist:


mindsdb/integrations/handlers/anyscale_endpoints_handler/anyscale_endpoints_handler.py
```diff
@@ -0,0 +1,175 @@
...
+    def predict(self, df: pd.DataFrame, args: Optional[Dict] = None) -> pd.DataFrame:
+        with self._anyscale_base_api():
+            return super().predict(df, args)
+
+    def finetune(self, df: Optional[pd.DataFrame] = None, args: Optional[Dict] = None) -> None:
+        with self._anyscale_base_api():
+            super().finetune(df, args)
+            # rewrite chat_completion_models to include the newly fine-tuned model
+            args = self.model_storage.json_get('args')
+            self.chat_completion_models = list(self.chat_completion_models) + [args['model_name']]
```
paxcema, line 64: @ea-rus just for confirmation (because this does work empirically for `n=1` users): the stateful class instances that our executor runs is at each user's level, correct? I.e. modifying this list with model `X` won't leak to other users who may have finetuned model `Y`?

dusvyat (approved):
other than conflicts, lgtm!

mindsdb/integrations/handlers/anyscale_endpoints_handler/anyscale_endpoints_handler.py
ea-rus: I don't know context much. Can the same model name be used by other user? If yes, it should have some unique part in name, for example model_name = <predictor id> + model name
 
Also question: is chat_completion_models variable persisted anywhere? if not it will be cleared after handler reinitialized

paxcema: > Can the same model name be used by other user?
Because it is a unique model fine-tuned by this user, we want to avoid that situation. There is a unique part in the name, the handler leverages logic for this in the OpenAI handler.

> It will be cleared after handler reinitialized
Nice, this is what my original question aimed at. Thanks for clarifying. I will have to modify this slightly to store all finetuned model names per user.

PR #7715, opened 2023-10-11
Windows Compatibiliy & rss+xml handling

RobStanfield:
Replace `fcntl `with `filelock` for Windows compatibility.
Allow handling of application/rss+xml in webhandler


StpMax (changes requested):
Hi @RobStanfield 
Changes related to file lock will not be accepted.
 - `filelock` does not have api for shared locks
 - at unix `filelock` use `fcntl`, but instead of `lockf` it use `flock`. It has issues on some file systems.

ZoranPandovski:
@RobStanfield Please check other help-wanted issues :raised_hands:

PR #7746, opened 2023-10-13
File handler unit tests

hamishfagg:
## Description

Adds unit tests to the `file_handler`. This is intended to provide an example of unit tests for our handlers in general, for our devs/community to aim for.

This PR also adds an example workflow for triggering tests based on changes made to the file handler.

(currently a WIP, there are a few items to complete)

## Type of change

- [x] 📄 This change requires a documentation update

## Checklist:


dusvyat:
@hamishfagg please can we make sure we don't remove any `file_handler` unit tests in `staging`? for e.g. `test_handle_txt `and `test_handle_pdf`

hamishfagg:
@dusvyat definitely. We might move them into the handler folder along with these new tests before merge

mindsdb/integrations/handlers/file_handler/tests/test_file_handler.py
```diff
@@ -0,0 +1,352 @@
...
+        def mock_delete(self, name):
+            raise Exception("File delete error")
+
+        monkeypatch.setattr(MockFileController, "delete_file", mock_delete)
+        file_handler = FileHandler(file_controller=MockFileController())
+        response = file_handler.query(DropTables([Identifier(parts=["one"])]))
+
+        assert response.type == RESPONSE_TYPE.ERROR
+
+    def test_query_select(self, csv_file):
```
hamishfagg, line 174: If i could get some help from @StpMax or @dusvyat or someone else:
I can't get this test to run, I always get `Exception: File 'test_data.csv' does not exists`

dusvyat: looks like `test_data.csv` is missing, you can add a small csv into [here](https://github.com/mindsdb/mindsdb/tree/e4eccd25aa404539c168fb7e8c0ac1409f176a52/tests/handler_tests/data)

hamishfagg: @dusvyat the `csv_file` fixture is creating and saving `test_data.csv` to a temp dir, and the test sets the `content` path to be that temp dir. So the file does exist (I added some debug statements to make sure earlier), but I think maybe the 'files' integration needs to be made aware of it?

.github/workflows/testing.yml
```diff
@@ -0,0 +1,33 @@
...
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+      - uses: actions/setup-python@v4.7.0
+        with:
+          python-version: 3.9
+      - name: Install python deps
+        run: |
+          pip install .
+          pip install pytest
```
ZoranPandovski, line 23: @hamishfagg Do we need to install `responses` here too? It looks like the tests fails when importing `responses`

ZoranPandovski (changes requested):

.github/workflows/testing.yml
hamishfagg: we need to install the testing requirements file, which I've added now. We shouldn't be installing any deps individually, because it means out requirements are out of date

ZoranPandovski (approved):

dusvyat (approved):
@hamishfagg lgtm

PR #7855, opened 2023-10-17
fix: Advance queries fixed for influx DB

### LINKED ISSUE #7766: [Bug]: InfluxDB add filter option
The current integration of InfluxDB pulls the whole data from the table and ignores the WHERE clause. We should add filtering option https://github.com/mindsdb/mindsdb/tree/staging/mindsdb/integrations/handlers/influxdb_handler

Tanmai2002:
## Description

The Advance Queries for Influx Table added

Fixes #7766

## Type of change

## Verification Process
image

image

image


MinuraPunchihewa (changes requested):
Hey @Tanmai2002,
The way this has been implemented will fetch all of the data in a table and then apply the various clauses of the SELECT statement as you can see here,
```
    def call_influxdb_tables(self):
        """Pulls all the records from the given  InfluxDB table and returns it select()
    
        Returns
        -------
        pd.DataFrame of all the records of the particular InfluxDB 
        """
        url = f"{self.connection_data['influxdb_url']}/query"

        params = {
            ('db',f"{self.connection_data['influxdb_db_name']}"),
            ('q','SELECT * FROM '+ f"{self.connection_data['influxdb_table_name']}" )
        }
        headers = {
            "Authorization": f"Token {self.connection_data['influxdb_token']}",
            "Accept": "application/csv",
        }

        response = requests.request("GET",url,params=params,headers=headers)
        influxdb_df = pd.read_csv(io.StringIO(response.text))

        return influxdb_df
```

This is not very efficient. Would it be possible for you to have look at fetching only the relevant data? I think this can be done by adding to the SQL statement that is passed to the above method.

Tanmai2002:
Surely, I later realised it. Ill be updating the PR. Thank you

Tanmai2002:
I Rephrased the Query. Advance Queries are working  and without 
image

I tried replacing the SQL directly, but there were issues, so I had to reassemble the query. Thank you

Tanmai2002:
@MinuraPunchihewa Kindly Review the PR

mindsdb/integrations/handlers/influxdb_handler/influxdb_tables.py
```diff
@@ -27,68 +29,52 @@ def select(self, query: ast.Select) -> pd.DataFrame:
...
+        except Exception as e:
+            logger.warn(e)
+
+        formatted_query=self.get_select_query(table_name,selected_columns,where_conditions,order_by_conditions,query.limit)
+        influxdb_tables_df  = self.handler.call_influxdb_tables(formatted_query)
 
         return influxdb_tables_df
 
-    def get_columns(self) -> List[str]:
+    def get_columns(self,dataframe) -> List[str]:
```
MinuraPunchihewa, line 53: Hey @Tanmai2002,
If you take a look at the `APITable` interface, `get_columns()` is an abstract method that needs to be implemented. Therefore, let's not change the method signature and keep in the same. Since this table is fixed, I think you can move your logic of getting the columns from the head of the `DataFrame` directly to this method without requiring a `dataframe` parameter. I mean this,
`head_influx_table = self.handler.call_influxdb_tables(f"SELECT * FROM {table_name} LIMIT 1")`

MinuraPunchihewa (changes requested):

ZoranPandovski (changes requested):
Can you please pull the latest changes to avoid the chromadb lint errors so we can merge this?

Tanmai2002:
Done @ZoranPandovski

Tanmai2002:
@ZoranPandovski @MinuraPunchihewa Kindly Check

ZoranPandovski (approved):

PR #8010, opened 2023-10-23
Added the pysqlite3-binary Dependency for Chroma

MinuraPunchihewa:
## Description

This PR adds the `pysqlite3-binary` for the Chroma because it is currently not working in MindsDB Cloud. This is the error that has been encountered,
`Your system has an unsupported version of sqlite3. Chroma requires sqlite3 >= 3.35.0.`

Because the version of the `sqlite3` module is tied to the Python version, I thought this would be the best way to go about it. This works locally after some initial testing, but the problem exists in Cloud.

Fixes https://github.com/mindsdb/mindsdb/issues/7949

## Type of change

## Verification Process

 - [X]   Initialize the handler locally to ensure it works
 - [ ]   Initialize the handler on Cloud to ensure the fix works

## Checklist:

### DIFF
--- a/mindsdb/integrations/handlers/chromadb_handler/chromadb_handler.py
+++ b/mindsdb/integrations/handlers/chromadb_handler/chromadb_handler.py
@@ -1,12 +1,13 @@
+__import__('pysqlite3')
+
+import sys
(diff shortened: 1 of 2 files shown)


dusvyat (approved):
let's test it on alpha

PR #8027, opened 2023-10-24
Test that all requirements files are inter-compatible

hamishfagg:
## Description

Adds a test that checks that all requirements files are compatible with each other. With so many requirements files it's difficult to know if changes cause conflicts between handler depends.

This check currently fails - we will need some ml/handler experts to help sort out the conflicts before merging this.


ZoranPandovski (approved):
This is great

PR #8134, opened 2023-10-26
Requirements cleanup

hamishfagg:
## Description

This is trying to resolve our deep and long-standing issue with the requirements in MindsDB.  This...

* Consolidated all requirements into the requirements folder (no more requirements files in the root)
* Auditing and cleaning up our requirements in every requirements file in this repo
    * Utilizing tools like [deptry](https://fpgmaas.github.io/deptry/) to help accomplish this
    * Removing requirements in handlers which would already be present in core requirements
    * TODO: Adding cross-handler requirements (relative links?)
* Removed empty requirements files to K.I.S.S./D.R.Y. things
* Adding automated CI jobs which run [deptry](https://fpgmaas.github.io/deptry/) to...
    * Validate someone doesn't add breaking or unused requirements to our "root" requirements file
    * Validate someone doesn't add breaking or unused requirements to our "handlers" requirements files
    * Check if a requirements in a handler is already included in our root requirement, thus preventing adding a useless and duplicated requirement (keep it D.R.Y.)
* Removed "with_nlp" unused code blocks and includes (old code)
* Modified setup.py to...
    * Add all handlers as valid extras, even if they don't have requirements
    * Added our "official" handlers as default extras
    * Added env var `MINDSDB_PIP_INSTALL_DEFAULT_EXTRAS` if set to false, disables default handlers which allows for an advanced user to have a minimalistic and/or laser focused MindsDB install with only a desired set of extras.  This will allow for creating smaller Docker images and such.
* Fixed some minor non-standard Python formatting issues

Ive had to disable some of the CI/CD handler tests until we refactor them to install the handler and/or not run them when we haven't changed the handler code.

## TODO

- [x] Fix/update handlers which rely on other handlers with relative req includes
- [ ] Validate the ENV var switch off logic works on a test Pypi
- [ ] Perform various tests to ensure everything works
- [x] Figure out how to resolve the deps for the base + default handlers

## Type of change

- [X] 📢 Breaking change (fix or feature that would cause existing functionality not to work as expected)


mindsdb/integrations/handlers/flaml_handler/requirements.txt
```diff
@@ -1,2 +1,2 @@
 flaml<=1.2.3
-dill
+type-infer
```
paxcema, line 2: `type-infer` is part of our ML packages. I think we probably want to make it part of our defaults given we're recommending people to actively use it when developing new ML engines.

It is fairly light in deps ([source](https://github.com/mindsdb/type_infer/blob/staging/pyproject.toml)), but if you think there are any here that would be problematic we can keep it as optional.

tests/scripts/check_duplicate_reqs.py
```diff
@@ -0,0 +1,33 @@
...
+    main_reqs = [
+        re.split(pattern, line)[0]
+        for line in main_f.readlines()
+        if re.split(pattern, line)[0]
+    ]
+
+reqs_files = glob.glob("**/requirements*.txt", recursive=True)
+reqs_files.remove(main_reqs_path)
+
+# print(main_reqs)
```
paxcema, line 19: nit: leftover print

requirements/requirements.txt
```diff
@@ -1,26 +1,18 @@
-Werkzeug==1.0.1
-lightwood >= 23.8.1.0, < 23.8.2.0
```
paxcema, line 2: Now that lightwood is not a main req anymore, it looks like if a user installs "core" mindsdb only, there would be no way of creating models. Not saying this is bad, per se. After all we've been meaning to write a lightweight default ML engine for a while now (see #5089). However it would be good to highlight this in the PR description, seems important.

paxcema (approved):
Great work @hamishfagg, much needed! 💪 

Left a couple of minor point that ideally would be addressed, but this looks good to merge as-is to me.

requirements/requirements.txt
mindsdb-devops: Technically this statement is not true @paxcema in setup.py we have a handful of "default extras" to install which Lightwood is currently part of.  So when pip installing you will still get lightwood installed.

However, related to this, after discussion with Jorge I am writing up a discussion for us to discuss as a team removing ALL internal ML engines from our default pip package in a future PR to simplify our package, shrink our docker image(s), improve our potential adoption in open source, etc.  If/when we do THAT we will need to coordinate with our documentation to update it along side our code.

mindsdb/integrations/handlers/flaml_handler/requirements.txt
hamishfagg: We're trying to only have things in the main requirements.txt that are used in the core codebase. That said, with the list of default handlers installed when you install `mindsdb`, it should get installed in most cases anyway

paxcema: Okay, let's leave it out of default/core reqs 👍

requirements/requirements.txt
paxcema: Thanks for the explanation! I see now that "default extras" include Lightwood indeed.

We're good as long as there's awareness that moving Lightwood out of the default/core stack will have implications in terms of "ML ease of use" for the average user, so whenever we end up doing this it must be communicated properly, as you said.

PR #8216, opened 2023-10-31
Fix ingress backend service mismatch

thien-nguyenthanh:
## Description

The PR addresses a backend service mismatch issue in our Helm chart's Ingress configuration. The mismatch occurs when the backend service specified in the Ingress resource does not align with the desired service for the application. This discrepancy could lead to routing errors and unreliable service discovery.


ZoranPandovski:
Closing as outdated, the helm support was removed

PR #8262, opened 2023-11-03
Added the Unit Tests for the Anthropic Integration

MinuraPunchihewa:
## Description

This PR adds the unit tests for the Anthropic integration. It also adds `BaseMLTest` and `BaseMLAPITest` classes, which contains some of the common functionality used throughout the tests for ML Engines (and API-based ML engines).

Fixes https://github.com/mindsdb/mindsdb/issues/8261

## Type of change

- [X] ⚡ Unit tests

## Verification Process

 - [X]   Test Location: pytest tests/unit/ml_handlers/test_anthropic.py

## Additional Media:

Test report (summary),
image

One unit test is failing due to a limitation in the handler; unknown model arguments are not handled.

## Checklist:

### DIFF
--- a/tests/unit/ml_handlers/base_ml_test.py
+++ b/tests/unit/ml_handlers/base_ml_test.py
@@ -5,6 +5,7 @@ import pandas as pd
 from mindsdb_sql import parse_sql
 from ..executor_test_base import BaseExecutorTest
 
+
 class BaseMLTest(BaseExecutorTest):
     """
     Base test class for ML engines
--- a/tests/unit/ml_handlers/base_ml_test.py
+++ b/tests/unit/ml_handlers/base_ml_test.py
@@ -1,0 +1,31 @@
+import time
+import pandas as pd
+
+from mindsdb_sql import parse_sql
+from ..executor_test_base import BaseExecutorTest
+
+class BaseMLTest(BaseExecutorTest):
+    """
+    Base test class for ML engines
+    """
+    def wait_predictor(self, project, name, timeout=100):
+        """Wait for the predictor to be created, raising an exception if predictor creation fails or exceeds timeout"""
+        for attempt in range(timeout):
+            ret = self.run_sql(f"select * from {project}.models where name='{name}'")
+            if not ret.empty:
+                status = ret["STATUS"][0]
+                if status == "complete":
+                    return
+                elif status == "error":
+                    raise RuntimeError("Predictor failed", ret["ERROR"][0])
+            time.sleep(0.5)
+        raise RuntimeError("Predictor wasn't created")
+    
+    def run_sql(self, sql):
+        """Execute SQL and return a DataFrame, raising an AssertionError if an error occurs"""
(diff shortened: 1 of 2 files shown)


paxcema:
@hamishfagg tagging you for reference given the refactor effort you're going through at the moment

tests/unit/ml_handlers/base_ml_test.py
```diff
@@ -0,0 +1,42 @@
...
+            columns = [col.alias if col.alias else col.name for col in ret.columns]
+            return pd.DataFrame(ret.data, columns=columns)
+
+
+class BaseMLAPITest(BaseMLTest):
+    """
+    Base test class for API-based ML engines
+    """
+    @staticmethod
+    def get_api_key(env_var: str):
```
paxcema, line 40: There is a more robust method for this in "integrations/utilities/handler_utils:get_api_key`, we probably want to use that instead.

tests/unit/ml_handlers/base_ml_test.py
```diff
@@ -0,0 +1,42 @@
+import os
+import time
+import pandas as pd
+
+from mindsdb_sql import parse_sql
+from ..executor_test_base import BaseExecutorTest
+
+
+class BaseMLTest(BaseExecutorTest):
```
paxcema, line 9: Please open an issue so that we eventually refactor all ML tests to use this class 👍

paxcema (changes requested):

tests/unit/ml_handlers/base_ml_test.py
MinuraPunchihewa: Understood, @paxcema. I will incorporate that method.

tests/unit/ml_handlers/base_ml_test.py
MinuraPunchihewa: Will do, @paxcema.

tests/unit/ml_handlers/base_ml_test.py
MinuraPunchihewa: Hey @paxcema,
I was just taking a look at the function that you mentioned and I had some thoughts,
1. Since we are running unit tests here, is it correct to assume that we will always be accessing the API keys using environment variables?
2. Are the `create_args` and `engine_storage` parameters relevant here? Because we are running SQL statements to perform our operations?

Based on the above, should this be the way to use the function within `BaseMLAPITest` class?
```
class BaseMLAPITest(BaseMLTest):
    """
    Base test class for API-based ML engines
    """
    @staticmethod
    def get_api_key(api_name: str):
        """Retrieve API key from environment variables"""
        return get_api_key(api_name, {}, None, strict=True)
```
Please let me know what you think, Patricio.

paxcema: Point 1) is actually great. Disregard my original comment and let's keep this to avoid overcomplicating things 😄 sorry for the nuisance

paxcema (approved):

PR #8264, opened 2023-11-03
update writer docs and local install requirements.txt

dusvyat:
## Description

update writer readme and local install requirements.txt

## Type of change


mindsdb/integrations/handlers/writer_handler/requirements.txt
```diff
@@ -1,14 +1 @@
...
-duckdb==0.7.1
-pydantic>=1.10.8
-sentence_transformers==2.2.2
-markupsafe == 2.0.1
-faiss-cpu>=1.7.4 #todo check support for 'faiss-gpu'
-datasets>=2.14.4
-nltk>=3.8.1
-rouge-score>=0.1.2
-dill==0.3.6 #for lightwood
+-r ../rag_handler/requirements.txt # same requirements
```
ZoranPandovski, line 1: This will work with `pip install` right? do we need the -r or it can be used from `pip install -r ,,,`>

ZoranPandovski (changes requested):
Looks good, I;ve added one minor comment

mindsdb/integrations/handlers/writer_handler/requirements.txt
dusvyat: yep it works, I tested it out locall.y

the idea is if user `pip -r ../writer_handler/requirements.txt`, it will install requirments from `rag_handler` dir.

As there is a massive overlap, it figured it is easier to maintain a single list of requirements

dusvyat (commented):
fixed, for now I will simply duplicate `rag_handler `requirements inside `writer_handler` to get over these checks in `setup.py`

ZoranPandovski:
Great, we can add enhancement issue to improve that in the future

mindsdb/integrations/handlers/writer_handler/requirements.txt
Wuflamez: -r ,,,

PR #8428, opened 2023-11-26
Docker tidy-up

hamishfagg:
## Description

- Remove files for unused microservices mode (all-in-one docker-compose, plus microservice dockerfiles)
- Remove nginx compose and config, was used for microservices
- Remove beta and release dockerfiles. One dockerfile to rule them all!
- Get rid of "auto updating" docker images. User's can update their image if they want a newer version. This only stands to cause issues
- Don't install from PyPI at build time, install from the repo.
- Add some dockerhub images to our bake file
- Add a 'dev' stage that installs `requirements-dev.txt`
- Get rid of beta config.json. With new logging, the release and beta configs are the same (updated docs)
- Remove `MDB_CONFIG_CONTENT` env var and `mindsdb_launcher.sh`
- Update the release github action to use docker bake

## Type of change


mindsdb-devops (approved):
LGTM

PR #8453, opened 2023-12-01
OpenBB Handler & Agent Improvements

tmichaeldb:
## Description

This PR makes various improvements for OpenBB & Agents:
- Handle `LIMIT` and `ORDER_BY` correctly
- Use `duckdb` to handle functions in query targets (i.e. `AVG`, `MAX`, etc)
- Add a more detailed prompt/description for the text-to-SQL tool so agents can automatically determine which tables to query based on context
- Don't return 500 errors on the `/analysis/data` endpoint when analysis fails (can't properly handle dict return types from OpenBB). Long-term we should fix the `/analysis/data` endpoint, but the user should not see cryptic 500 errors on the GUI when this happens.
- Answer prompts in separate threads in `langchain_handler`

See the following demos to see how to use OpenBB with Chatbots:
- [MindsDB & OpenBB integration](https://www.loom.com/share/bc3e8a5fd013446c8f03fed69854be7e)
    - [SQL commands here](https://pastebin.com/8spNESHE) 
- [World News Knowledge Base](https://www.loom.com/share/859d9e672fce43e7b3f9212bb02a6fc1)
    - [SQL commands here](https://pastebin.com/gTCypTW9)
- [Creating & Publishing a Chatbot](https://www.loom.com/share/b674b13368e148fd91a85980f7d4b8c9)
    - [SQL commands here](https://pastebin.com/ep2QQsVY)
- [Chatbot Demo with text-to-SQL and knowledge bases](https://www.loom.com/share/42df3f26f29f403b9eaea9595a925974)

## Type of change

## Verification Process

 - [x]   Verification Steps: Experiment with text-to-SQL, OpenBB handler, and agents locally

## Additional Media:

## Checklist:

### DIFF
--- a/mindsdb/integrations/handlers/langchain_handler/langchain_handler.py
+++ b/mindsdb/integrations/handlers/langchain_handler/langchain_handler.py
@@ -2,7 +2,7 @@ import re
 import os
 from typing import Optional, Dict
 
-from concurrent.futures import as_completed, ThreadPoolExecutor, TimeoutError
+from concurrent.futures import as_completed, TimeoutError
 import numpy as np
 import pandas as pd
 
@@ -21,6 +21,7 @@ from mindsdb.integrations.handlers.langchain_handler.tools import setup_tools
 from mindsdb.integrations.libs.base import BaseMLEngine
 from mindsdb.integrations.utilities.handler_utils import get_api_key
 from mindsdb.utilities import log
+from mindsdb.utilities.context_executor import ContextThreadPoolExecutor
 
 
 _DEFAULT_MODEL = 'gpt-3.5-turbo'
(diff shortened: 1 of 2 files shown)


DidierRLopes:
Legend.

Thank you for this @tmichaeldb

ZoranPandovski (approved):
LGTM. @TylerSandman We had some compatibilities issues between llama and langchain, is this version `langchain==0.0.347` compatible for both?

mindsdb/integrations/handlers/openbb_handler/openbb_tables.py
```diff
@@ -43,6 +44,20 @@ def _get_params_from_conditions(self, conditions: List) -> Dict:
 
             return params
 
+        def _select_dataframe(self, query: ast.Select, df: pd.DataFrame) -> pd.DataFrame:
```
ea-rus, line 47: looks like this function could be used instead: mindsdb.api.mysql.mysql_proxy.utilities.sql.query_df

ea-rus:
Also in code of openbb handler: looks like self.obb is a global variable. It won't work on cloud (connection will be shared between users)

I think it should be replaced to something like:
```python
from openbb_core.app.static.app_factory import create_app
self.obb = create_app()
```

PR #8454, opened 2023-12-01
Build arm64 docker images

hamishfagg:
Currently we only build amd64 containers, and these won't run on M1/M2 macs even with `--platform linux/amd64` - they crash with `illegal instruction` due to a library compiled for amd64 (I suspect numpy related). We have had a few community reports of this happening.

If we can provide arm64-native containers then folks will be able to run mindsdb on their m1/m2 mac without issue.


mindsdb-devops (approved):
LGTM

PR #8475, opened 2023-12-07
Fix Knowledge Base Insert

tmichaeldb:
## Description

When inserting values into a knowledge base (e.g. [this demo](https://pastebin.com/gTCypTW9)), it currently fails with a `"None of [Index([('i', 'd'), ('i', 'd')], dtype='object')] are in the [index]"` error due to how we are inserting. Additionally, the way we fill in missing IDs when the ID column is present causes errors as well.

This PR fixes both of those issues.

Fixes #issue_number

## Type of change

## Verification Process

 - [x]   Verification Steps: Create and insert into a knowledge base locally

## Additional Media:

## Checklist:

### DIFF
--- a/mindsdb/integrations/libs/vectordatabase_handler.py
+++ b/mindsdb/integrations/libs/vectordatabase_handler.py
@@ -329,7 +329,9 @@ class VectorStoreHandler(BaseHandler):
             df[id_col] = df[content_col].apply(gen_hash)
         else:
             # generate for empty
-            df.loc[df[id_col], id_col] = df[content_col].apply(gen_hash)
+            for i in range(len(df)):
+                if pd.isna(df.loc[i, id_col]):
+                    df.loc[i, id_col] = gen_hash(df.loc[i, content_col])
 
         # remove duplicated ids
         df = df.drop_duplicates([TableField.ID.value])
--- a/mindsdb/interfaces/knowledge_base/controller.py
+++ b/mindsdb/interfaces/knowledge_base/controller.py
@@ -123,11 +123,10 @@ class KnowledgeBaseTable:
 
         # add embeddings
         df_emb = self._df_to_embeddings(df)
-        df = pd.concat([df, df_emb], axis=1)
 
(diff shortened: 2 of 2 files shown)


QuantumPlumber (approved):
Looks good to me, and appears to work.

mindsdb/integrations/libs/vectordatabase_handler.py
```diff
@@ -329,7 +329,9 @@ def gen_hash(v):
             df[id_col] = df[content_col].apply(gen_hash)
         else:
             # generate for empty
-            df.loc[df[id_col], id_col] = df[content_col].apply(gen_hash)
```
ea-rus, line 332: looks like here was missed .isna()
It should be:
```python
df.loc[df[id_col].isna(), id_col] = df[content_col].apply(gen_hash)
```

mindsdb/interfaces/knowledge_base/controller.py
```diff
@@ -123,11 +123,10 @@ def insert(self, df: pd.DataFrame):
 
         # add embeddings
         df_emb = self._df_to_embeddings(df)
-        df = pd.concat([df, df_emb], axis=1)
 
         # send to vector db
         db_handler = self._get_vector_db()
-        db_handler.do_upsert(self._kb.vector_database_table, df)
+        db_handler.do_upsert(self._kb.vector_database_table, df_emb)
```
ea-rus, line 129: df_emb doesn't content text ('content' value). how it works for you?

QuantumPlumber: It works on my system.. is there an error it is throwing for you?

ea-rus: Discussed yesterday with @tmichaeldb 
Some embedding models returns input column (cohere) and some not (openai)
- _df_to_embeddings should filter model output and return only one column (embeddings)
- return concat of  df and df_emb

PR #8523, opened 2023-12-18
added missing reqs to youtube handler

martyna-mindsdb:
## Description

added missing reqs to youtube handler

discussed here: https://mindsdbcommunity.slack.com/archives/C04EWFQEFEV/p1702899569784249

Fixes #issue_number

## Type of change

### DIFF
--- a/mindsdb/integrations/handlers/file_handler/requirements.txt
+++ b/mindsdb/integrations/handlers/file_handler/requirements.txt
@@ -1,5 +1,4 @@
 charset-normalizer
 python-magic >= 0.4.27
-langchain==0.0.347
 pypdf  # Optional dep of langchain for PDF files
 openpyxl  # Optional dep of langchain for txt files
--- a/mindsdb/integrations/handlers/gmail_handler/requirements.txt
+++ b/mindsdb/integrations/handlers/gmail_handler/requirements.txt
@@ -1,3 +1,2 @@
 google-api-python-client
 google-auth-httplib2
-google-auth-oauthlib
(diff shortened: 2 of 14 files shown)


ZoranPandovski (approved):

ZoranPandovski:
Is this required @martyna-mindsdb there is a check that fails for:

```
    None:None: DEP002 'google-auth-oauthlib' defined as a dependency but not used in the codebase
```

martyna-mindsdb:
@ZoranPandovski 
I had to install this package to have the YouTube handler available in MindsDB. Also, another user had to do the same.

The error message in the handlers table says explicitly that this package is missing, so I guess it is required for YouTube:

image

MinuraPunchihewa (commented):
Hey @ZoranPandovski,
Yes, I believe this is required. I think the reason that the CI check is failing is because the code that actually uses this package is [here](https://github.com/mindsdb/mindsdb/blob/staging/mindsdb/integrations/handlers/utilities/auth_utilities/google_oauth_utilities.py).
This is not ideal, but I think the way this can be solved is probably to do,
`import google_auth_oauthlib # noqa, required by GoogleOAuth2Manager`

hamishfagg:
This has highlighted a blind spot with the requirements checks that it didn't check `mindsdb/integrations/handlers/utilities` with the core codebase.
So I've fixed the check, and fixed the errors caused by checking ALL of the code now :)

I've moved `google-auth-oauthlib` to the main requirements file, as it is imported in the core codebase. So checks should pass now

ZoranPandovski (approved):

PR #8559, opened 2023-12-27
Add new GitHub Loader llama_index Handler

pranavvp16:
The Existing llama Index Handler was only able to use Webpages and documents as knowledge source. This integration allows mindsdb to use Github Repositories as knowledge source with the help of llama Index github loader. 

Screenshots of working handler
image
image
image

### DIFF
--- a/mindsdb/integrations/handlers/llama_index_github_handler/README.md
+++ b/mindsdb/integrations/handlers/llama_index_github_handler/README.md
@@ -1,0 +1,42 @@
+## LlamaIndex Github Handler
+LlamaIndex Github ML handler for MindsDB, create indexes over data plugged in mindsdb from GitHub Repositories and use it to create a Question & Answer (Q&A) system
+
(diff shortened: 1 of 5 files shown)


pranavvp16:
Please review this Handler ASAP as I want to use it Mindsdb X Hashnode hackathon

paxcema (commented):
Why add a brand new integration instead of extending the current one? I think it'd be best if we improve the existing integration, and thinking of support for any given loader as a "feature" in it. Thoughts @ZoranPandovski?

pranavvp16:
Yes I also thought to improve the existing llama_index integration, but this integration has many arguments which are different and it would be hard to add it in the existing integration. Keeping things sorted I thought it would be better to add a whole new integration. I had a talk regarding this with @ZoranPandovski

pranavvp16:
Closing this cause https://github.com/mindsdb/mindsdb/pull/8562

PR #8568, opened 2023-12-31
[fix] Update lightwood handler requirements

paxcema:
## Description

This avoids packages like `qiskit` blocking the handler installation in macOS. This one in particular and `image` are experimental and not meant to be supported in MindsDB (yet). There are others, like `extra` and `extra_ts` which we may want to offer in the future.

## Type of change


hamishfagg:
From memory we previously fixed "all_extras" because we needed pillow in mindsdb @paxcema?
I think we had people complaining that they wanted to use image processing and couldn't because pillow wasn't there.

Is this image processing something that mindsdb won't support for now? Or should we add pillow to the lightwood handler requirements?

hamishfagg (approved):

paxcema:
Deferring merging to @StpMax on this one, as he mentioned he'd already fixed this with another approach.

PR #8593, opened 2024-01-08
FIX llama_index Handler

pranavvp16:
This PR fixes 3 issues in `llama_index_handler`

1. The `get_api_key` reverted changes where not merged properly, which is now fixed @MinuraPunchihewa 
2. `OpenAI(**llm_kwargs)` defaults the model to `text-davinci-003` which has been shutdown by [OpenAI](https://platform.openai.com/docs/deprecations). So `model = gpt3.5-turbo-instruct` is been passed through llm kwargs #8592 
3. `LLMPredictor` from llama_index has been deprecated  and doesn't work anymore


mindsdb/integrations/handlers/llama_index_handler/llama_index_handler.py
```diff
@@ -304,3 +282,37 @@ def _setup_index(self, documents):
         )
 
         return index
+
+    def _get_llama_index_api_key(self, args, strict=True):
```
ZoranPandovski, line 286: There is a util function for this so better to use one in all handlers https://github.com/mindsdb/mindsdb/blob/82dfb9750a3bd3dfec5adc86db5ba6c64a204fd2/mindsdb/integrations/utilities/handler_utils.py#L10

ZoranPandovski (changes requested):

mindsdb/integrations/handlers/llama_index_handler/llama_index_handler.py
pranavvp16: @ZoranPandovski I implemented this in the previous PR please check, this util can only be used to take input of `api_key`, what I want is `openai_api_key` and `github_token`. So I had to revert my changes #8593

paxcema: For reference: relevant PR that should fix what you stumbled upon is #8245 (@martyna-mindsdb), relevant issue is #7496.

pranavvp16: Why don't we just add one more parameter to the `get_api_key()` such as key which will be passed accordingly @paxcema

pranavvp16: [reverted change](https://github.com/pranavvp16/mindsdb/blob/4fd457cbc5c4d3b4af54339b2825114c3b9c340f/mindsdb/integrations/utilities/handler_utils.py) something like this, which will make the utils usable to great extent. Also I searched for usage, the util has somewhat 5 usages, So we can change them too. If this isn't possible please suggest me a change, as this makes the llama_index handler unusable totally. And I have built my hackathon project using it!

pranavvp16: @paxcema #8245 adds a prefix `{name}_api_key` which won't work for `github_token`. Please look at the [reverted changes](https://github.com/pranavvp16/mindsdb/blob/4fd457cbc5c4d3b4af54339b2825114c3b9c340f/mindsdb/integrations/utilities/handler_utils.py) this seems the best approach to me.

paxcema:
@pranavvp16 out of curiosity, by "reverted changes where not merged properly" do you mean the breaking change introduced in #8562 where any model with this handler now fails to train? Error message is:

```python
TypeError: get_api_key() got multiple values for argument 'strict'
```

pranavvp16:
@paxcema yes yes that's what I'm trying to say, if you look at the commit history of the #8562 it was in working state, later I was asked to use the `get_api_key()` util, but due to the limitation caused by it I made some change in the util, but was later asked to keep the handler same as before removing the `get_api_key()` util. So I reverted by changes back, but it seems there was a problem in the merge and `llama_index_handler.py` didn't change. So I made this PR fixing the handler with the reverted changes.

mindsdb/integrations/handlers/llama_index_handler/llama_index_handler.py
```diff
@@ -269,31 +256,22 @@ def predict(
...
-            openai_api_key = get_api_key(
-                key.lower(), "llama_index", args["using"], engine_storage, strict=True
-            )
-
+        openai_api_key = self._get_llama_index_api_key(args["using"])
         openai.api_key = openai_api_key  # TODO: shouldn't have to do this! bug?
-        llm_kwargs = {"openai_api_key": openai_api_key}
+        llm_kwargs = {
+            "openai_api_key": openai_api_key,
+            "model": "gpt-3.5-turbo-instruct",
```
paxcema, line 263: Let's switch to `gpt-3.5-turbo`, following [OpenAI's official recommendation](https://platform.openai.com/docs/models/gpt-3-5). Also, please comment this line to indicate we should remove this specific parameter enforcing once the `langchain` dependency is updated.

mindsdb/integrations/handlers/llama_index_handler/llama_index_handler.py
```diff
@@ -269,31 +256,22 @@ def predict(
...
+        }
         if "temperature" in args["using"]:
             llm_kwargs["temperature"] = args["using"]["temperature"]
         if "model_name" in args["using"]:
-            llm_kwargs["model_name"] = args["using"]["model_name"]
+            llm_kwargs["model"] = args["using"]["model_name"]
```
paxcema, line 268: Is this key renaming correct? Previously we were passing `model_name` to `langchain.llm.OpenAI`. Does `model` match the expected parameter?

mindsdb/integrations/handlers/llama_index_handler/llama_index_handler.py
paxcema: Let's go for this as a hotfix for now, because the handler is broken atm @pranavvp16 @ZoranPandovski. But we need to refactor this ASAP as there's a lot of duplicated logic going on at the moment

paxcema (changes requested):

mindsdb/integrations/handlers/llama_index_handler/llama_index_handler.py
pranavvp16: We need instruct models for this handler like the `text-davinci-003`, passing a chat model throws error from openai. Something like `Received v1 endpoints chat model` .

mindsdb/integrations/handlers/llama_index_handler/llama_index_handler.py
pranavvp16: yes this was wrong the object takes model as a argument

pranavvp16: Wait I'll check this once again, and update this correctly

pranavvp16: We can pass both `model` and `model_name` still changing this to `model_name` as recommended in this [link](https://github.com/langchain-ai/langchain/issues/4331)

mindsdb/integrations/handlers/llama_index_handler/llama_index_handler.py
pranavvp16: currently passing a chat model like `gpt-3.5-turbo` will fail due to OpenAI's migration which gives below error. So I have added a `v1/completion` version of `gpt-3.5-turbo` which is `gpt-3.5-turbo-instruct`. 
Screenshot 2024-01-16 at 9 37 38 PM

pranavvp16:
@paxcema I have changed the code as per request also added a review regarding the `gpt-3.5-turbo` . This needs to be fixed and mindsdb team will need to migrate to `openai>=1.0.0` to continue using chat models. This can be merged now.

torrmal:
@paxcema what do you think?

mindsdb/integrations/handlers/llama_index_handler/llama_index_handler.py
paxcema: @pranavvp16 please put a `# TODO` comment on this so that we can refactor ASAP

paxcema (changes requested):
@pranavvp16 Try merging `staging` into this branch. We've since updated the `openai` requirement to `>=1.0.0` so you should be able to fix the error above.

pranavvp16:
Seems I merged incorrectly opened a new pull request #8733

PR #8609, opened 2024-01-10
More deploy triggers and simplify docker

hamishfagg:
- Add staging and prod deployments on merge to stable/releases respectively.
- Simplify the docker bake file
- On release, push versioned images to ECR as well as dockerhub (this just makes env cleanup easier later on)


mindsdb-devops (approved):
lgtm

PR #8692, opened 2024-01-25
[ENH] Improved data preparation for LLM finetuning

paxcema:
## Description

This PR introduces some common data preparation utilities for fine-tuning LLMs that operate following the ChatCompletion format from OpenAI. Currently, these coupled versions of these methods are actively being used in our integrations for Anyscale and OpenAI (hence indirectly by LangChain, too).

Added tests for all of these, so that we can use them on any other LLM engines.

## Type of change

- [x] ⚡ New feature (non-breaking change which adds functionality). Refactor.

## Verification Process

 - [x]   Test Location: added unit tests, which pass.

## Additional Media:

## Checklist:


tests/unit/test_llm_utils.py
```diff
@@ -0,0 +1,178 @@
...
+                {"role": "user", "content": "hi"},
+                {"role": "assistant", "content": None},  # should always be a string
+            ],
+
+            # invalid - no assistant in the chat
+            [
+                {"role": "user", "content": "hi"},
+            ],
+        ]
+
```
dusvyat, line 112: minor nit: perhaps we can make invalid_chats and valid_chats a fixture. Stylistic, doesn't block merge

dusvyat (approved):
LGTM, see minor point

mindsdb/integrations/handlers/anyscale_endpoints_handler/anyscale_endpoints_handler.py
```diff
@@ -119,63 +120,35 @@ def _check_ft_cols(df, cols):
...
-            chats.append({'messages': chat})
+    # TODO: add unit tests
+    @staticmethod
+    def _prepare_ft_jsonl(df, temp_storage_path, temp_filename, _, test_size=0.2):
+        # 1. format data
+        chats = ft_chat_formatter(df)
 
+        # 2. split chats in training and validation subsets
         series = pd.Series(chats)
         if len(series) < 20 * 2:
```
pedrofluxa, line 131: This seems a little arbitrary; shouldn't the length of the dataset depend on how large the model is? I would at least leave that "20" as a keyword parameter with a default value.

pedrofluxa (approved):
Left a comment on a hard-coded value. I am approving this PR as it should work as it is, but I strongly suggest to refactor the code a little bit.

mindsdb/integrations/handlers/anyscale_endpoints_handler/anyscale_endpoints_handler.py
paxcema: Good shout, I'll change the row number to be customizable. As for why: Anyscale endpoints has a non-documented cutoff at 20 -- if the validation dataset is smaller than that, the entire process fails. I'll add this info. as a comment.

Could you expand on how I should go about this one?
> I strongly suggest to refactor the code a little bit.

chandrevdw31:
Unable to Finetune as it says False 

image

paxcema:
@chandrevdw31 status is marked as `generating` in the screenshot above, and these finetune runs can easily take hours so this is normal. The model cannot be `active` before its status is `complete`. 

If the user reports an actual error in the model record, please ask them to open a bug with it.

PR #8711, opened 2024-01-30
Move lanchain skills to langchain handler.

hamishfagg:
## Description

This fixes an issue that has caused our base docker image to grow from ~500MB to 3GB.
A dependency. of the langchain skills was added that eventually requires torch. The goal is for our base image to not require torch and be a lot smaller as a result.

This PR moves the skills code into the langchain handler, which can then not be installed by default. The langchain handler is the only handler that uses this code at present.

## Type of change


hamishfagg:
@ea-rus @paxcema I've asked for your reviews as I don't understand the langchain skills and I don't know how to test my changes. They are deployed to https://alpha.dev.mindsdb.com now if you could take a look please :)

paxcema:
Looping in @tmichaeldb as well. Code changes look good to me, though it's possible we may want to bring these back into `interfaces/skills` if/once ML engines other than Langchain can be used as agents.

For that, it'd be important to understand what is using `torch` here. I may be mistaken but my reading is that we're not directly using it in any of the `skill_tool` imports, so there may be a way to refactor this so that we can move tools back to their original place without needing `torch`.

ea-rus:
@paxcema, skill_tool.py uses langchain handler. When we return it to skills controller it should be set of functions used by langchain handler (but not dependent on langchain)

PR #8719, opened 2024-01-31
[fix] #6821 - deprecate normal completion mode for OpenAI handler

### LINKED ISSUE #6821: [OpenAI] Deprecate non-chat text completion endpoint
As seen [here](https://openai.com/blog/gpt-4-api-general-availability), we should disable this endpoint sooner rather than later.

paxcema:
## Description

This PR deprecates the normal completion mode for OpenAI handler. Default mode now is `conversational` which uses the `ChatCompletion` endpoint.

Fixes #6821

## Type of change

## Verification Process

## Additional Media:

## Checklist:

### DIFF
--- a/mindsdb/integrations/handlers/openai_handler/openai_handler.py
+++ b/mindsdb/integrations/handlers/openai_handler/openai_handler.py
@@ -43,10 +43,9 @@ class OpenAIHandler(BaseMLEngine):
         self.default_model = 'gpt-3.5-turbo'
         self.default_image_model = 'dall-e-2'
         self.default_mode = (
-            'default'  # can also be 'conversational' or 'conversational-full'
+            'conversational'  # can also be 'conversational-full'
         )
         self.supported_modes = [
-            'default',
             'conversational',
             'conversational-full',
             'image',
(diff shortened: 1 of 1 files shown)


dusvyat:
So I thought completion was being deprecated in latest openai python client but seems to still be supported which is strange - https://github.com/openai/openai-python/blob/main/src/openai/resources/completions.py

paxcema:
Interesting, yeah that's a great point. OpenAI's marked the endpoint as legacy and they discontinued their models, but perhaps other providers will not, and in that case we may still want to use this bit of code.

What about simply changing the default mode, then? And we keep the relevant conditional instead of removing it.

dusvyat:
> Interesting, yeah that's a great point. OpenAI's marked the endpoint as legacy and they discontinued their models, but perhaps other providers will not, and in that case we may still want to use this bit of code.
> 
> What about simply changing the default mode, then? And we keep the relevant conditional instead of removing it.

Good idea, that sounds like a fair trade off

paxcema:
Closing for now.

PR #8720, opened 2024-02-01
clean up handler install log messages

hamishfagg:
Gives the user a message with instructions for installing handlers in docker if they're running in docker, and pip instructions otherwise


ea-rus (approved):

mindsdb/api/executor/command_executor.py
```diff
@@ -1048,7 +1048,7 @@ def _create_integration(self, name: str, engine: str, connection_args: dict):
             )
             handler_meta = handlers_meta[engine]
             if handler_meta.get("import", {}).get("success") is not True:
-                raise ExecutorException(f"Handler '{engine}' can not be used")
+                raise ExecutorException(f"The '{engine}' handler isn't installed.\n" + get_handler_install_message(engine))
```
StpMax, line 1051: Maybe it makes sense to add `get_handler_install_message(engine)` only if it is not a cloud env.

StpMax (approved):

PR #8652, opened 2024-01-18
Added dependencies to k8s, LLM fixes

ea-rus:
## Description

Added handlers to cloud requirements:
- llama_index
- writer
- rag
  - added sentence-transformers for rag

Fixes llama handler error (presumably added by #8562):
- DFReader is default reader
- wrong usage of get_api_key
- fix openai deprecation error
- remove global api key binding

Fixes: [#346](https://github.com/mindsdb/mindsdb_private/issues/346)
Fixes:  [#357](https://github.com/mindsdb/mindsdb_private/issues/357)
Fixes:  [#358](https://github.com/mindsdb/mindsdb_private/issues/358)
Fixes: https://github.com/mindsdb/mindsdb_private/issues/361

## Type of change

(Please delete options that are not relevant)

- [x] 🐛 Bug fix (non-breaking change which fixes an issue)

## Verification Process

To ensure the changes are working as expected:

 - [ ]   Test Location: Specify the URL or path for testing.
 - [ ]   Verification Steps: Outline the steps or queries needed to validate the change. Include any data, configurations, or actions required to reproduce or see the new functionality.

## Additional Media:

- [ ] I have attached a brief loom video or screenshots showcasing the new functionality or change.

## Checklist:

- [ ] My code follows the style guidelines(PEP 8) of MindsDB.
- [ ] I have appropriately commented on my code, especially in complex areas.
- [ ] Necessary documentation updates are either made or tracked in issues.
- [ ] Relevant unit and integration tests are updated or added.

PR #8733, opened 2024-02-05
Hotfix : llama index handler 

pranavvp16:


martyna-mindsdb:
@paxcema 
Updated to llama_index re get_api_key() function were made in this PR: https://github.com/mindsdb/mindsdb/pull/8245

So these PRs should be aligned.

paxcema:
That's a good point @martyna-mindsdb, looked into it and the hotfix is no longer required. Fixes were introduced and merged in #8652, so I'm closing this.

PR #8759, opened 2024-02-08
[ENH] Code and contextual QA formatters for LLM finetuning

paxcema:
## Description

Fixes mindsdb_private#394

## Type of change

## Verification Process

 - [x]  Added unit tests

## Checklist:


pedrofluxa (approved):
No comments. Clean implementation with lots of room to add more features. Good job!!

If anything, I'd add some heavy unit testing to make it absolutely bullet-proof. That is something we will be building shortly as we start using the codebase so approving the PR as-is is perfectly acceptable.

requirements/requirements.txt
```diff
@@ -35,4 +35,5 @@ scikit-learn==1.3.2
 protobuf==3.20.3
 hierarchicalforecast~=0.4.0
 google-auth-oauthlib
-msal
\ No newline at end of file
+msal
+langchain==0.1.5
```
paxcema, line 39: @hamishfagg could you expand a bit on what skill this introduced torch as a dependency? Ref: #8711. 

I'm looking to bump langchain's version on a PR that would immediately follow up this one upon merging, and would like to avoid re-introducing the size increase in our docker images. I think the separation of langchain since 0.0.347 into `langchain, langchain-core, langchain-experimental, langchain-community` may help us here, but I'm not sure what the original culprit was, any info. on that would be appreciated!

dusvyat (approved):
Sorry for the delay. All looks good

requirements/requirements.txt
hamishfagg: Yeah sure,
`skill_tool.py` required the rag handler, which requires `sentence-transformers`, which requires `torch`.
So I think as long as we don't end up having to include the rag handler by default, we shouldn't introduce the same issue.

PR #8766, opened 2024-02-13
Stored MS Teams Redirect URI in Handler Storage

MinuraPunchihewa:
## Description

This PR updates the logic for the MS Teams authentication process to generate a redirect URI only on the first time that the connection initialized. At this point the redirect URI will be stored in handler storage and subsequent attempts at initializing the same connection will access it from this storage.

This update was required because as the chat bot task runs on a separate thread and does not come through a request, it does not have access to the request headers used to generate the redirect URI.

## Type of change

## Verification Process

 - [x]   Verification Steps: This change needs to be validated by creating a MS Teams chat bot.

## Checklist:

- [X] Necessary documentation updates are either made or tracked in issues - no updates required.
- [X] Relevant unit and integration tests are updated or added -  no updates required.


QuantumPlumber (approved):
Looks good to me and works on my local dev instance

PR #8795, opened 2024-02-21
Added Support for Summarization to the Twelve Labs Integration

MinuraPunchihewa:
## Description

This PR adds the ability to run summarization tasks using the Twelve Labs integration. The queries given below are an example of how this task can be executed.

First, run a `CREATE MODEL` statement where `task` is set to `summarization` (other task specific parameters have been outlined in the README),
```
CREATE MODEL mindsdb.twl_model
PREDICT result
USING
  engine = 'twelve_labs_engine_1',
  api_key = '<api_key>',
  engine_id = 'pegasus1',
  index_name = 'indxx_1',
  index_options = ['visual', 'conversation'],
  video_files = ['/home/minura/Videos/twelve_labs_data/test.mp4'],
  task = "summarization",
  summarization_type = 'summary'
```

Now, running `SELECT` on this model requires a video_id to be specified. In order to extract the IDs of the videos that were indexed, a DESCRIBE command will need to be executed,
`DESCRIBE twl_model.indexed_videos`

A SELECT statement can now be running using these IDs,
```
SELECT * FROM twl_model
WHERE video_id = '<video_id>'
```

## Type of change

## Verification Process

## Additional Media:

## Checklist:


MinuraPunchihewa:
@ZoranPandovski There are a couple of limitations/notes that I thought I would mention here,
1. When video URLs are used to run `CREATE MODEL` statements, there is no way, at the moment, to identify which video each video ID refers to. In the image given below, the first video was specified using the `video_urls` parameter, while the second was specified using `video_files` in a separate `CREATE MODEL` statement; when a path to the video is provided, the `filename` column is populated, but this is not so when using URLs,
 
Screenshot from 2024-02-21 16-38-28

2. In this implementation, when running `SELECT` on summarization models, the column name expected in the `WHERE` clause is `video_id`. Should we make it so that the name of this column can be specified as a parameter when running `CREATE MODEL`? When running search, we expect a `query_column` parameter.

3. Should we change the `query_column` parameter mentioned above to `search_query_column`? Just so that there is a clear distinction (in terms of prefixes) between parameters used for search, summarization and any other task we hope to support in the future?

ZoranPandovski:
@MinuraPunchihewa I just tried using the TwelveLabs API and it looks like there is an UPDATE endpoint that provides an option for setting up the `video_title` inside metadata object. I think once we create the index, we can set this up for each video in the index. Then, the DESCRIBE will return this alongside the ids. Can you try that?

MinuraPunchihewa:
@ZoranPandovski `DESCRIBE` will now return a column called video_reference,
image

Note: I tried adding `video_title` at first, but discovered that it is not returned when listing videos for some reason (only when getting individual ones).

MinuraPunchihewa:
@ZoranPandovski I've done some testing and it looks like all summarization types are working well,

Summary:
image

Chapter:
image

Highlight:
image

mindsdb/integrations/handlers/twelve_labs_handler/README.md
```diff
@@ -71,4 +87,18 @@ WHERE query = 'search query';
...
+SELECT *
+FROM mindsdb.twelve_labs_summarization
+WHERE video_id = 'video_1';
+```
+
+Here, the video IDs that were indexed by a model can be found by running a `DESCRIBE` statement on the it. The URL or file path of the video will be available in the `video_reference` column. The following is an example of how to run such a `DESCRIBE` statement,
+```sql
+DESCRIBE mindsdb.twelve_labs_summarization.indexed_videos;
+```
+
```
ZoranPandovski, line 103: Let's include a response of the DESCRIBE as an example here

mindsdb/integrations/handlers/twelve_labs_handler/settings.py
```diff
@@ -90,6 +94,33 @@ def check_param_typos(cls, values):
 
         return values
 
+    @root_validator(pre=True, allow_reuse=True, skip_on_failure=True)
+    def check_for_valid_engine_options(cls, values):
```
ZoranPandovski, line 98: Can we do the same for task: check for => summarization or search

ZoranPandovski: Or this is already handled by the type check of the task

mindsdb/integrations/handlers/twelve_labs_handler/settings.py
```diff
@@ -182,7 +226,7 @@ class TwelveLabsHandlerConfig(BaseSettings):
         Default wait duration when polling video indexing tasks created via the Twelve Labs API.
     """
 
-    BASE_URL = "https://api.twelvelabs.io/v1.1"
+    BASE_URL = "https://api.twelvelabs.io/v1.2"
```
ZoranPandovski, line 229: There are some scenarions where users need control over this. Maybe we set it up as optional parameter and this will be default value

mindsdb/integrations/handlers/twelve_labs_handler/twelve_labs_api_client.py
```diff
@@ -100,6 +104,77 @@ def get_index_by_name(self, index_name: str) -> str:
...
+        result = self._submit_request(
+            method="GET",
+            endpoint=f"indexes/{index_id}/videos",
+        )
+        data.extend(result['data'])
+
+        while result['page_info']['page'] < result['page_info']['total_page']:
+            result = self._submit_request(
+                method="GET",
+                endpoint=f"indexes/{index_id}/videos?page_token={result['page_info']['next_page_token']}",
```
ZoranPandovski, line 134: Maybe we can also add the page_limit if there are more items

ZoranPandovski (changes requested):
Looks great, I've added few minor comments

mindsdb/integrations/handlers/twelve_labs_handler/README.md
MinuraPunchihewa: Done.

mindsdb/integrations/handlers/twelve_labs_handler/settings.py
MinuraPunchihewa: This was the handled within the handler, but it makes more sense to move it here, so I added another task for it.

mindsdb/integrations/handlers/twelve_labs_handler/settings.py
MinuraPunchihewa: Done.

MinuraPunchihewa:
@ZoranPandovski Do you think the `query_column` parameter should be changed to `search_query_column`? Just to keep things consistent (and mark it as a search specific parameter) or you don't think this is necessary?

ZoranPandovski:
Sure, lets do it and we can merge this @MinuraPunchihewa

MinuraPunchihewa:
> Sure, lets do it and we can merge this @MinuraPunchihewa

Done.

ZoranPandovski (approved):

PR #8885, opened 2024-03-05
anthropic to use messages api (instead of legacy completions api)

martyna-mindsdb:
## Description

updated anthropic handler to use messages API (instead of completions API which is a legacy API now)

Fixes #issue_number

## Type of change


paxcema (approved):
LGTM

paxcema:
Oh, it seems like Claude 3 needs to use the Messages API, which is not the one this integration uses right now (Completions). The change is not so simple, then. Do you want to take a stab at it? Otherwise, I can try to help.

martyna-mindsdb:
@paxcema 
The support for claude-3 models is in place now.

The output provided by Messages API comes in the format as below. Shall we keep it this way or just extract text from it?

image

image

image

paxcema (changes requested):
I think we should extract text from it.

It seems there can be other types of content (e.g. images) if the model is multimodal. I would recommend to check the type is text, and if so, then extract the actual content inside.

martyna-mindsdb:
@paxcema 
I updated the code and tested for all Anthropic models. Now the output contains only the text value for type==text.
Please review.

paxcema (approved):

PR #8902, opened 2024-03-11
Clean up github actions

hamishfagg:
## Description

- Organise testing jobs by when they should be executed. This brings testing in line with, and divides up testing according to [our software development lifecycle](https://github.com/mindsdb/internal-documentation/blob/master/sdlc.md). All tests have been moved into:
  - `test_on_push.yml`: to be run whenever new code is pushed to a PR
  - `test_on_deploy.yml`: to be run whenever code is deployed to an environment (ideally these would be run against the deployed env - currently they are not). This is triggered by the `build_deploy_dev` and `build_deploy_staging` workflows and not run directly itself.
- Uses `matrix_include.json` to define the matrix of python versions/OSs to run tests on, for all testing jobs.


.github/workflows/test_on_deploy.yml
```diff
@@ -0,0 +1,112 @@
...
+          pytest -vx tests/integration_tests/flows/test_home_rental_prices_tutorial.py
+
+          echo -e "\n===============Forecast Quaterly House Sales===============\n"
+          pytest -vx tests/integration_tests/flows/test_forecast_quaterly_house_tutorial.py
+
+          echo -e "\n===============Predict Text Sentiment Wih Huggingface===============\n"
+          pytest -vx tests/integration_tests/flows/test_predict_text_sentiment_huggingface_tutorial.py
+
+          echo -e "\n===============Predict Text Sentiment Wih OpenAI===============\n"
+          pytest -vx tests/integration_tests/flows/test_predict_text_sentiment_openai_tutorial.py
```
lucas-koontz, line 108: Nitpick, and feel free to ignore it. 
I prefer having the list of tests in a makefile, which I can run before submitting the commit.

Other than that, LGTM.

lucas-koontz (approved):

.github/workflows/test_on_deploy.yml
hamishfagg: Ive created #8957 to do this later

PR #8905, opened 2024-03-12
Fix supported python versions for pip installs

hamishfagg:
## Description

Won't allow pip installs on python versions we don't support (3.11+)

## Type of change


mindsdb-devops (approved):
lgtm

PR #8911, opened 2024-03-12
Updated to the GET /handlers Endpoint to Support Forms in the GUI

MinuraPunchihewa:
## Description

At the moment, through our GUI, it is not possible to initialize all of our integrations through the forms section. The reason for this is because certain information that is required for the form is maintained as constants [here](https://github.com/mindsdb/mindsdb-frontend/blob/main/src/utils/content/databases.js#L84C9-L84C14) in the front-end repository.

It would be more robust and extensible if this information can be delivered through our API endpoints and the objective of this PR is to accomplish that. This is done by introducing the following new attributes to our existing `GET /handlers` endpoint,
- `most_popular`: determined by querying the top 10 data sources and AI engines as per our data warehouse in BigQuery.
- `is_aws`: determined checking certain keywords against the title of the integration.
- `is_new`: determined by checking the creation/modification date of the __about__.py file of each integration.

Is this a good way to go about it? The additional computation will require a little more time for this endpoint to return a response. Should this maybe be separated out to a new endpoint?

## Type of change

## Verification Process

## Checklist:


martyna-mindsdb:
@MinuraPunchihewa 
Is it still valid?

MinuraPunchihewa:
@martyna-mindsdb This was something I was exploring on the side; to basically display all of the integrations that we support on GUI. 
I think it is worth leaving this in draft to pick up later.

StpMax:
i deleted myself from reviewers to not show the PR in awaiting list while it is not ready

PR #8938, opened 2024-03-18
updated docker desktop docs

martyna-mindsdb:
## Description

updated docker desktop docs

Fixes #issue_number

## Type of change

### DIFF
--- a/docs/setup/self-hosted/docker-desktop.mdx
+++ b/docs/setup/self-hosted/docker-desktop.mdx
@@ -1,26 +1,10 @@
 ---
-title: Setup as a Docker Extension
-sidebarTitle: Docker Extension
+title: Docker Desktop Extension for MindsDB
+sidebarTitle: Docker Desktop
 ---
 
-This is work in progress but [Ajeet Singh Raina](https://github.com/ajeetraina) has built a great docker extension, should look like this
+MindsDB provides an extension for Docker Desktop that facilitates running MindsDB on Docker Desktop.
 
-<p align="center">
-  <img src="https://user-images.githubusercontent.com/34368930/230012152-c60bc8a8-3c00-48c9-961c-6d10421e7e26.png"/>
-</p>
-
-
-
-<Info>
-## Install
-Follow the instructions here if you want to try it:
-
-<CardGroup cols={1}>
-
-    <Card title="MindsDB Docker extension guide" icon="link" href="https://github.com/collabnix/mindsdb-docker-extension"></Card>
-
-
-</CardGroup>
-
-</Info>
-
+<Tip>
+Visit the [GitHub repository for MinsdDB Docker Desktop Extension](https://github.com/mindsdb/mindsdb-docker-extension) to learn more.
+</Tip>


MinuraPunchihewa (commented):
@martyna-mindsdb Do you think it will be useful for us to include some information on how extension can be installed on Docker Desktop?
E.g. Go to the Extensions page (marketplace) -> Search for MindsDB -> Install the extension etc.?

docs/setup/self-hosted/docker.mdx
```diff
@@ -86,44 +68,63 @@ Now you can access the MindsDB editor by going to `127.0.0.1:47334` in your brow
...
+docker run -e MKL_SERVICE_FORCE_INTEL=1 -p 47334:47334 -p 47335:47335 mindsdb/mindsdb
+```
+</Warning>
+
+### Install dependencies
 
 MindsDB integrates with numerous data sources and AI frameworks. To use any of the integrations, you should enure that the required dependencies are installed in the Docker container.
 
-Here is how to install the dependencies in the Docker container:
+<Info>
```
MinuraPunchihewa, line 83: Should the 'Manage Integrations' page be pointed out as the main method for installing dependencies? And should the other methods (via the editor and the Add 'Datasource' page) be included as well?
@ZoranPandovski What do you think?

docs/setup/self-hosted/docker.mdx
```diff
@@ -86,44 +68,63 @@ Now you can access the MindsDB editor by going to `127.0.0.1:47334` in your brow
...
+
+```bash
+exit
+```
 
-### Default Configuration
+Restart the container:
+
+```bash
+docker restart container-name
```
MinuraPunchihewa, line 120: If we are going to include this, do you think it will be important to point out how the name of the container can be found? If the name is not specified in the `docker run` command, Docker automatically assigns a random name to the container, therefore, users will need to run a `docker ps` command to find out what the name assigned to their container is.

docs/setup/self-hosted/docker-desktop.mdx
```diff
@@ -1,26 +1,38 @@
...
+Visit the [GitHub repository for MinsdDB Docker Desktop Extension](https://github.com/mindsdb/mindsdb-docker-extension) to learn more.
+</Tip>
+
+## Prerequisites
+
+Before proceeding, ensure you have installed Docker Desktop, following the [official Docker Desktop documentation](https://www.docker.com/products/docker-desktop/).
 
+## Setup
 
+This setup of MindsDB uses one of the available Docker images, as [listed here](/setup/self-hosted/docker).
```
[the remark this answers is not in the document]
martyna-mindsdb: Thanks Minura. I corrected it.

docs/setup/self-hosted/docker.mdx
martyna-mindsdb: Installing dependencies is part of the setup - therefore, I'd keep it here.

As of now, I mentioned about the 'Manage Integrations' page as the main method, as other methods blend in the flow. But if you think, we can mention it as well.

Btw. Do you think we should get rid of the method presented below that requires a container to be restarted? Or just keep it as an alternative.

docs/setup/self-hosted/docker.mdx
martyna-mindsdb: I modified `docker run` and added info about `docker ps`.

docs/setup/self-hosted/docker.mdx
```diff
@@ -49,35 +18,54 @@ It is the Docker image of MindsDB that comes with [these integrations](https://g
...
+-   `docker run` is a native Docker command used to spin up a container.
+-   `--name mindsdb_container` defines a name for the container.
+-   `-p 47334:47334` publishes port 47334 to access MindsDB GUI and HTTP API.
+-   `-p 47335:47335` publishes port 47335 to access MindsDB MySQL API.
+-   `mindsdb/mindsdb` is a Docker image provided by MindsDB. You can choose a different one from the list above.
+
+Once the container is created, you can use the following commands:
+
+-   `docker stop mindsdb_container` to stop the container. Note that this may not always be necessary because when turning off the host machine, the container will also be shut down.
+-   `docker start mindsdb_container` to start the container back up.
```
MinuraPunchihewa, line 50: I think it might be important to mention here that this command is used to restart a stopped container with all its previous changes (such as any dependencies that are installed) intact.
I am not sure if it is necessary to mention that using the above `docker run` again when the previous container is stopped will start an entirely new container. What do you think?

docs/setup/self-hosted/docker.mdx
MinuraPunchihewa: Yes, that makes sense. Since the other methods are part of the flow, let's not mention it here.

I think it might be best to keep it as an alternative for two reasons,
1. Some of our users might already be accustomed to this method.
2. Certain advanced users might actually prefer it (unlikely).

docs/setup/self-hosted/docker.mdx
```diff
@@ -86,44 +74,67 @@ Now you can access the MindsDB editor by going to `127.0.0.1:47334` in your brow
...
+```
+</Warning>
+
+### Install dependencies
 
 MindsDB integrates with numerous data sources and AI frameworks. To use any of the integrations, you should enure that the required dependencies are installed in the Docker container.
 
-Here is how to install the dependencies in the Docker container:
+<Info>
+Alternative way is to install dependencies directly from MindsDB editor. Go to *Settings* and *Manage Integrations*, select integrations you want to use and click on *Install*.
```
MinuraPunchihewa, line 90: This still says 'Alternative way'; should we change this section a little?

docs/setup/self-hosted/docker-desktop.mdx
```diff
@@ -1,26 +1,38 @@
...
+  <img src="https://user-images.githubusercontent.com/34368930/230012152-c60bc8a8-3c00-48c9-961c-6d10421e7e26.png"/>
+</p>
 
-</CardGroup>
+## Install dependencies
 
-</Info>
+In the MindsDB editor, go to *Settings* and *Manage Integrations*.
 
+Select integrations you want to use and click on *Install*.
```
MinuraPunchihewa, line 38: Should we maybe add a picture of the page here (and maybe down in the Docker doc) too? Just wondering, not mandatory.

MinuraPunchihewa (changes requested):

docs/setup/self-hosted/docker.mdx
martyna-mindsdb: Good point. I added more explanation to distinguish between docker start and docker run.

MinuraPunchihewa (commented):
@martyna-mindsdb Thank you, this is really great. I just have a couple more comments about the images that we have used here,
1. For the screenshot of the extension of the extension; can we please use [this](https://github.com/mindsdb/mindsdb-docker-extension/blob/main/mindsdb_screenshot.png)?
2. In the screenshot of the 'Manage Integrations' page; the table does not seem centered. Is that how it appears to you? Mine seems to be centered,

image

martyna-mindsdb:
@MinuraPunchihewa 
Thanks. I updated this image. As for the Manage Integratons image, I cropped it - now I updated it as well.

MinuraPunchihewa (approved):
Thanks, @martyna-mindsdb! I believe we are good to go.
Thank you so much for all your effort.

PR #8991, opened 2024-03-26
create prometheus dir if it doesnt exist

hamishfagg:
- Create prometheus dir if `PROMETHEUS_MULTIPROC_DIR` is set, but the dir doesnt exist
- Remove `PROMETHEUS_MULTIPROC_DIR` from the dockerfile as the average user won't use prometheus
- Make log message `info` as it is not an issue for normal users.
- Remove `mkdir prometheus_metrics` from test script

### DIFF
--- a/.github/workflows/test_on_deploy.yml
+++ b/.github/workflows/test_on_deploy.yml
@@ -52,8 +52,6 @@ jobs:
 
         if [ "$RUNNER_OS" == "Linux" ]; then
 
-          mkdir prometheus_metrics
-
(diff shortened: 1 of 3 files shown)


MinuraPunchihewa (approved):
I have tested this locally and it works great!
I was able to run the container with the env variable set as follows,
`sudo docker run --name mdb -p 47334:47334 -p 47335:47335 -e PROMETHEUS_MULTIPROC_DIR='prometheus_metrics' mindsdb`

PR #8999, opened 2024-03-27
Docker build improvements

hamishfagg:
Currently we build each platform's docker image and push cache layers to a repo separately. This is because buildkit doesn't support building and pushing cache for multiple platforms at once.

The issue with this approach is that we need to push images suffixed with `-amd64` and `-arm64` to dockerhub before building a manifest that merges those two images. This is messy and confusing for users, and triples the number of tags we have in our dockerhub repo.

This PR uses another approach to overcome buildkit's shortcoming
- Build all images for all platforms and push the images but *not* cache
- Build each platform's images again (should now be cached) and don't push the images, but push cache instead

This approach means we can also continue with other jobs without waiting for cache to be pushed (which can take a while).

### DIFF
--- a/docker/docker-bake.hcl
+++ b/docker/docker-bake.hcl
@@ -13,31 +13,35 @@ variable "IMAGE" {
 variable "VERSION" {
   default = "unknown"
 }
-variable "PLATFORM" {
-  default = "linux/amd64"
+variable "PLATFORMS" {
+  default = "linux/amd64,linux/arm64"
+}
+variable PLATFORM_LIST {
+  default = split(",", PLATFORMS)
 }
 variable "BRANCH" {
   default = "stable"
 }
-
-function "get_platform_tag" {
-  params = []
-  result = replace("${equal(PLATFORM, "") ? "" : "-"}${PLATFORM}", "linux/", "")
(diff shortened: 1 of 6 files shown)


docker/docker-bake.hcl
```diff
@@ -13,31 +13,35 @@ variable "IMAGE" {
...
 }
 variable "BRANCH" {
   default = "stable"
 }
-
-function "get_platform_tag" {
-  params = []
-  result = replace("${equal(PLATFORM, "") ? "" : "-"}${PLATFORM}", "linux/", "")
+variable "ECR_REPO" {
+  default = "454861456664.dkr.ecr.us-east-2.amazonaws.com"
```
lucas-koontz, line 26: Is it ok to publish this info in this repository?

lucas-koontz: NVM, it was already in the code.

lucas-koontz (approved):
LGTM!

docker/docker-bake.hcl
hamishfagg: Good catch but yeah, I checked :)

PR #9001, opened 2024-03-27
Schema refactor

ea-rus:
## Description

Changes:

**`information_schema` database**

Created `information_schema` database 
- Tables visible in GUI:  MODELS, DATABASES, ML_ENGINES, HANDLERS, JOBS, TRIGGERS, CHATBOTS, KNOWLEDGE_BASES, SKILLS, AGENTS, VIEWS
- Also information_schema has system tables which are not visible in GUI. But it is possible to select from them because they used by database tools to get database objects, for example:  TABLES, SCHEMATA, PLUGINS, ...

image

**Syntax update**

SHOW:
- [`SHOW PROJECTS` support ](https://github.com/mindsdb/mindsdb/issues/5696)
- `SHOW <object>` displays objects only from current project
- To show from other project: `SHOW <object> FROM <project name>`
- Also it is possible to filter output:
  - filter by condition `show models where engine='openai'`
  - filter by name `show models like 'home%'`

DESCRIBE:
- Implemented `DESCRIBE <object_type> <object_name>` command.
  -  It works the same as query: `select * from informational_schema.<object_type> where name=<object_name>`

Update the active model version:
- was: `update models_versions set active=1 where version=11 and name='task_model' `
- will be: `Set active task_model.11`

Remove the model version:
- was: `delete from models_versions where version=2 and name='task_model'`
- will be: `Drop model task_model.2`

**Changes in project objects**

Tables removed from projects (moved to information_schema):
- models and models_versions: combined into one table `information_schema.MODELS`
  - it has 'version' column and shows all models versions
- mdb_triggers. Moved to information_schema.TRIGGERS
  - 'select * from information_schema.triggers' displays different set of columns for sql and http server
     in mysql server columns are extended by [system columns](https://github.com/mindsdb/mindsdb/blob/253cd5a9d82fb71605a934e8dfb65746e7f28503/mindsdb/api/executor/datahub/datanodes/mindsdb_tables.py#L225) required by db tools
- jobs and chatbots also removed

For back compatibility: it is possible to select from these tables in project (for example `select * from mindsdb.models`) but this feature be removed in next updates.

Added to projects:
- Agents are visible in project, because they behave like a model

In action
https://www.loom.com/share/5ac2ce8872384293ba4ea48bc09c2081

Dependent on https://github.com/mindsdb/mindsdb_sql/pull/371

## Type of change

## Verification Process

## Additional Media:

## Checklist:


mindsdb/api/executor/datahub/datanodes/mindsdb_tables.py
```diff
@@ -0,0 +1,418 @@
...
+from mindsdb_sql.parser.ast.base import ASTNode
+
+from mindsdb.interfaces.agents.agents_controller import AgentsController
+from mindsdb.interfaces.jobs.jobs_controller import JobsController
+from mindsdb.interfaces.skills.skills_controller import SkillsController
+from mindsdb.interfaces.database.views import ViewController
+from mindsdb.interfaces.database.projects import ProjectController
+
+
+from .system_tables import Table
```
StpMax, line 14: It is better to use absolute imports where possible

mindsdb/api/executor/datahub/datanodes/mindsdb_tables.py
```diff
@@ -0,0 +1,418 @@
...
+            and query.where.op == "="
+            and query.where.args[0].parts == ["project"]
+            and isinstance(query.where.args[1], Constant)
+        ):
+            project_name = query.where.args[1].value
+
+        data = triggers_controller.get_list(project_name)
+
+        columns = self.mindsdb_columns
+        if inf_schema.session.api_type == 'sql':
```
StpMax, line 270: Do we need `api_type` arg only for this specific case?
 Is it a good idea to return a different set of columns for each API?

mindsdb/api/executor/datahub/datanodes/information_schema_datanode.py
```diff
@@ -344,31 +57,7 @@ def __init__(self, session):
...
-            "KNOWLEDGE_BASES": self._get_knowledge_bases,
-            "SKILLS": self._get_skills,
-            "AGENTS": self._get_agents
-        }
-        for table_name in self.information_schema:
-            if table_name not in self.get_dataframe_funcs:
-                self.get_dataframe_funcs[table_name] = partial(
-                    self._get_empty_table, table_name
-                )
+        self.tables = {t.name: t() for t in self.tables_list}
```
StpMax, line 60: Looks like tables classes do not use `self`. May be make them static? so we do not need to instantiate them each time.

StpMax (changes requested):

StpMax:
`Set model_active = task_model.11`  looks strange. In mysql there is few `set` commands without '=', like 
`SET CHARACTER SET 'name'`
or
`SET DEFAULT ROLE admin TO user`
May use use something like this?
`SET MODEL task_model ACTIVE VERSION TO 11 `

mindsdb/api/executor/datahub/datanodes/mindsdb_tables.py
ea-rus: yes, only for this case. 
we use the whole set columns for db managers (like dbeaver) and limited set for web GUI, to make able user see only essential information

Also division between system_tables and mindsdb_tables - it is only for SQL server, to make sql clients working. In GUI we show only mindsb tables (though allow to select from all of them)

ea-rus:
`Set model_active = task_model.11`
changed to 
`Set active task_model.11`

mindsdb/api/executor/datahub/datanodes/information_schema_datanode.py
ea-rus: changed to class_methods

StpMax (approved):

PR #9002, opened 2024-03-27
Remove unused files

ZoranPandovski:
This PR removes unused:
*  protos directory
* quickstart guide in docs 
*  1.35.0 file
* empty api/nlp dir 
* integrations net_helpers
* all files in microservices_grpc
* grpc handlers wrappers
* grpc db handlers clients
* helm charts unmaintained directory

One of the biggest changes was changing the `create_tmp_handler` method so as not to use the DBClient dependent on the `grpc`.
@StpMax : Integration status check:
Screenshot from 2024-05-07 16-37-38
Using the wrong username:
Screenshot from 2024-05-07 16-51-25
Successful select after creating new integration:
Screenshot from 2024-05-07 17-49-58


ZoranPandovski:
@StpMax , @ea-rus can you please double check this?

mindsdb/interfaces/database/integrations.py
```diff
@@ -411,41 +399,6 @@ def _make_handler_args(self, name: str, handler_type: str, connection_data: dict
 
         return handler_ars
 
-    def create_tmp_handler(self, handler_type: str, connection_data: dict) -> object:
```
ea-rus, line 414: this function is used, for example [here](https://github.com/mindsdb/mindsdb/blob/11b34a7eac44cd8c50754a6895ee1e1db74bf3d8/mindsdb/api/http/namespaces/config.py#L151)

ZoranPandovski: This is only for the tests. Do we need this since it brings a lot of complexity with it, which is not required now?

StpMax: @ZoranPandovski  yes, 'test connection' does not work anymore 
image

StpMax: > which is not required now?

@ZoranPandovski  you mean not required in new cloud? but what about opensource version - should it have 'test connection' ability?

ZoranPandovski: @StpMax , @ea-rus I've changed the `create_tmp_handler` class to not use the `DBClient`, which was quite complex, and used grpc dependencies.

StpMax (approved):

PR #9072, opened 2024-04-15
Remove custom GH actions and use shared ones

hamishfagg:
Removes all GH actions in this repo and uses shared actions in: https://github.com/mindsdb/github-actions/

Fixes some warnings from GH actions too


mindsdb-devops (approved):
lgtm

PR #9120, opened 2024-04-24
Fix mysql connection status tracking

hamishfagg:
We have an issue right now with mysql handler tests failing: https://github.com/mindsdb/mindsdb/actions/runs/8794637442/job/24175332978

It seems like something has changed in either the mysql connector, our mariadb container, or somewhere else that causes disconnection after some queries.
This wouldn't be an issue, as the connector is built to reconnect for queries - except our tracking of the connection status with `is_connected` is getting out of sync with reality and causing re-connection to be skipped.

This PR removes our own tracking of connection status and defers that job to the mysql connector library.


ZoranPandovski (approved):
Maybe we also need to update the mariadb test to include MariaDBHandler even it just extends MySQL https://github.com/mindsdb/mindsdb/blob/staging/tests/handler_tests/test_mariadb_handler.py#L88

ea-rus (approved):

PR #9171, opened 2024-05-03
Added the Troubleshooting Guide for the MindsDB Inference Handler

MinuraPunchihewa:
## Description

This PR adds the troubleshooting guide for the MindsDB Inference handler as part of the steps for making it enterprise-ready.

## Type of change

- [x] 📄 This change is a documentation update

## Verification Process

## Checklist:

### DIFF
--- a/docs/integrations/ai-engines/mindsdb_inference.mdx
+++ b/docs/integrations/ai-engines/mindsdb_inference.mdx
@@ -71,3 +71,27 @@ Here is the output:
 ## Supported Models
 
 For an overview of the models supported, visit the [following docs](https://docs.mdb.ai/). This list will help you quickly identify the right models for your needs.
+
+## Troubleshooting Guide
+
+<Warning>
+`Authentication Error`
+
+* **Symptoms**: Failure to authenticate to MindsDB Inference Endpoints.
+* **Checklist**:
+    1. Make sure the your MindsDB Inference Endpoints account is active.
+    2. Confirm that your API key is correct.
+    3. Ensure that your API key has not been revoked.
+</Warning>
+
+<Warning>
+`SQL statement cannot be parsed by mindsdb_sql`
+
+* **Symptoms**: SQL queries failing or not recognizing table and model names containing spaces or special characters.
+* **Checklist**:
+    1. Ensure table names with spaces or special characters are enclosed in backticks.
+    2. Examples:
+        * Incorrect: SELECT * FROM integration.travel data
+        * Incorrect: SELECT * FROM integration.'travel data'
+        * Correct: SELECT * FROM integration.\`travel data\`
+</Warning>
--- a/docs/integrations/ai-engines/mindsdb_inference.mdx
+++ b/docs/integrations/ai-engines/mindsdb_inference.mdx
@@ -90,8 +90,23 @@ For an overview of the models supported, visit the [following docs](https://docs
 * **Symptoms**: SQL queries failing or not recognizing table and model names containing spaces or special characters.
 * **Checklist**:
     1. Ensure table names with spaces or special characters are enclosed in backticks.
-    2. Examples:
-        * Incorrect: SELECT * FROM integration.travel data
-        * Incorrect: SELECT * FROM integration.'travel data'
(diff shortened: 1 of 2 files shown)


docs/integrations/ai-engines/mindsdb_inference.mdx
```diff
@@ -71,3 +71,27 @@ Here is the output:
...
 For an overview of the models supported, visit the [following docs](https://docs.mdb.ai/). This list will help you quickly identify the right models for your needs.
+
+## Troubleshooting Guide
+
+<Warning>
+`Authentication Error`
+
+* **Symptoms**: Failure to authenticate to MindsDB Inference Endpoints.
+* **Checklist**:
+    1. Make sure the your MindsDB Inference Endpoints account is active.
```
martyna-mindsdb, line 82: A small typo in this line - `Make sure **that** your...`

mindsdb/integrations/handlers/mindsdb_inference/README.md
```diff
@@ -71,3 +71,27 @@ Here is the output:
...
 For an overview of the models supported, visit the [following docs](https://docs.mdb.ai/). This list will help you quickly identify the right models for your needs.
+
+## Troubleshooting Guide
+
+<Warning>
+`Authentication Error`
+
+* **Symptoms**: Failure to authenticate to MindsDB Inference Endpoints.
+* **Checklist**:
+    1. Make sure the your MindsDB Inference Endpoints account is active.
```
martyna-mindsdb, line 82: As above:
A small typo in this line - `Make sure **that** your...`

MinuraPunchihewa:
@martyna-mindsdb I have fixed the typos. By the way, can you think of anything else that needs to be included here, Martyna?

martyna-mindsdb:
@MinuraPunchihewa 
In fact, I haven't encountered any issue with MindsDB Inference, except for the authentication error, which you already listed.

As for the `SQL queries failing...` issue, I'm not sure if it is relevant here, because the way we use the models via MindsDB Inference is by sending a message/prompt to the model (not a query). Am I missing something?

MinuraPunchihewa:
> @MinuraPunchihewa In fact, I haven't encountered any issues with MindsDB Inference, except for the authentication error, which you already listed.
> 
> As for the `SQL queries failing...` issue, I'm not sure if it is relevant here, because the way we use the models via MindsDB Inference is by sending a message/prompt to the model (not a query). Am I missing something?

Yes, you are right. The queries that can be run are quite similar to the OpenAI handler actually. So, it is possible to run queries like this:
```
SELECT output.sentiment, input.review
FROM example_db.demo_data.amazon_reviews AS input
JOIN mindsdb.sentiment_classifier AS output
LIMIT 3;
```
Running such queries can potentially lead to such errors, but the context is a little different. Shall I add different examples here?

martyna-mindsdb:
@MinuraPunchihewa 
That's right. One thing to break down - there are two ways of using MindsDB Inference:

1. Calling MindsDB Inference endpoints via Python SDK. This can result in the `authentication error`. There are no SQL queries being run here.
2. Connecting MindsDB Inference as an integration to MindsDB. Then, one can run SQL queries that may result in the second error listed.

Does that make sense? I think we can separate these two usage examples of MindsDB Inference and mention that in Troubleshooting.

MinuraPunchihewa:
> @MinuraPunchihewa That's right. One thing to break down - there are two ways of using MindsDB Inference:
> 
> 1. Calling MindsDB Inference endpoints via Python SDK. This can result in the `authentication error`. But, there are no SQL queries being run here.
> 2. Connecting MindsDB Inference as an integration to MindsDB. Then, one can run SQL queries that may result in the second error listed.
> 
> Does that make sense? I think we can separate these two usage examples of MindsDB Inference and mention that in Troubleshooting.

@martyna-mindsdb Do you think we should add the use of it via the Python SDK in this section of the docs? Since this is the documentation for integrations, is it relevant here?

martyna-mindsdb (approved):

martyna-mindsdb:
@MinuraPunchihewa 
This looks good now. Thank you.

ZoranPandovski (approved):

PR #9194, opened 2024-05-10
Improve caching of docker builds

hamishfagg:
This change makes a large portion of the docker build time (pip installing all our requirements) be able to be cached more often.

Cache misses were happening when almost any file changed in the repo, and most requirements would have to be installed again without being cached.

In this PR we instead:
- create a layer containing *only* the files required to run `pip install .` (requirements files, `setup.py` etc).
- Copy files from this layer and run `pip install .` and `pip install .[<extras>]` (this will now only miss if a requirements file or `setup.py` etc have changed)
- *Now* copy the whole repo into the container and run `pip install .` again - this will only install MindsDB itself
- Continue with the build as before

### DIFF
--- a/docker/mindsdb.Dockerfile
+++ b/docker/mindsdb.Dockerfile
@@ -1,7 +1,7 @@
 # This first stage produces a file structure with ONLY files required to run `pip install .`
 # We want to do this because we have a lot of dependencies which take a long time to install
 # So we avoid invalidating the cache of `pip install .` at all costs by ignoring every other file
-FROM python:3.10
+FROM python:3.10 as deps
 WORKDIR /mindsdb
 
 # Copy everything to begin with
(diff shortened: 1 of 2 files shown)


docker/mindsdb.Dockerfile
```diff
@@ -1,34 +1,55 @@
-# Bare mindsdb with no extras is built as a separate stage for caching
+# This first stage produces a file structure with ONLY files required to run `pip install .`
+# We want to do this because we have a lot of dependencies which take a long time to install
+# So we avoid invalidating the cache of `pip install .` at all costs by ignoring every other file
+FROM python:3.10
```
mindsdb-devops, line 4: Layer name missing

docker/mindsdb.Dockerfile
```diff
@@ -1,34 +1,55 @@
...
-WORKDIR /mindsdb
-
-# Copy just requirements and install them to cache the layer
-# This won't include any of the default handlers, but it should still speed things up
-COPY requirements/requirements.txt /mindsdb/requirements/requirements.txt
-RUN --mount=type=cache,target=/root/.cache/pip pip install -r requirements/requirements.txt
+    && apt-get install -y freetds-dev  # freetds required to build pymssql on arm64 for mssql_handler. Can be removed when we are on python3.11+
 
+# Copy the requirements files, setup.py etc from above
+COPY --from=0 /mindsdb .
```
mindsdb-devops, line 34: Layer name missing

mindsdb-devops (changes requested):
see comments, minor layer name missing

mindsdb-devops (approved):
lgtm

PR #9274, opened 2024-05-30
Mongo updates

ZoranPandovski:
## Description

This PR updates the mongo handler README file and removes the unnecessary certificate generation.

Fixes #https://github.com/mindsdb/mindsdb/security/code-scanning/15

## Type of change

## Additional Media:

- 
mongo

## Checklist:

### DIFF
--- a/mindsdb/integrations/handlers/mongodb_handler/mongodb_handler.py
+++ b/mindsdb/integrations/handlers/mongodb_handler/mongodb_handler.py
@@ -62,11 +62,6 @@ class MongoDBHandler(DatabaseHandler):
         if re.match(r'/?.*tls=false', self.host.lower()):
             kwargs['tls'] = False
 
-        if re.match(r'.*.mongodb.net', self.host.lower()) and kwargs.get('tls', None) is None:
-            kwargs['tlsCAFile'] = certifi.where()
-            if kwargs.get('tls', None) is None:
-                kwargs['tls'] = True
-
         connection = MongoClient(
             self.host,
             port=self.port,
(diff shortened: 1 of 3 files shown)


mindsdb/integrations/handlers/mongodb_handler/mongodb_handler.py
```diff
@@ -62,11 +62,6 @@ def connect(self):
         if re.match(r'/?.*tls=false', self.host.lower()):
             kwargs['tls'] = False
 
-        if re.match(r'.*.mongodb.net', self.host.lower()) and kwargs.get('tls', None) is None:
```
MinuraPunchihewa, line 65: @ZoranPandovski From what I have understood, this block of code was added to ensure secure connections to MongoDB Atlas. Is it OK to remove it?

I think we should also be able to solve the CodeQL alert by changing the regular expression as follows?
```
if re.match(r'.*\.mongodb.net', self.host.lower()) and kwargs.get('tls', None) is None:
    kwargs['tlsCAFile'] = certifi.where()
    if kwargs.get('tls', None) is None:
        kwargs['tls'] = True
```

By escaping the dot before `mongodb.net`, we can ensure that this dot matches a literal dot. This will not match other hosts as the CodeQL alert suggests. This is only necessary if keeping this block of code is required.

ZoranPandovski: Yes, this is not needed anymore. I've tested Atlas connections, and it works; you can check the screenshot above. Mongo Atlas provides certificate to establish the connection. You can check [pymongo docs](https://pymongo.readthedocs.io/en/stable/examples/tls.html#tls-ssl-and-pymongo).

MinuraPunchihewa: Understood. Thanks, Zoran.

MinuraPunchihewa (approved):

PR #9306, opened 2024-06-06
Dspy implementation

abhayasr:
## Description

Made a new pull request for draft dspy implementation


mindsdb/integrations/handlers/dspy_handler/constants.py
```diff
@@ -0,0 +1,91 @@
+SUPPORTED_PROVIDERS = {'openai', 'anthropic', 'anyscale', 'litellm', 'ollama'}
+# Chat models
+ANTHROPIC_CHAT_MODELS = {
```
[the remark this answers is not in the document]
abhayasr: as of right now, DSPy only works with openai, I just didn't want to delete this file, what do you suggest I change it to?

mindsdb/integrations/handlers/dspy_handler/mindsdb_chat_model.py
```diff
@@ -0,0 +1,181 @@
...
+            "name": message.name,
+        }
+    else:
+        raise ValueError(f"Got unknown type {message}")
+    if "name" in message.additional_kwargs:
+        message_dict["name"] = message.additional_kwargs["name"]
+    return message_dict
+
+
+class ChatMindsdb(BaseChatModel):
```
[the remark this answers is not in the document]
abhayasr: Yes, when I created the dspy handler folder I copied over all the files from langchain_handler, I just didn't know whether to delete it or not because this is not being used in dspy handler

mindsdb/integrations/handlers/dspy_handler/constants.py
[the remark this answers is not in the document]
abhayasr: I just redirected it to the one in langchain_hander, originally I had just copied all the files over but I will take care of this for this and all other comments now

mindsdb/integrations/handlers/dspy_handler/dspy_handler.py
```diff
@@ -0,0 +1,211 @@
...
+
+        # retrives llm and pass it around as context
+        model = args.get('model_name')
+        api_key = get_api_key('openai', args, self.engine_storage, strict=False)
+        # llm = dspy.OpenAI(model=model, api_key=api_key, api_base = 'https://llm.mdb.ai')
+        llm = dspy.OpenAI(model=model, api_key=api_key)
+
+        df = df.reset_index(drop=True)
+        if len(self.llm_data_controller.list_all_llm_data(1)) > 100:
+            for i in range(15):
```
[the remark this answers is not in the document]
abhayasr: I was using this as a way to make the df smaller because with too many examples dspy is super slow but I realized it is not needed because later in the code I am only taking the last 25 examples anyways

mindsdb/integrations/handlers/dspy_handler/dspy_handler.py
```diff
@@ -0,0 +1,211 @@
...
+            **kwargs):
+        super().__init__(model_storage, engine_storage, **kwargs)
+        # if True, the target column name does not have to be specified at creation time.
+        self.generative = True
+        self.log_callback_handler = log_callback_handler
+        self.llm_data_controller = LLMDataController()
+        if self.log_callback_handler is None:
+            self.log_callback_handler = LogCallbackHandler(logger)
+
+    def _get_llm_provider(self, args: Dict) -> str:
```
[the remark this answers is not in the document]
abhayasr: There are only 2 functions that are shared between the 2 and both require imports at the top of the file, should I still try to move it over to a new file or keep it for now because only 1 is exactly the same

mindsdb/interfaces/skills/skill_tool.py
```diff
@@ -26,6 +25,7 @@ def __init__(self):
     def get_command_executor(self):
         if self.command_executor is None:
             from mindsdb.api.executor.command_executor import ExecuteCommands
+            from mindsdb.api.executor.controllers import SessionController  # Top-level import produces circular import in some cases TODO: figure out a fix without losing runtime improvements (context: see #9304)  # noqa
```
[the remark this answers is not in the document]
abhayasr: I didn't even add this so not sure what this is but I can remove it

mindsdb/integrations/libs/llm/utils.py
```diff
@@ -165,7 +165,7 @@ def get_llm_config(provider: str, config: Dict) -> BaseLLMConfig:
         )
     if provider == 'mindsdb':
         return MindsdbConfig(
-            llm_model_name=config['model_name'],
+            model_name=config['model_name'],
```
paxcema, line 168: Out of curiosity, why are we renaming this parameter? I may be missing something, but it looks like we don't need to rename it.

mindsdb/integrations/libs/llm/config.py
```diff
@@ -85,5 +85,5 @@ class OllamaConfig(BaseLLMConfig):
 
 
 class MindsdbConfig(BaseLLMConfig):
-    llm_model_name: str
+    model_name: str
```
paxcema, line 88: Same comment as in `integrations/libs/llm/utils.py`

mindsdb/integrations/handlers/dspy_handler/requirements.txt
```diff
@@ -0,0 +1,11 @@
+openai==1.6.1
+wikipedia==1.4.0
+tiktoken==0.5.2
+anthropic>=0.26.1
+litellm==1.35.0
+chromadb # Knowledge bases.
+langchain-experimental==0.0.58
+dspy-ai
```
paxcema, line 8: As with other requirements, we probably want to pin `dspy` and `dspy-ai` to guarantee an update there won't suddenly break our integration.

mindsdb/interfaces/storage/db.py
```diff
@@ -551,3 +551,13 @@ class LLMLog(Base):
...
     total_tokens: int = Column(Integer, nullable=True)
     success: bool = Column(Boolean, nullable=False, default=True)
+
+
+class LLMData(Base):
+    __tablename__ = "llm_data"
+    id: int = Column(Integer, primary_key=True)
+    input: str = Column(String, nullable=False)
+    output: str = Column(String, nullable=False)
+    model_id: int = Column(Integer, nullable = False)
```
paxcema, line 561: Consider adding `model_name`, as well. We may want to log that.

mindsdb/interfaces/storage/db.py
```diff
@@ -551,3 +551,13 @@ class LLMLog(Base):
     completion_tokens: int = Column(Integer, nullable=True)
     total_tokens: int = Column(Integer, nullable=True)
     success: bool = Column(Boolean, nullable=False, default=True)
+
+
+class LLMData(Base):
```
paxcema, line 556: A docstring explaining what this table stores (and why) would be great!

mindsdb/interfaces/llm/llm_controller.py
```diff
@@ -0,0 +1,63 @@
...
+    def __init__(self) -> None:
+        '''Initializes the LLMDataController with configuration settings.'''
+        self.config = Config()
+
+    def add_llm_data(self, input_data: str, output_data: str, model_id: int) -> db.LLMData:
+        '''
+        Adds LLM input and output data to the database.
+        Parameters:
+            input_data (str): The input to the LLM.
+            output_data (str): The output from the LLM.
```
paxcema, line 17: Missing param in the docstring: `model_id`

mindsdb/interfaces/llm/llm_controller.py
```diff
@@ -0,0 +1,63 @@
...
+        Returns:
+            LLMData: The created LLMData object.
+        '''
+        # TODO: check the hash to avoid adding duplicate records (SQLAlchemy-based)
```
paxcema, line 21: This is fairly important to reason about. I think we _do_ want duplicate records at different timestamps, as they could indicate incorrect tool usage, for example. If you agree, we can remove this comment and keep implementation as-is, because the key in this table is `id`, so duplicate records are fine.

mindsdb/interfaces/llm/llm_controller.py
```diff
@@ -0,0 +1,63 @@
...
+    def list_all_llm_data(self, model_id: int):
+        '''
+        Lists all LLM data entries for a specific model.
+        Parameters:
+            model_id (int): The ID of the model/agent to filter the data.
+        Returns:
+            List[Dict]: A list of all LLMData objects in the database for the specified model.
+        '''
+        llm_data_objects = db.session.query(db.LLMData).filter_by(model_id=model_id).all()
+        return [{'input': llm_data.input, 'output': llm_data.output} for llm_data in llm_data_objects]
```
paxcema, line 53: Following the duplicates point above, I think we ought to return `created_at` as well.

mindsdb/interfaces/llm/llm_controller.py
```diff
@@ -0,0 +1,63 @@
...
+            llm_data_id (int): The ID of the LLM data to delete.
+        '''
+        llm_data = self.get_llm_data(llm_data_id)
+        if llm_data is None:
+            raise ValueError("LLM Data not found")
+
+        db.session.delete(llm_data)
+        db.session.commit()
+
+    def list_all_llm_data(self, model_id: int):
```
paxcema, line 44: nit: type hint for the returned object would be great!

mindsdb/integrations/handlers/dspy_handler/dspy_handler.py
```diff
@@ -0,0 +1,260 @@
...
+
+
+_PARSING_ERROR_PREFIX = 'An output parsing error occured'
+
+logger = log.getLogger(__name__)
+
+START_URL = 'http://20.102.90.50:2017/wiki17_abstracts'
+DEMOS = 4
+DF_EXAMPLES = 25
+MODEL_ID = 1
```
paxcema, line 33: I see we're not really updating this ID yet. At the OpenAI level there's quite a few model choices, so we probably want to have a mapping to ensure extensibility as we add support for more models. Perhaps the `model_id` attribute in the `db` should be an enum:

```python
import enum
from sqlalchemy import Enum

class DSPyModels(enum.Enum):
    gpt4 = 1
    gpt4o = 2
    # etc...

class LLMData(Base):
    model_id: int = Column(Enum, nullable = False)
```

Though there's also the problem of robustly mapping back to the API's string. @tmichaeldb would welcome your thoughts here!

As far as choices go, we should also make it quite explicit in the `README` that for now this supports a single model, for user awareness.

mindsdb/integrations/handlers/dspy_handler/dspy_handler.py
```diff
@@ -0,0 +1,260 @@
...
+            llm: OpenAI model
+
+        Returns:
+            df (DataFrame): Dataframe of responses
+        """
+        responses = []
+        for index, row in df.iterrows():
+            question = row[args['user_column']]
+            answer = self.generate_dspy_response(question, chain, llm)
+            responses.append({'answer': answer, 'question': question})  # TODO: check that columns are right here
```
paxcema, line 252: Comment here seems outdated, remove if so.

mindsdb/integrations/handlers/dspy_handler/dspy_handler.py
```diff
@@ -0,0 +1,260 @@
...
+            # TODO: check this only adds new incoming rows
```
paxcema, line 253: If we agree on duplicity (see above), this comment can be removed.

paxcema (changes requested):
There's also three empty files we should remove before merging: `assertion.log`, `openai_usage.log` and `azure_openai_usage.log`.

Forgot to mention, nice work! 🥳

mindsdb/integrations/libs/llm/utils.py
abhayasr: Tyler and I thought it would be easier to keep it consistent with the naming of the other models

mindsdb/integrations/handlers/dspy_handler/README.md
```diff
@@ -0,0 +1,74 @@
...
+      openai_api_key = 'api-key-value'];
+```
+
+<Tip>
+
+DSPy was created to optimize prompts to improve the performance of LLMs. You can read more about this in the [DSPy documentation](https://dspy-docs.vercel.app/docs/deep-dive/language_model_clients/remote_models/OpenAI).
+</Tip>
+
+
+Create a conversational model using `dspy_engine` as an engine and OpenAI as a model provider.  To get started, download and upload this file to the SQL editor: [download dspy_train](https://drive.google.com/file/d/1fztGxsIHTGWx6bgDShnFrzdcdagmeJp9/view?usp=sharing).  In order to upload the file, in the SQL editor click add and directly upload as a CSV and name it `dspy_train_dataset`.  This will serve as a cold start dataset for DSPy to improve on.
```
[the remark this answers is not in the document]
abhayasr: I can make the file publicly available, I just added it there for now so you and Tyler also had access to it, open to changing it to whatever form is most helpful

mindsdb/interfaces/llm/llm_controller.py
abhayasr: yes makes sense

mindsdb/interfaces/llm/llm_controller.py
abhayasr: Why do we need to distinguish between duplicates?

mindsdb/integrations/libs/llm/utils.py
paxcema: Got it, makes sense!

mindsdb/integrations/handlers/dspy_handler/README.md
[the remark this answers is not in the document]
paxcema: Ok, let's sync internally on how to make it easier for users to try out 👍

mindsdb/interfaces/llm/llm_controller.py
paxcema: Just in case time is a relevant dimension later on, for the model to self-improve or for us as architects. Right now it is a bit of future proofing. If you disagree, we can omit for now, as it's not a huge deal to come back and add it in.

mindsdb/integrations/handlers/dspy_handler/dspy_handler.py
abhayasr: done!

mindsdb/interfaces/llm/llm_controller.py
abhayasr: right makes sense, I will add it- only thing is will we want to keep this column when concatenating with the starting df because that will not have a created_at column?

mindsdb/integrations/handlers/dspy_handler/README.md
```diff
@@ -0,0 +1,74 @@
...
+
+```sql
+CREATE MODEL dspy_openai_model
+FROM files (SELECT input as question, output as answer FROM dspy_train_dataset)
+PREDICT answer
+USING
+     engine = 'dspy_engine',       -- engine name as created via CREATE ML_ENGINE
+     provider = 'openai',               -- one of the available providers
+     openai_api_key = 'api-key-value',  -- if not provided in CREATE ML_ENGINE
+     model_name = 'gpt-3.5-turbo',      -- choose one of the available OpenAI models
```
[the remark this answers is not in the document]
abhayasr: Do we not support all the OpenAI models?

PR #9342, opened 2024-06-13
Fixed LangChain Circular Import

MinuraPunchihewa:
## Description

This PR fixes a circular import in the LangChain integration that was making it unusable by moving it inside of the relevant function.  

Fixes https://github.com/mindsdb/mindsdb/issues/9333

## Type of change

## Verification Process

## Checklist:


ZoranPandovski (changes requested):
As far as I know this was an issue since importing the session every request, it initialize the command executor and takes more then 5s. Is there any other way to solve this>

MinuraPunchihewa:
Hmm.. I did notice that it takes a while for some statements to execute at times. I will keep looking for another way.

MinuraPunchihewa:
Hey @ZoranPandovski, @tmichaeldb,
I discovered that this circular import is caused by this [line](https://github.com/mindsdb/mindsdb/blob/2a5e61a7e994ddc06264665a4a03ccc21cd56c51/mindsdb/api/executor/controllers/session_controller.py#L16); the `IntegrationController` loads all handler modules via this [function](https://github.com/mindsdb/mindsdb/blob/2a5e61a7e994ddc06264665a4a03ccc21cd56c51/mindsdb/interfaces/database/integrations.py#L684) when it is initialized, including of course, the LangChain handler which attempts to import the `SessionController` again, which causes this issue.

By doing this, I think we can also avoid problem of re-loading the session each time that you guys pointed out previously.

Let me know what you think.

ea-rus (approved):
I've tested, it works locally. 
@tmichaeldb, do you have issues with it on windows?

ZoranPandovski (approved):

PR #7437, opened 2023-09-27
ML task queue

StpMax:
## Description

Please include a summary of the change and the issue it solves. 

**Fixes** #(issue)

## Type of change

(Please delete options that are not relevant)

- [ ] 🐛 Bug fix (non-breaking change which fixes an issue)
- [ ] ⚡ New feature (non-breaking change which adds functionality)
- [ ] 📢 Breaking change (fix or feature that would cause existing functionality not to work as expected)
- [ ] 📄 This change requires a documentation update

## Verification Process

To ensure the changes are working as expected:

 - [ ]   Test Location: Specify the URL or path for testing.
 - [ ]   Verification Steps: Outline the steps or queries needed to validate the change. Include any data, configurations, or actions required to reproduce or see the new functionality.

## Additional Media:

- [ ] I have attached a brief loom video or screenshots showcasing the new functionality or change.

## Checklist:

- [ ] My code follows the style guidelines(PEP 8) of MindsDB.
- [ ] I have appropriately commented on my code, especially in complex areas.
- [ ] Necessary documentation updates are either made or tracked in issues.
- [ ] Relevant unit and integration tests are updated or added.

PR #9359, opened 2024-06-17
Run prediction in partitions

### LINKED ISSUE #8842: [Bug]: Error if `JOIN` model with empty data

If run batch prediction query (`join` model with data), where data query returns empty list of records, then error appear.

### Video or screenshots

_No response_

### Expected behavior

_No response_

1. Follow `Forecast Quarterly House Sales`
2. instead of query:
```sql
SELECT m.saledate as date, m.ma as forecast
FROM mindsdb.house_sales_model as m
JOIN example_db.house_sales as t
WHERE t.saledate > LATEST
AND t.type = 'house'
AND t.bedrooms=2
LIMIT 4;
```
run
```sql
SELECT m.saledate as date, m.ma as forecast
FROM mindsdb.house_sales_model as m
JOIN example_db.house_sales as t
WHERE t.saledate > LATEST AND t.type = 'unit'
AND t.bedrooms=4
LIMIT 4;
```

### Anything else?

_No response_

ea-rus:
## Description

Depends on: https://github.com/mindsdb/mindsdb_sql/pull/382

Fixes https://linear.app/mindsdb/issue/SQL-20/research-partitioning-approach

Execute prediction in partitions. If partition_size param is used:
```sql
select * from int.tab1 a
    join proj.pred1 p1          
   using partition_size=1000
```
Planner will split fetched data into partitions by size using MapReduce step. Inside this step prediction and join steps will be executed over chunk of data. Execution will be split between sub-threads using [execute_in_threads](https://github.com/mindsdb/mindsdb/pull/9359/files#diff-b5ca614df036c24e57c3442aca7435c9ad1bc3d75d60eebe14274fbc0071af86R19) function. It keeps task queue in optimal size for workers (to have enough tasks available and not generate all tasks in memory at once)
On cloud ApplyPredictorStep will be executed by mindsdb-ml pod using task queue (implemented in [ML task queue PR](https://github.com/mindsdb/mindsdb/pull/7437)).
 By increasing count of these pods we will able to scale prediction horizontally 
Max count of parallel chunk executions is:
- for cloud: defined in MAX_QUERY_PARTITIONS env variable. 10 if not set
- for local: count of CPUs - 2 (keep 2 CPU cores for mindsdb server)

Minor changes:
- [changed approach](https://github.com/mindsdb/mindsdb/pull/9359/commits/5c835cf953690f1a0141085c469f9b8ec59ea077) to fix #8842)
- added 'stop after first failure' flag for some tests in CI

Demo: 
https://www.loom.com/share/c8269745c225492c85a25141e925fd59

## Type of change

## Verification Process

## Additional Media:

## Checklist:


mindsdb/api/executor/sql_query/steps/map_reduce_step.py
```diff
@@ -64,10 +55,98 @@ class MapReduceStepCall(BaseStepCall):
...
+        data = ResultSet()
+
+        df = input_data.get_raw_df()
+
+        # tasks
+        def split_data_f(df):
+            chunk = 0
+            while chunk * partition < len(df):
+                # create results with partition
+                df1 = df[chunk * partition: (chunk + 1) * partition]
```
paxcema, line 93: Just in case, we may wanna control for `partition > 0`, too, because if it's negative this won't behave correctly.

tests/unit/test_project_structure.py
```diff
@@ -955,6 +955,43 @@ def test_delete_from_table(self):
...
+
+        # use model
+        ret = self.run_sql('''
+             SELECT *
+               FROM files.tasks as t
+               JOIN mindsdb.task_model as m
+               using partition_size=2
+        ''')
+
+        # the same rows in output
```
paxcema, line 988: Shouldn't we also check for the index and/or row order? Or is merge order not guaranteed?

ea-rus: Yes, the order might be different.

mindsdb/api/executor/sql_query/steps/map_reduce_step.py
ea-rus: Added checking value of partition

tests/unit/test_project_structure.py
ea-rus: yes. planner don't split TS queries into partition at the moment

paxcema (approved):

mindsdb/api/executor/sql_query/steps/map_reduce_step.py
```diff
@@ -64,10 +55,100 @@ class MapReduceStepCall(BaseStepCall):
...
```
StpMax, line 95: looks like .iloc is faster
```
import pandas as pd
import numpy as np
import timeit

df = pd.DataFrame(np.random.randint(0, 100, size=(100000000, 10)), columns=list('ABCDEFGHIJ'))
def xx():
    return df[100:200]
def yy():
    return df.iloc[100:200]
print(timeit.timeit(xx, number=10000))
print(timeit.timeit(yy, number=10000))
```

mindsdb/api/executor/sql_query/steps/map_reduce_step.py
```diff
@@ -64,10 +55,100 @@ class MapReduceStepCall(BaseStepCall):
...
+        tasks = split_data_f(df)
+
+        # workers count
+        is_cloud = Config().get('cloud', False)
+        if is_cloud:
+            max_threads = os.getenv('MAX_QUERY_PARTITIONS')
+            if max_threads is not None:
+                max_threads = int(max_threads)
+            else:
+                max_threads = 10
```
StpMax, line 108: `max_threads = int(os.getenv('MAX_QUERY_PARTITIONS', 10))` :)

StpMax (changes requested):

StpMax (approved):

mindsdb/api/executor/sql_query/steps/map_reduce_step.py
ea-rus: thanks! updated

mindsdb/api/executor/sql_query/steps/map_reduce_step.py
ea-rus: changed to iloc

PR #9404, opened 2024-06-26
Re-organized Imports for BigQuery and Vertex to Allow Installation via GUI

MinuraPunchihewa:
## Description

This PR re-organizes the `google.cloud` imports used in the BigQuery and Vertex AI handlers to allow installation of dependencies via the GUI.

Fixes https://github.com/mindsdb/mindsdb/issues/9403

## Type of change

## Verification Process

## Additional Media:

Screenshot from 2024-06-26 19-55-57

Screenshot from 2024-06-26 19-55-49

## Checklist:

### DIFF
--- a/mindsdb/integrations/handlers/bigquery_handler/bigquery_handler.py
+++ b/mindsdb/integrations/handlers/bigquery_handler/bigquery_handler.py
@@ -1,7 +1,7 @@
-from google.cloud import bigquery
 from typing import Text, Dict, Any
 from google.api_core.exceptions import BadRequest
 from sqlalchemy_bigquery.base import BigQueryDialect
+from google.cloud.bigquery import Client, QueryJobConfig
 
 from mindsdb.utilities import log
 from mindsdb_sql.parser.ast.base import ASTNode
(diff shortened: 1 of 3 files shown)


MinuraPunchihewa:
@ZoranPandovski The above issue was a result of the following,
The error messages returned for these integrations from the `/handlers` endpoint was in the following format:
`cannot import name 'aiplatform' from 'google.cloud' (unknown location)
`
In the front-end, the only 'known error' that is accepted are messages starting with 'No module named', because this means that there are dependencies that need to be installed. It is only when such a message is given that the 'Install' button in the Manage Integrations page is displayed. Any other messages will result in an 'Unknown Error'.

I think the reason that we see a different error message for the above dependencies is because of how these imports are structured. By default, we install the `google-auth-oauthlib` package, which is imported as follows:
`from google.oauth2 import ...`

This probably causes somewhat of a conflict when the other `google` packages come into the picture.

Do you think we need to make any changes to the condition I have mentioned above (in the front-end)?

ZoranPandovski (approved):
As discussed let's keep this for now

PR #9442, opened 2024-07-03
Updated Amazon S3 to an Enterprise-ready API Integration

MinuraPunchihewa:
## Description

This PR converts the Amazon S3 integration to be an enterprise-ready API integration.

Fixes #issue_number

## Type of change

## Verification Process

## Additional Media:

## Checklist:


MinuraPunchihewa:
Hey @ea-rus, @ZoranPandovski,
I am in the process of converting our Amazon S3 integration to an API handler and there were a couple of questions I wanted to run past you guys.
1. The `boto3` package provides a [method](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/s3/client/select_object_content.html) that can be used to run SQL queries on files (of certain formats) stored in S3 directly, which is currently being used in our handler. Do you think this is a good way to go about or should I look at alternate options like `duckdb`? The advantage of using this method is that it does not require the entire file to be loaded into memory, however, there aren't any equivalent methods for the other operations (insert, update etc.).
2. I was thinking of adding 3 tables that users can interact with called `buckets`, `objects` and `object_contents`. The latter would be used to query the actual contents of the file. Do you think this makes it somewhat complicated? The other option I was considering is to ask the user to provide the bucket and key as part of the connection data and only allow that file to be queried. What are your thoughts on what the best approach would be?
An example of the type of query that would need to be run with the first approach would be something like this:
```
SELECT * FROM s3_datasource.object_contents
WHERE Bucket = 'mdb-dev-bucket' AND Key = 'iris.csv';
```

ea-rus:
@MinuraPunchihewa 
1. I think using of select_object_content is good because it will filter data remotely
2. what if 
- bucket it is connection argument. to use different bucket - create one more database in mindsdb
- list of tables - it is list of objects
- select from table - is querying the object

MinuraPunchihewa:
> @MinuraPunchihewa
> 
> 1. I think using of select_object_content is good because it will filter data remotely
> 2. what if
> 
> * bucket it is connection argument. to use different bucket - create one more database in mindsdb
> * list of tables - it is list of objects
> * select from table - is querying the object

I like the sound of this, honestly. I think in this case, we would need to develop this as a Database handler, right? I don't think it is possible to do this as an API handler?

ea-rus:
Yes, looks like data handler is more suitable for this task
>  I think in this case, we would need to develop this as a Database handler, right?

MinuraPunchihewa:
> Yes, looks like data handler is more suitable for this task
> 
> > I think in this case, we would need to develop this as a Database handler, right?

Thanks, Andrey. I appreciate your feedback. I will re-structure this as a Database handler.

MinuraPunchihewa:
Closing in favor of https://github.com/mindsdb/mindsdb/pull/9464

PR #9464, opened 2024-07-08
Amazon S3 Integration Improvements

MinuraPunchihewa:
## Description

This PR improves the Amazon S3 integration and makes it enterprise-ready.

Fixes https://github.com/mindsdb/mindsdb_private/issues/609

## Type of change

## Verification Process

## Checklist:


MinuraPunchihewa:
Hey @ZoranPandovski,
This makes a few changes to the existing structure of the Amazon S3 handler,
1. The users no longer need to provide the key they want to query as part of the connection arguments, just the bucket. They can then query any (supported) file in that bucket.
2. When querying the files, users will need to replace the periods (.) in the file names (typically those that come before the extension) with an underscore (_). For instance, if an user has a file named iris.csv in their bucket, the syntax for querying it would be as follows:
```
SELECT * 
FROM s3_datasource.iris_csv
```
What do you think of this? I think we should not allow the original file name to be specified because it'll probably parsed as multiple namespaces (schema.table).

MinuraPunchihewa:
Closing in favor of https://github.com/mindsdb/mindsdb/pull/9485

PR #9480, opened 2024-07-10
Pinned New Version of snowflake-sqlalchemy Supporting SQLAlchemy 2

MinuraPunchihewa:
## Description

This PR pins the new release of `snowflake-sqlalchemy` that supports SQLAlchemy 2. Here is the relevant reference: https://github.com/snowflakedb/snowflake-sqlalchemy/issues/380

In relation to this change, the following has also been done:
1. A previously added tip to install Git (when using Docker) in order to support the use of the forked repository for the above package has been removed from the documentation and the README.
2. The imports in the handler have been re-organized to prevent a circular dependency issue.

Fixes https://github.com/mindsdb/mindsdb/issues/9024

## Type of change

## Verification Process

 - [X]   Test Location: `tests/unit/handlers/test_snowflake.py` for unit tests and `mindsdb/integrations/handlers/snowflake_handler/tests/test_snowflake_handler.py` for integration tests.
 - [X]   Verification Steps: Run the above test modules (for integration tests, a Snowflake account will be required).

## Checklist:

### DIFF
--- a/docs/integrations/data-integrations/snowflake.mdx
+++ b/docs/integrations/data-integrations/snowflake.mdx
@@ -13,24 +13,6 @@ Before proceeding, ensure the following prerequisites are met:
 1. Install MindsDB locally via [Docker](/setup/self-hosted/docker) or [Docker Desktop](/setup/self-hosted/docker-desktop).
 2. To connect Snowflake to MindsDB, install the required dependencies following [this instruction](/setup/self-hosted/docker#install-dependencies).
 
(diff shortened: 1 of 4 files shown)


MinuraPunchihewa:
@ZoranPandovski I believe this works, but I was not able to test it by actually connecting to Snowflake because it seems like the free trial that had been granted to the account that we were using has been revoked.

ZoranPandovski (approved):

PR #9485, opened 2024-07-11
Amazon S3 Integration Improvements

MinuraPunchihewa:
## Description

This PR improves the Amazon S3 integration and makes it enterprise-ready.

Fixes https://github.com/mindsdb/mindsdb_private/issues/609

## Type of change

## Verification Process

 - [X]   Test Location: `tests/unit/handlers/test_s3.py` for unit tests.
 - [X]   Verification Steps: Run the above test module.

## Checklist:


MinuraPunchihewa:
@ZoranPandovski This is ready for review. It is only the integration tests that I have not been able to add yet. I will create an issue for this and work on it.

mindsdb/integrations/handlers/s3_handler/s3_handler.py
```diff
@@ -17,129 +15,198 @@
...
-        self.dialect = 's3'
         self.connection_data = connection_data
         self.kwargs = kwargs
+        self.key = None
 
         self.connection = None
         self.is_connected = False
 
+        self.is_select_query = False
+        self.key = None
```
ZoranPandovski, line 52: We initialize this above too

mindsdb/integrations/handlers/s3_handler/s3_handler.py
```diff
@@ -17,129 +15,198 @@
...
-            aws_access_key_id=self.connection_data['aws_access_key_id'],
-            aws_secret_access_key=self.connection_data['aws_secret_access_key'],
-            region_name=self.connection_data['region_name']
-        )
+        # Validate mandatory parameters.
+        if not all(key in self.connection_data for key in ['aws_access_key_id', 'aws_secret_access_key', 'bucket']):
+            raise ValueError('Required parameters (aws_access_key_id, aws_secret_access_key, bucket) must be provided.')
+
+        # Connect to S3 via DuckDB and configure mandatory credentials.
+        self.connection = self._connect_duckdb()
```
ZoranPandovski, line 76: We may catch exception here or in the connect_duckdb bellow

mindsdb/integrations/handlers/s3_handler/s3_handler.py
```diff
@@ -17,129 +15,198 @@
...
         """
-
         response = StatusResponse(False)
         need_to_close = self.is_connected is False
 
+        # Check connection via boto3.
         try:
-            connection = self.connect()
-            connection.head_object(Bucket=self.connection_data['bucket'], Key=self.connection_data['key'])
+            boto3_conn = self._connect_boto3()
```
ZoranPandovski, line 150: Can we somehow do this through Duckdb, or is this enough for now?

mindsdb/integrations/handlers/s3_handler/s3_handler.py
```diff
@@ -150,56 +217,137 @@ def native_query(self, query: str) -> StatusResponse:
...
+    def _create_table_from_file(self) -> None:
+        """
+        Creates a table from a file in the S3 bucket.
+
+        Raises:
+            CatalogException: If the file does not exist in the S3 bucket.
+        """
+        connection = self.connect()
+        try:
+            connection.execute(f"CREATE TABLE {self.table_name} AS SELECT * FROM 's3://{self.connection_data['bucket']}/{self.key}'")
```
ZoranPandovski, line 229: Can we check if table exist first?

ZoranPandovski (changes requested):
Thanks few comments added. One thing we should discuss is should we create a table for each query or re use some of the existing functions as `read_csv` to directly SELECT from the s3

mindsdb/integrations/handlers/s3_handler/s3_handler.py
MinuraPunchihewa: @ZoranPandovski I noticed that even if we run something like `SELECT 1;` with incorrect credentials, it runs without an issue. I think the only way to reliably check the connection through DuckDB is to get a list of files in the bucket and query one of them. Do you think we should do this?

mindsdb/integrations/handlers/s3_handler/s3_handler.py
MinuraPunchihewa: @ZoranPandovski I noticed that even if we provide incorrect credentials, these `SET` commands run without an issue. From what I have understood, the credentials are checked only when running a query against the S3 bucket. I can add some generic error handling around this. Do you think it is necessary?

MinuraPunchihewa:
@ZoranPandovski I have updated the logic to create a table only if one does not exist already by using `CREATE TABLE IF NOT EXISTS` queries.

From what I have gathered, it is not possible to run queries outside of `SELECT` without creating a table first, therefore, I do not think this can be optimized further (for other types of queries).

It is possible to optimize `SELECT` by querying the files directly without creating the table first. However, this requires us to run a `CREATE SECRET` command first as explained [here](https://duckdb.org/docs/extensions/httpfs/s3api.html); it does not work with the SET commands that are being in the above implementation.

Running `CREATE SECRET` queries require a newer version of DuckDB in comparison to what we have pinned to our main requirements (0.9.1). Since DuckDB is used (sparingly) in our core code-base, I was wondering if I should explore upgrading DuckDB to a newer version?

ZoranPandovski (approved):

PR #9493, opened 2024-07-16
Mindsdb starting speedup

ea-rus:
## Description

Changes:
- Speed up starting of mindsdb
  - Don't try to import all handlers in integration controller.  Import only when they used. 
  - Exception is '/api/handlers/' endpoint, which tries to import all handlers.
  - As temporal workaround handlers are imported in thread, without blocking start of the server
    -  but also consume some resources
- Changed endpoints:
  - added `/api/handlers/<handler name>` endpoint. It tries to import handler (if it is not) and return its info
  - Added 'lazy' arg to endpoint : `/api/handlers/?lazy=1` to get all handlers info without importing them.
- Side optimizations:
  - Removed using 'mysql' dialect, it adds several seconds in initialization time
  - moved import of 'analyze_dataset' to the place when it is used.
  - Fixes in tests (they used to leave temporal files) 

What to do next:

Stage 1: Update frontend
- in 'add..' -> 'new datasource':
  - use `/api/handlers/?lazy=1` for getting list of handlers
  - use `/api/handlers/<handler name>` when user clicks on handler icon to check import status

Stage 2: Remove import all handlers (in threads) on backend 

Fixes https://linear.app/mindsdb/issue/SQL-28/a-long-start-of-mindsdb

## Type of change

## Verification Process

## Additional Media:

## Checklist:


mindsdb/interfaces/database/integrations.py
```diff
@@ -695,33 +691,97 @@ def _load_handler_modules(self):
...
+            self.handlers_import_status[handler_name] = handler_meta
+
+        # import all handlers in thread
+        def import_handlers():
+            # give time to start server
+            time.sleep(3)
+            self.get_handlers_import_status()
 
-        self.handlers_import_status[handler_meta['name']] = handler_meta
+        thread = threading.Thread(target=import_handlers)
```
ea-rus, line 719: This is importing all handlers in thread. 
We can remove it because frontend

mindsdb/api/http/namespaces/handlers.py
```diff
@@ -26,13 +26,18 @@ class HandlersList(Resource):
     @api_endpoint_metrics('GET', '/handlers')
     def get(self):
         '''List all db handlers'''
-        handlers = ca.integration_controller.get_handlers_import_status()
+
+        if request.args.get('lasy') == '1':
```
StpMax, line 30: la**Z**y

StpMax (changes requested):

PR #9495, opened 2024-07-16
Added a Thread Safe Option for Retrieving from the Handler Cache

MinuraPunchihewa:
## Description

This PR adds a thread safe option for retrieving handlers from the handler cache.

## Type of change

## Verification Process

## Checklist:


MinuraPunchihewa:
Hey @StpMax,
Do you think that we should also allow items to be deleted from the cache in a thread safe manner? I think it is possible to do this, but do you see any negative implications?

mindsdb/interfaces/database/integrations.py
```diff
@@ -94,19 +94,31 @@ def set(self, handler: DatabaseHandler):
                 pass
             self._start_clean()
 
-    def get(self, name: str) -> Optional[DatabaseHandler]:
+    def get(self, name: str, thread_safe: bool = False) -> Optional[DatabaseHandler]:
```
StpMax, line 97: I don't really like to add additional arguments to functions. Better keep interface simple. What if add 'thread_safe' arg to the handler class itself?
Therefore 'set' may look like:
```python
def set(self, handler: DatabaseHandler):
    ...
    key = (handler.name, ctx.company_id, 0 if getattr(handler, 'thread_safe', False) else thread_id)
    ...
```
and 'get' like:
```python
def get(self, name: str) -> Optional[DatabaseHandler]:
    ...
    key = (name, ctx.company_id, thread_id)
    if key not in cache:
        key = (name, ctx.company_id, 0)
    if (
        key not in self.handlers
        or self.handlers[key]['expired_at'] < time()
    ):
        return None
    ...
```
what do you think?

MinuraPunchihewa: @StpMax This makes sense. I will push the suggested changes shortly.

MinuraPunchihewa:
Closing in favor of https://github.com/mindsdb/mindsdb/pull/9517

MinuraPunchihewa:
@StpMax I created a new PR with the suggested changes [here](https://github.com/mindsdb/mindsdb/pull/9517).

PR #9521, opened 2024-07-22
Added Model Modes to the Ollama Integration

MinuraPunchihewa:
## Description

This PR adds model modes to similar to some of our other integrations such as OpenAI, while fixing the bug reported below.

Fixes https://github.com/mindsdb/mindsdb/issues/9499

## Type of change

## Verification Process

## Checklist:


MinuraPunchihewa:
Hey @paxcema,
While coming up with a fix for the issue reported above, I was testing the handler using `moondream` (because it is the smallest model) and I noticed that it is possible to run this in two modes: embeddings and generate.
As a result, simply making the fix that would have solved the above issue makes the model run in embedding mode. Therefore, I added a `mode` parameter similar to our other engines and tweaked the logic for setting a default mode.
Let me know what you think of this and I will also update the documentation.

MinuraPunchihewa:
> Nice catch @MinuraPunchihewa, I was not aware some models supported both endpoints. This is a great solution!

Thanks, @paxcema. I have just pushed another commit by updating the docs with the use of the `mode` parameter.

PR #9664, opened 2024-08-27
updated elasticsearch docs

martyna-mindsdb:
## Description

updated elasticsearch docs

Fixes #issue_number

## Type of change

### DIFF
--- a/docs/integrations/data-integrations/elasticsearch.mdx
+++ b/docs/integrations/data-integrations/elasticsearch.mdx
@@ -1,31 +1,54 @@
 ---
-title: Elasticsearch
-sidebarTitle: Elasticsearch
+title: ElasticSearch
+sidebarTitle: ElasticSearch
 ---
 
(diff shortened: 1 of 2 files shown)


ZoranPandovski (approved):

MinuraPunchihewa (commented):
Hey @martyna-mindsdb,
Sorry for the late review, but I think we need to make a few changes,

- For Cloud connections, it is not necessary to provide both the `cloud_id` and the `endpoint`. Only one of them is required.

- Similarly, only one of `api_key` or a combination of `user` and `password` is required.
- I think we should also avoid mentioning the default `elastic` user. I am quite certain it is possible to create other users as well.

- The entire section below is in a code block; is that how it is meant to be?
image

martyna-mindsdb:
Hey @MinuraPunchihewa 
Thanks for the comments!

I tested different ways of connecting ElasticSearch to MindsDB, but it seems that `cloud_id` is required even if `endpoints` is provided.
Screenshot 2024-08-28 at 12 35 40
Screenshot 2024-08-28 at 12 35 02

PR #9681, opened 2024-08-30
Return Quick Response as a Message, not Output

tmichaeldb:
We treat `output` as a final response on the frontend, so the quick response should be yielded as a normal message.


mpacheco12 (approved):
LGTM

PR #9700, opened 2024-09-06
Added a Check for Directory Path Traversal

MinuraPunchihewa:
## Description

This PR adds a check for directory path traversal when accessing resources from the server.

Fixes https://linear.app/mindsdb/issue/BE-526/uncontrolled-data-used-in-path-expression

## Type of change

## Checklist:

### DIFF
--- a/mindsdb/integrations/handlers/confluence_handler/__init__.py
+++ b/mindsdb/integrations/handlers/confluence_handler/__init__.py
@@ -1,6 +1,7 @@
 from mindsdb.integrations.libs.const import HANDLER_TYPE
 
 from .__about__ import __version__ as version, __description__ as description
+from .connection_args import connection_args, connection_args_example
 
 try:
     from .confluence_handler import ConfluenceHandler as Handler
@@ -23,4 +24,6 @@ __all__ = [
     "description",
     "import_error",
     "icon_path",
+    "connection_args_example",
+    "connection_args",
 ]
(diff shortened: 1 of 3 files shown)


mindsdb/api/http/initialize.py
```diff
@@ -216,7 +216,17 @@ def root_index(path):
                 'Not found',
                 'The endpoint you are trying to access does not exist on the server.'
             )
-        if static_root.joinpath(path).is_file():
+
+        # Check for directory path traversal.
+        full_path = os.path.normpath(os.path.join(static_root, path))
+        if not full_path.startswith(str(static_root)):
```
StpMax, line 222: Better to use pathlib.Path. Something like
```python
if not (static_path / path).resolve().is_relative_to(static_path):
```

StpMax (changes requested):

mindsdb/api/http/initialize.py
MinuraPunchihewa: Hey @StpMax,
From what I have understood, it looks like the `is_relative_to` method is only available in Python 3.10 and above as explained in this [issue](https://github.com/Significant-Gravitas/AutoGPT/issues/2027). Should we still make this change?

MinuraPunchihewa: Hey @StpMax,
Sorry for the delay in getting to this, but can you please check if the change I've made is in line with what you have advised?

PR #9783, opened 2024-10-03
Speed up CI/CD workflows with UV

hamishfagg:
Replaces `pip install` with `uv pip install` in our docker builds and github actions.
Uv is a python project management tool, but we are just using the pip-compatible interface because it is much faster than pip.


hamishfagg:
FYI, the integration test that's failing is fixed in #10075

mindsdb-devops (approved):
lgtm

PR #9978, opened 2024-10-22
Added Support for SELECT * With Alias for Files

MinuraPunchihewa:
## Description

This PR adds supports for running queries such as the following against files:
`SELECT x.* FROM files.<file_name> x`

Fixes https://github.com/mindsdb/mindsdb/issues/9977

## Type of change

## Verification Process

## Additional Media:

image

## Checklist:


mindsdb/api/executor/utilities/sql.py
```diff
@@ -127,6 +127,8 @@ def adapt_query(node, is_table, **kwargs):
             return
         if isinstance(node, Identifier):
             if len(node.parts) > 1:
+                if isinstance(node.parts[-1], Star):
```
ea-rus, line 130: is better to move it to file_handler because the reason of error: file handler doesn't support sql completely and tries to  compensate it with query_df
also changing something globally is more dangerous

MinuraPunchihewa: @ea-rus I just pushed a change that handles this in the File handler, but is this the right way to go about it? 
When you execute a `SELECT <alias>.*` query, the inner method `adapt_query` of `query_df()` returns an `Identifier(Star:<*>)` object and this is treated (correctly) as an `Identifier` object in `mindsdb_sql` rather than a `Star` object, which is what causes the error (the Star class does not have an upper() function). 
I feel that maybe this is something that should be fixed?

ea-rus: I don't understand, do you mean if the query has 'select t.*' then it converted to 'select *' in file handler it leads to error?

MinuraPunchihewa: This is what the query looks like after going through the query traversal:
image

The alias is removed from the target but it is an Identifier object (`Identifier:<*>`).

When this is run through `mindsdb_sql` (`SqlalchemyRender`), since it is an `Identifier`, it is treated as such and the error gets raised [here](https://github.com/mindsdb/mindsdb_sql/blob/bc8518cb391c7d0d7a5157373e78ad7117ed36e8/mindsdb_sql/render/sqlalchemy_render.py#L125).

ea-rus: But looking at [code](https://github.com/mindsdb/mindsdb/pull/9978/files#diff-59572afa2a6d13b473a40b975991eee85f18ec530dd0d27a93f4b6735e174d09R104) it have to be Star() in targets, not Identifier([Star()])

martyna-mindsdb:
@MinuraPunchihewa @ea-rus 
Is this PR still valid?

ea-rus:
I just checked, this query works for me on current 'main'
```sql
SELECT x.* FROM files.test x
```

PR #9879, opened 2024-10-15
S3 across buckets

ea-rus:
## Description

Bucket is optional paremeter

```sql
CREATE DATABASE s3_datasource
WITH
    engine = 's3',
    parameters = {
       "aws_access_key_id": "-"
      , "aws_secret_access_key": "-"
      , "aws_session_token": "-"
      , "bucket": "mdb-dev-developer-test" -- < -- optional
    };
```

Updates:
- If database is created without bucket: bucked is added to part of the file
- 'region_name' param is not used
- In files table added column 'bucket'
- To speed up selecting from `files` if bucked is not set in database you can use:
  - limit
  - filter by `bucket` column

Demo: https://www.loom.com/share/501d0fc0b299499884a43f0bf6d1c453

Fixes https://linear.app/mindsdb/issue/BE-304/s3-switch-buckets

## Type of change

(Please delete options that are not relevant)

- [ ] 🐛 Bug fix (non-breaking change which fixes an issue)
- [ ] ⚡ New feature (non-breaking change which adds functionality)
- [ ] 📢 Breaking change (fix or feature that would cause existing functionality not to work as expected)
- [ ] 📄 This change requires a documentation update

## Verification Process

To ensure the changes are working as expected:

 - [ ]   Test Location: Specify the URL or path for testing.
 - [ ]   Verification Steps: Outline the steps or queries needed to validate the change. Include any data, configurations, or actions required to reproduce or see the new functionality.

## Additional Media:

- [x] I have attached a brief loom video or screenshots showcasing the new functionality or change.

## Checklist:

- [x] My code follows the style guidelines(PEP 8) of MindsDB.
- [ ] I have appropriately commented on my code, especially in complex areas.
- [ ] Necessary documentation updates are either made or tracked in issues.
- [x] Relevant unit and integration tests are updated or added.

PR #9979, opened 2024-10-22
Updated the Documentation for S3 Integration Based on Recent Improvements

MinuraPunchihewa:
## Description

This PR updates the documentation for S3 integration based on the recent improvements made by https://github.com/mindsdb/mindsdb/pull/9804 and https://github.com/mindsdb/mindsdb/pull/9879.

Fixes https://linear.app/mindsdb/issue/BE-364/update-the-documentation-for-the-s3-integration-based-on-recent

## Type of change

- [X] 📄 This change is a documentation update

### DIFF
--- a/docs/integrations/data-integrations/amazon-s3.mdx
+++ b/docs/integrations/data-integrations/amazon-s3.mdx
@@ -24,6 +24,10 @@ WITH
     };
 ```
 
+<Note>
+Note that sample parameter values are provided here for reference, and you should replace them with your connection parameters.
+</Note>
+
 Required connection parameters include the following:
 
 * `aws_access_key_id`: The AWS access key that identifies the user or IAM role.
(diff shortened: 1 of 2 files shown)


docs/integrations/data-integrations/amazon-s3.mdx
```diff
@@ -20,7 +20,6 @@ WITH
     parameters = {
       "aws_access_key_id": "AQAXEQK89OX07YS34OP"
       "aws_secret_access_key": "wJalrXUtnFEMI/K7MDENG/bPxRfiCYEXAMPLEKEY",
```
ea-rus, line 22: maybe also to remove values which looks like real and replace them with something like 'YOUR ACCESS KEY',

MinuraPunchihewa: @ea-rus It looks like the documentation for our other integration also use examples that are fake but look real such as [this](https://docs.mindsdb.com/integrations/data-integrations/snowflake). 
@martyna-mindsdb What should our approach be here?

martyna-mindsdb: What we do in the Minds docs is:
- we provide sample (fake) values for users' reference, as it used to happen in the past that users provided the wrong value because they were not clear on what should be the sample format.
- we note that these must be replaced with real values like `Note that sample parameter values are provided here for reference, and you should replace them with your connection parameters.`

Ref - https://docs.mdb.ai/docs/data_sources

MinuraPunchihewa: @martyna-mindsdb I have added this in as a note.

martyna-mindsdb (approved):

PR #10055, opened 2024-10-26
Refactor DuckDBHandler to improve connection handling and handle NoneType DataFrame in show table query

### LINKED ISSUE #10054: [Bug]: Concurrency problem in duckdb

Follow up: #10043

According to duckdb [docs](https://duckdb.org/docs/connect/concurrency.html):

DuckDB has two configurable options for concurrency:

- One process can both read and write to the database.
- Multiple processes can read from the database, but no processes can write.

---

My code:

```sql
-- connect with DuckDB data source and chromadb vector store

CREATE DATABASE duckdb_datasource
WITH
engine='duckdb',
parameters={
    "database":"db.duckdb"
};


CREATE DATABASE chroma_dev_local
WITH ENGINE = 'chromadb'
PARAMETERS = {
    "host": "localhost",
    "port": "8000"
};

-- Create an embedding model for the knowledge base using LangChain as an engine
CREATE ML_ENGINE embedding
FROM langchain_embedding;

CREATE MODEL embedding_model
PREDICT embeddings
USING
	engine = "langchain_embedding",
	class = "openai",
	    openai_api_key = "sk-",
	input_columns = ["content"];


-- Create an index in the vector store and insert all data from duckdb_datasource
CREATE TABLE chroma_dev_local.world_news_index (
    SELECT CAST(w.id AS CHAR) AS id, w.content, e.embeddings 
    FROM duckdb_datasource.world_news w
    INNER JOIN embedding_model e
    ON w.content = e.content
);


-- Create a knowledge base
CREATE KNOWLEDGE_BASE world_news_kb
USING
	model = embedding_model,
    storage = chroma_dev_local.world_news_index;

-- Query the knowledge base
SELECT *
FROM world_news_kb;

SELECT *
FROM world_news_kb
WHERE content = "Global leaders"
LIMIT 1;


-- Create a job
CREATE JOB keep_knowledge_base_up_to_date AS (

    -- to import data from `duckdb_datasource` to vector storage if available
    INSERT INTO chroma_dev_local.world_news_index (
        SELECT CAST(w.id AS CHAR) AS id, w.content, e.embeddings 
        FROM duckdb_datasource.world_news w
        JOIN embedding_model e
        ON w.content = e.content
        WHERE id > LAST
    );

    -- import data from vector stroage to knowledge base (KB) if available
    INSERT INTO world_news_kb (
        SELECT id, content
        FROM chroma_dev_local.world_news_index
        WHERE id > LAST
    );

) EVERY 1 min;

show jobs;


-- Query the knowledge base
SELECT *
FROM world_news_kb;

-- Query original table
SELECT *
FROM duckdb_datasource.world_news;

SELECT *
FROM world_news_kb
WHERE content = "Health officials"
LIMIT 1;
```

Error (Only one time occurs beacuase job is taking the r/w lock after this error and therefore insertion to the same db is impossible):

```bash
2024-10-26 20:37:49,927            jobs ERROR    mindsdb.interfaces.jobs.jobs_controller: [duckdb/duckdb_datasource]: IO Error: Could not set lock on file "/home/ritwickrajmakhal/mindsdb/db.duckdb": Resource temporarily unavailable
```
My opinion with current implementation:

- This handler should create and close the connection in each query.
- Even if we implement the whole code in read_only mode still we will not be able to insert any data to that DB until we stop the mindsdb with closing the connection.
```
Error: unable to open database "/home/ritwickrajmakhal/mindsdb/db.duckdb": IO Error: Could not set lock on file "/home/ritwickrajmakhal/mindsdb/db.duckdb": Conflicting lock is held in /usr/bin/python3.10 (PID 49412) by user ritwickrajmakhal. However, you would be able to open this database in read-only mode, e.g. by using the -readonly parameter in the CLI. See also https://duckdb.org/docs/connect/concurrency
```
### Video or screenshots

_No response_

### Expected behavior

I have fixed the problem by closing the connection with each query. Now I am able to sync the Knowledge base as expected.

_No response_

### Anything else?

_No response_

ritwickrajmakhal:
Fixes #10054 

## Type of change

## Checklist:

### DIFF
--- a/mindsdb/integrations/handlers/duckdb_handler/duckdb_handler.py
+++ b/mindsdb/integrations/handlers/duckdb_handler/duckdb_handler.py
@@ -99,8 +99,6 @@ class DuckDBHandler(DatabaseHandler):
         Returns:
             Response: The query result.
         """
-        need_to_close = self.is_connected is False
-
         connection = self.connect()
         cursor = connection.cursor()
 
@@ -125,8 +123,7 @@ class DuckDBHandler(DatabaseHandler):
             response = Response(RESPONSE_TYPE.ERROR, error_message=str(e))
 
         cursor.close()
-        if need_to_close is True:
-            self.disconnect()
+        self.disconnect()
 
         return response
 
@@ -153,7 +150,10 @@ class DuckDBHandler(DatabaseHandler):
         q = 'SHOW TABLES;'
         result = self.native_query(q)
         df = result.data_frame
-        result.data_frame = df.rename(columns={df.columns[0]: 'table_name'})
+
+        if df is not None:
+            result.data_frame = df.rename(columns={df.columns[0]: 'table_name'})
+
         return result
 
     def get_columns(self, table_name: str) -> Response:


mindsdb/integrations/handlers/duckdb_handler/duckdb_handler.py
```diff
@@ -125,8 +123,7 @@ def native_query(self, query: str) -> Response:
             response = Response(RESPONSE_TYPE.ERROR, error_message=str(e))
 
         cursor.close()
-        if need_to_close is True:
```
MinuraPunchihewa, line 128: Why has this call to `self.disconnect()` been removed?

ritwickrajmakhal: @MinuraPunchihewa Can you check carefully? I havn't remove the call to `self.disconnect()` method, only removed the condition. So that the the `disconnect` method executes every time.

ritwickrajmakhal: @MinuraPunchihewa 
Can you check it carefully? I havn't remove the call to `self.disconnect()` method, only removed the condition. So that the the `disconnect` method executes every time.

ritwickrajmakhal: Because of that condition I was unable to sync the knowledge base. You can see the demo [here](https://github.com/mindsdb/mindsdb/issues/10043#issuecomment-2439634021) .

ZoranPandovski: We need to be careful when removing checks that work only for specific use cases.

ritwickrajmakhal: Yes, I am also agree with  @ZoranPandovski, For this type of situation we can check git logs and look whether these lines are changed previously or not for any reason.

In my case that `need_to_close is True:` condition was there from the begining of the handler.

duck

chandrevdw31:
@ritwickrajmakhal please dm me once this is merged

mindsdb/integrations/handlers/duckdb_handler/duckdb_handler.py
ZoranPandovski: Removing the if condition shouldn't be the way to solve this, closing this PR

PR #10099, opened 2024-10-31
GROUP BY must go before ORDER BY Update neuralforecast.mdx 

chuangyeshuo:
GROUP BY must go before ORDER BY

## Description

Please include a summary of the change and the issue it solves. 

Fixes #issue_number

## Type of change

(Please delete options that are not relevant)

- [x] 🐛 Bug fix (non-breaking change which fixes an issue)
- [x] ⚡ New feature (non-breaking change which adds functionality)
- [x] 📢 Breaking change (fix or feature that would cause existing functionality not to work as expected)
- [x] 📄 This change requires a documentation update

## Verification Process

To ensure the changes are working as expected:

 - [x]   Test Location: Specify the URL or path for testing.
 - [x]   Verification Steps: Outline the steps or queries needed to validate the change. Include any data, configurations, or actions required to reproduce or see the new functionality.

## Additional Media:

- [x] I have attached a brief loom video or screenshots showcasing the new functionality or change.

## Checklist:

- [x] My code follows the style guidelines(PEP 8) of MindsDB.
- [x] I have appropriately commented on my code, especially in complex areas.
- [x] Necessary documentation updates are either made or tracked in issues.
- [x] Relevant unit and integration tests are updated or added.

PR #10119, opened 2024-11-05
Update neuralforecast.mdx

chuangyeshuo:
I think this is more well, thanks:
the correct SQL writing order is
(SELECT)
Getting Data (FROM/JOIN)
Row Filter (WHERE)
Grouping (GROUP BY)
Group Filter (HAVING)
Order & Paging (ORDER BY & LIMIT/OFFSET)


martyna-mindsdb:
@chuangyeshuo 
Please move all these updates into your initial PR here - https://github.com/mindsdb/mindsdb/pull/10099

PR #10146, opened 2024-11-11
Optimize join tables from different databases: executor

ea-rus:
## Description

Updates:

Subselect step is used to get value from previous step data.
```sql
select distinct <column> from <step data1>
```
Next step is fetching data using these values as filter
```sql
select * from db2.table2 where <column> in (<ids from previous step>)
```

Side fix: 
If join is without condition:
- add `0=0` filter (multiply rows)
- but use limitation. If expected number of rows is exceed limit - raise exception

Dependent on https://github.com/mindsdb/mindsdb_sql/pull/412

Fixes #issue_number

## Type of change

## Verification Process

## Additional Media:

## Checklist:

- [x My code follows the style guidelines(PEP 8) of MindsDB.

### DIFF
--- a/mindsdb/api/executor/sql_query/steps/join_step.py
+++ b/mindsdb/api/executor/sql_query/steps/join_step.py
@@ -3,7 +3,7 @@ import copy
 import numpy as np
 
 from mindsdb_sql.parser.ast import (
-    Identifier,
+    Identifier, BinaryOperation, Constant
 )
 from mindsdb_sql.planner.steps import (
     JoinStep,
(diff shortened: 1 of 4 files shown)


mindsdb/api/executor/sql_query/steps/join_step.py
```diff
@@ -74,7 +74,11 @@ def adapt_condition(node, **kwargs):
                     return Identifier(parts=['table_b', col_name])
 
             if step.query.condition is None:
-                raise NotSupportedYet('Unable to join table without condition')
+                # prevent memory overflow
+                if len(left_data) * len(right_data) < 10 ** 7:
```
StpMax, line 78: Are `left_data` and `right_data` dataframes? If so, then may be better to get real size (`df.memory_usage(index=True, deep=True).sum()`) and compare with free memory?

StpMax (approved):

mindsdb/api/executor/sql_query/steps/join_step.py
ea-rus: They are [ResultSets](https://github.com/mindsdb/mindsdb/blob/1f4907b9041b3f5ebc8bc2f6ba12598553588371/mindsdb/api/executor/sql_query/result_set.py#L38)

PR #10184, opened 2024-11-18
KnowledgeBase insert/select only with id_column if provided

### LINKED ISSUE #9652: [Bug]: strange behavior if  `knowledge_base`  `id_column` != `id`

If create a `knowledge_base` with `id_column` that is different from the `id`, then inserting a row with `id` column still update `id` column of knowledge base.

### Video or screenshots

_No response_

### Expected behavior

If i create a `knowlage_base` with `id_column` != `id`, then inserting row with `id` column should not update `id` column of `knowlage_base`

```sql
CREATE KNOWLEDGE_BASE world_news_kb USING id_column='index';
insert into world_news_kb (SELECT 1 as id, 'v1' as content FROM example_db.home_rentals LIMIT 1);
insert into world_news_kb (SELECT 2 as index, 'v2' as content FROM example_db.home_rentals LIMIT 1);
select * from  world_news_kb;
```
output is: 
```
id    content
1     v1
2     v2
```

### Anything else?

_No response_

parthiv11:
## Description
- added custom columns (if were provided) to returned records for `select query`
- for `insert query` removed default `id` column consider used as default 

Fixes #9652 

## Type of change

## Verification Process

## Additional Media:

## Checklist:


parthiv11:
this is updated behaviour
image

mindsdb/interfaces/knowledge_base/controller.py
```diff
@@ -285,9 +292,9 @@ def _adapt_column_names(self, df: pd.DataFrame) -> pd.DataFrame:
             # wrong name
             id_column = None
 
-        if id_column is None and TableField.ID.value in columns:
-            # default value
-            id_column = TableField.ID.value
+        # if id_column is None and TableField.ID.value in columns:
+        #     # default value
+        #     id_column = TableField.ID.value
```
ea-rus, line 297: This removes logic to use default 'id' column if `id_column` is not defined.

Swapping these two blocks should fix the issue:
First sets default value for id, second clears column id if setting is set but column is not exists
```python
        if id_column is None and TableField.ID.value in columns:
            # default value
            id_column = TableField.ID.value

        if id_column is not None and id_column not in columns:
            # wrong name
            id_column = None

```

mindsdb/interfaces/knowledge_base/controller.py
```diff
@@ -88,6 +88,13 @@ def select_query(self, query: Select) -> pd.DataFrame:
         # send to vectordb
         db_handler = self.get_vector_db()
         resp = db_handler.query(query)
+        new_columns = [
+            self._kb.params.get('id_column', TableField.ID.value),
+            self._kb.params.get('content_columns', TableField.CONTENT.value),
+            self._kb.params.get('metadata_columns', TableField.METADATA.value)
+        ]
+
+        resp.data_frame.columns = new_columns
```
ea-rus, line 97: is better not use it, the number of columns could be different. and output column names should be already correct

parthiv11: ok, but i think id_column need to be reflected in output

mindsdb/interfaces/knowledge_base/controller.py
dusvyat: for second row, I think it should just be 2 and not 2_content, for first row it looks correct as we specify 'index' is the id_column. so it treats 'id' column as a content column

dusvyat:
@parthiv11 please create some unit tests for us to better test and validate expected behaviour

mindsdb/interfaces/knowledge_base/controller.py
parthiv11: **Description:**

1. **Multiple Content Columns:**
   - Current behavior: When multiple content columns (e.g., `content` and `content2`) are specified in an `INSERT` statement, separate records are created with IDs like `{id}_content` and `{id}_content2`.
   - Question: Should the content from these columns be appended into a single document (with a single ID) or remain as separate documents?

   **Example:**
   ```sql
   INSERT INTO world_news_kb (
       SELECT 1 AS index, 'v1' AS content, 'v2' AS content2 
       FROM example_db.home_rentals LIMIT 1
   );
   ```
   - Current Result:
     - ID: `1_content`, Content: `v1`
     - ID: `1_content2`, Content: `v2`
   - Expected Result: ???

2. **Chunked Content:**
   - Current behavior: When content is too long and split into chunks, all chunks are assigned the same ID, which causes duplicates. During deduplication, only the first chunk is kept.
   - Question: How should IDs for chunks be handled? Options:
     - Use a suffix like `{id}_chunk_{1, 2, 3}` for each chunk.
     - Other suggestions?

   **Example:**
   ```sql
   INSERT INTO world_news_kb (
       SELECT 2 AS index, 'very_long_text_will_divided_in_chunks' AS content 
       FROM example_db.home_rentals LIMIT 1
   );
   ```
   - Current Result:
     - All chunks have ID: `2`
   - Expected Result: ???

parthiv11:
@dusvyat should I close this PR? as #10210 will solve this issue

dusvyat:
@parthiv11 sounds good please

PR #10210, opened 2024-11-25
fix bug with kb

dusvyat:
## Description

bug:

1. metadata is failing to be added as json.loads fails due to single quotes
2. when iterating thru df rows it tries to reference original column names but this fails as columns have been nested into metadata array column already
3. upsert fails as non-deterministic ids are created during preprocessing step

https://www.loom.com/share/36ce5203b2954fd19812279c418af126

```

CREATE KNOWLEDGE_BASE kb11 USING 
model = openai_emb, 
metadata_columns=['sqft', 'number_of_rooms'],
content_columns=['location', 'neighborhood'];

insert into kb11 (SELECT 1 as id, * FROM example_db.home_rentals LIMIT 1);
```

Fixes #issue_number

## Type of change

## Verification Process

 - [x]   Test Location: [Specify the URL or path for testing.](https://www.loom.com/share/bddb9e902610479280a8cc4656f43785)

## Additional Media:

## Checklist:

### DIFF
--- a/mindsdb/integrations/handlers/chromadb_handler/chromadb_handler.py
+++ b/mindsdb/integrations/handlers/chromadb_handler/chromadb_handler.py
@@ -370,10 +370,15 @@ class ChromaDBHandler(VectorStoreHandler):
                     if existing_dim != embedding_dims[0]:
                         raise Exception(f"Embedding dimension {embedding_dims[0]} does not match collection dimensionality {existing_dim}")
 
-        # For SELECT statements without IDs, deduplicate based on content
+        # Only generate IDs if they're not provided
         if TableField.ID.value not in df.columns:
             df = df.drop_duplicates(subset=[TableField.CONTENT.value])
             df[TableField.ID.value] = [str(uuid.uuid4()) for _ in range(len(df))]
+        else:
+            # Ensure IDs are strings
+            df[TableField.ID.value] = df[TableField.ID.value].astype(str)
+            # Drop duplicates based on ID to handle upserts properly
+            df = df.drop_duplicates(subset=[TableField.ID.value], keep='last')
 
         # Extract data from DataFrame
         data_dict = df.to_dict(orient="list")
--- a/mindsdb/integrations/handlers/chromadb_handler/chromadb_handler.py
+++ b/mindsdb/integrations/handlers/chromadb_handler/chromadb_handler.py
@@ -1,6 +1,7 @@
 import ast
 import sys
 from typing import Dict, List, Optional, Union
+import uuid
 
 import pandas as pd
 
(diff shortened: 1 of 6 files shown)


dusvyat:
@parthiv11 fyi, I think this resolves the issues mentioned in https://github.com/mindsdb/mindsdb/pull/10184
PTAL

dusvyat:
@tmichaeldb fyi

parthiv11:
https://github.com/mindsdb/mindsdb/blob/056fa0f48fe9f59caf04f78713fd02fa30be9ce8/mindsdb/integrations/handlers/chromadb_handler/chromadb_handler.py#L288

**`id_filters`** should be made `None` if  `[ ]` (empty array) before passing it to `collection.get` func, otherwise `SELECT * FROM kb11` wont work

parthiv11:
https://github.com/mindsdb/mindsdb/blob/056fa0f48fe9f59caf04f78713fd02fa30be9ce8/mindsdb/interfaces/knowledge_base/controller.py#L97-L98

before returning `resp.data_frame`, need to update `id_column` in `df` columns, if provided(from self._kb.params)

dusvyat:
> https://github.com/mindsdb/mindsdb/blob/056fa0f48fe9f59caf04f78713fd02fa30be9ce8/mindsdb/interfaces/knowledge_base/controller.py#L97-L98
> 
> before returning `resp.data_frame`, need to update `id_column` in `df` columns, if provided(from self._kb.params)
> 
> This will reflect `id_column` in query result like below:

@parthiv11 select * from kb is working for me with chromadb

parthiv11:
> > https://github.com/mindsdb/mindsdb/blob/056fa0f48fe9f59caf04f78713fd02fa30be9ce8/mindsdb/interfaces/knowledge_base/controller.py#L97-L98
> > 
> > before returning `resp.data_frame`, need to update `id_column` in `df` columns, if provided(from self._kb.params)
> > This will reflect `id_column` in query result like below:
> 
> @parthiv11 I don't think we want to return id column as the one that is specified by user, that is only for ingestion purposes

Shouldn't this be the expected behavior? It seems more intuitive for a typical user.

mindsdb/interfaces/knowledge_base/preprocessing/document_preprocessor.py
```diff
@@ -122,27 +146,34 @@ def _split_document(self, doc: Document) -> List[Document]:
...
+            # Single chunk case
+            if len(chunk_docs) == 1:
+                context = self._generate_context(chunk_docs[0].content, doc.content)
+                processed_content = f"{context}\n\n{chunk_docs[0].content}"
 
-                # Need a unique ID for each document. Can track source ID in metadata.
-                metadata = chunk_doc.metadata or doc.metadata or {}
-                metadata['doc_id'] = doc.id
                 processed_chunks.append(ProcessedChunk(
-                    id=uuid4().hex,
```
dusvyat, line 141: these need to be deterministic otherwise upsert won't work @tmichaeldb fyi

mindsdb/integrations/handlers/chromadb_handler/chromadb_handler.py
```diff
@@ -316,44 +314,86 @@ def select(
...
+                if existing and existing['embeddings']:
+                    existing_dim = len(existing['embeddings'][0])
+                    if existing_dim != embedding_dims[0]:
+                        raise Exception(f"Embedding dimension {embedding_dims[0]} does not match collection dimensionality {existing_dim}")
 
-        # convert to dict
+        # For SELECT statements without IDs, deduplicate based on content
+        if TableField.ID.value not in df.columns:
+            df = df.drop_duplicates(subset=[TableField.CONTENT.value])
+            df[TableField.ID.value] = [str(uuid.uuid4()) for _ in range(len(df))]
```
ea-rus, line 376: Just in case: if `insert` is called from KB controller - it always has ID column

ea-rus (approved):

PR #10233, opened 2024-12-03
Fixed the Errors Associated with Importing Handler Modules

MinuraPunchihewa:
## Description

This PR adds a fix to the error often encountered in MindsDB with the following message:
`AttributeError: 'NoneType' object has no attribute 'Handler'`

This is done by improving the error handling around the logic used to import handler modules.

Fixes https://linear.app/mindsdb/issue/BE-524/please-fix-the-most-annoying-constant-error-in-mindsdb-attributeerror

## Type of change

## Verification Process

## Additional Media:

## Checklist:

### DIFF
--- a/mindsdb/api/http/namespaces/databases.py
+++ b/mindsdb/api/http/namespaces/databases.py
@@ -63,10 +63,10 @@ class DatabasesResource(Resource):
         if check_connection:
             try:
                 handler = session.integration_controller.create_tmp_handler(name, database['engine'], parameters)
-            except ImportError as imort_error:
+            except ImportError as import_error:
                 return http_error(
                     HTTPStatus.BAD_REQUEST, 'Error',
-                    f'Could not create database handler: {str(imort_error)}'
+                    f'Could not create database handler: {str(import_error)}'
                 )
 
             status = handler.check_connection()
@@ -168,10 +168,10 @@ class DatabaseResource(Resource):
                 handler = session.integration_controller.create_tmp_handler(
                     temp_name, existing_integration['engine'], parameters
                 )
-            except ImportError as imort_error:
+            except ImportError as import_error:
                 return http_error(
                     HTTPStatus.BAD_REQUEST, 'Error',
-                    f'Could not create database handler: {str(imort_error)}'
+                    f'Could not create database handler: {str(import_error)}'
                 )
 
             status = handler.check_connection()
--- a/mindsdb/api/http/namespaces/databases.py
+++ b/mindsdb/api/http/namespaces/databases.py
@@ -16,6 +16,7 @@ from mindsdb.metrics.metrics import api_endpoint_metrics
 from mindsdb_sql_parser import parse_sql, ParsingException
 from mindsdb_sql_parser.ast import CreateTable, DropTables
 from mindsdb.utilities.exception import EntityNotExistsError
+from mindsdb.integrations.libs.response import HandlerStatusResponse
 
 
 @ns_conf.route('/')
@@ -63,13 +64,10 @@ class DatabasesResource(Resource):
         if check_connection:
             try:
                 handler = session.integration_controller.create_tmp_handler(name, database['engine'], parameters)
+                status = handler.check_connection()
             except ImportError as import_error:
-                return http_error(
-                    HTTPStatus.BAD_REQUEST, 'Error',
-                    f'Could not create database handler: {str(import_error)}'
-                )
+                status = HandlerStatusResponse(success=False, error_message=str(import_error))
 
-            status = handler.check_connection()
             if status.success is not True:
                 if hasattr(status, 'redirect_url') and isinstance(status, str):
                     return {
@@ -168,13 +166,10 @@ class DatabaseResource(Resource):
                 handler = session.integration_controller.create_tmp_handler(
                     temp_name, existing_integration['engine'], parameters
                 )
+                status = handler.check_connection()
             except ImportError as import_error:
-                return http_error(
-                    HTTPStatus.BAD_REQUEST, 'Error',
-                    f'Could not create database handler: {str(import_error)}'
-                )
+                status = HandlerStatusResponse(success=False, error_message=str(import_error))
 
-            status = handler.check_connection()
             if status.success is not True:
                 return http_error(
                     HTTPStatus.BAD_REQUEST, 'Connection error',
--- a/mindsdb/api/http/namespaces/databases.py
+++ b/mindsdb/api/http/namespaces/databases.py
@@ -170,9 +170,9 @@ class DatabaseResource(Resource):
                 )
             except ImportError as imort_error:
                 return http_error(
-                HTTPStatus.BAD_REQUEST, 'Error',
-                f'Could not create database handler: {str(imort_error)}'
-            )
+                    HTTPStatus.BAD_REQUEST, 'Error',
+                    f'Could not create database handler: {str(imort_error)}'
+                )
 
             status = handler.check_connection()
             if status.success is not True:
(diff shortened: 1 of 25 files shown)


StpMax (changes requested):
You added `try/except` to each usage of `get_handler_module`, except in `create_tmp_handler`. Let move `try/except` inside of `get_handler_module`, and modify  `create_tmp_handler`. So code will be simpler.

MinuraPunchihewa:
> You added `try/except` to each usage of `get_handler_module`, except in `create_tmp_handler`. Let move `try/except` inside of `get_handler_module`, and modify `create_tmp_handler`. So code will be simpler.

@StpMax From what I have understood, the fact that an error is raised here is not an issue; the handler module cannot be imported and therefore, an error should be raised, but the problem right now is that the error is vague (`NoneType' object has no attribute 'Handler`). However, there are other uses of this function that don't necessarily mean that an error should be raised. For instance, in the `process_cache`. This is why I have not included error handling in `create_tmp_handler()`. What do you think?

StpMax:
@MinuraPunchihewa initially `def get_handler_module()` returned `None` if `handler_meta["import"]["success"]` is not True. You changed it, now it raises `ImportError` in this case. But in same time, you added `try/except` like
```python
try:
    handler_module = integration_controller.get_handler_module(integraion_record.engine)
except ImportError:
    handler_module = None
```
to each usage of `get_handler_module`, except one in `create_tmp_handler`. I can't understand why it needed. If rollback your changes, then `handler_module = None` if `handler_meta["import"]["success"]` is not True.

MinuraPunchihewa:
> @MinuraPunchihewa initially `def get_handler_module()` returned `None` if `handler_meta["import"]["success"]` is not True. You changed it, now it raises `ImportError` in this case. But in same time, you added `try/except` like
> 
> ```python
> try:
>     handler_module = integration_controller.get_handler_module(integraion_record.engine)
> except ImportError:
>     handler_module = None
> ```
> 
> to each usage of `get_handler_module`, except one in `create_tmp_handler`. I can't understand why it needed. If rollback your changes, then `handler_module = None` if `handler_meta["import"]["success"]` is not True.

OK, so this is my thought process:
In `_get_integration_record_data()`, for instance, which is the code you've pointed out above, it is acceptable for the result of `get_handler_module()` to be `None`. The code that follows this call will check if the result is `None` and then proceed further.

However, in `create_tmp_handler()`, it is not acceptable for the result of this function to be `None`, because we need to be able to access the `Handler` class of the module in line 453:
`HandlerClass = self.get_handler_module(engine).Handler`

I believe that in situations where this is `None` is when we get the error described in the [issue](https://linear.app/mindsdb/issue/BE-524/please-fix-the-most-annoying-constant-error-in-mindsdb-attributeerror):
`AttributeError: 'NoneType' object has no attribute 'Handler'`

I was thinking that since this error is not very descriptive, it might better to raise an `ImportError` instead.

Does this make sense?

The other way I can think of approaching this is to remove the logic for raising an `ImportError` in `get_handler_module()`, check if the result of calling `get_handler_module()` in `create_tmp_handler()` is `None` (before attempting to access the `Handler`) and raise an error there:
```
handler_module = self.get_handler_module(engine)
if handler_module:
    HandlerClass = handler_module.Handler
else:
    raise ImportError()
```

But, doing so directly in `get_handler_module()` made more sense to me because we have easy access to the `error_message` as well.

Am I missing something here?

StpMax:
@MinuraPunchihewa let me show what i mean: https://github.com/mindsdb/mindsdb/pull/10244
if there is issue only in `create_tmp_handler`, then what is the reason to modify another calls of `get_handler_module`

MinuraPunchihewa:
> @MinuraPunchihewa let me show what i mean: #10244 if there is issue only in `create_tmp_handler`, then what is the reason to modify another calls of `get_handler_module`

@StpMax Got it. This makes sense. Let me merge your PR and take it from there.

StpMax:
@MinuraPunchihewa

> For educational purposes, could you please explain your reasoning for preferring this approach?

Just to make it simpler. Integration controller too overloaded with methods, arguments, execption and so on. 

> My argument is that when a function might result in an exception, it seems more logical to raise the exception directly within that function (following the Single Responsibility Principle).


You mean `get_handler_module`? It never resulted an exception, the only purpouse of the method is return module or None.
If you mean changes in `create_tmp_handler` - then yes, this method raised unexpected exception `'NoneType' object has no attribute 'Handler'`, now it raise it's own `ImportError`.

> Otherwise, the calling code needs to handle the error by performing additional checks to determine its cause—like we're doing now.

Calling code of `get_handler_module`? It should not handle exception, because calling code expect that method return module or None.
If you mean changes in `create_tmp_handler` - it is only to return a relevant error message to the user.
If you about changes in `http/namespaces/config.py` - this need to solve initial task. If not catch exception here, then user will get 500 error. But in fact failed connection check is successful connection check :) so we need to return 200 code with connection status=false.

> This also leads to calling get_handler_meta() twice: once in get_handler_module() and again when checking the import status. Just looking to learn :)
Could you please point to line number, i can't find this place.

mindsdb/api/http/namespaces/databases.py
```diff
@@ -61,7 +61,14 @@ def post(self):
             )
 
         if check_connection:
-            handler = session.integration_controller.create_tmp_handler(name, database['engine'], parameters)
+            try:
+                handler = session.integration_controller.create_tmp_handler(name, database['engine'], parameters)
+            except ImportError as imort_error:
```
StpMax, line 66: typo im**P**ort_error

mindsdb/api/http/namespaces/databases.py
```diff
@@ -61,7 +61,14 @@ def post(self):
...
+                return http_error(
```
StpMax, line 67: Let do here the same as here: https://github.com/mindsdb/mindsdb/pull/10233/files#diff-5383daaa24b7faf0baa480508fd471d5c21b96d8b3e53fa1274930e7c64ce601R156
It won't be correct to return to the user 400 code, because it is not user's error. The best will be return 200 with relevant error message.

StpMax (approved):

MinuraPunchihewa:
> Just to make it simpler. Integration controller too overloaded with methods, arguments, execption and so on.

That makes a lot of sense; I didn't really think about it that way. I was just looking at the individual components that caused the issue.

> You mean get_handler_module? It never resulted an exception, the only purpouse of the method is return module or None.
> If you mean changes in create_tmp_handler - then yes, this method raised unexpected exception 'NoneType' object has no attribute 'Handler', now it raise it's own ImportError.

What I meant is that `get_handler_module()` is responsible for importing and returning the handler module and therefore, it would make more sense (purely from a responsibility standpoint) that it raises an error if there is an issue importing a particular module. As opposed to raising it where it is called. I was thinking that in the future, if we had use `get_handler_module()` for a similar sort of operation, if `get_handler_module()` returns `None`,  we would need to call `get_handler_meta()` again to figure out what is causing it to fail and to raise an error if required.

> Calling code of get_handler_module? It should not handle exception, because calling code expect that method return module or None.
> If you mean changes in create_tmp_handler - it is only to return a relevant error message to the user.
> If you about changes in http/namespaces/config.py - this need to solve initial task. If not catch exception here, then user will get 500 error. But in fact failed connection check is successful connection check :) so we need to return 200 code with connection status=false.

Yes, I was referring to the code that calls `get_handler_module()`. I thought it would be cleaner to raise the error within this function for the reasons mentioned above, rather than having the code that calls it raise it.
That being said, I understand that another approach to go about it is to either the return the module or return `None` if a problem occurs.

> Could you please point to line number, i can't find this place.

So, within `get_handler_module()`, we call `get_handler_meta()` once and now we call it in `create_tmp_handler()` (line 450) to understand the cause of the problem.

mindsdb/api/http/namespaces/databases.py
MinuraPunchihewa: Done.

PR #10256, opened 2024-12-09
docs: files into integrations

martyna-mindsdb:
## Description

[Please include a summary of the change and the issue it solves. ](docs: files into integrations)

Fixes #issue_number

https://linear.app/mindsdb/issue/BE-560/file-upload-supported-files

## Type of change


docs/integrations/files/parquet.mdx
```diff
@@ -0,0 +1,39 @@
+---
+title: Upload Parquet files to MindsDB
+sidebarTitle: Parquet
+---
+
+You can upload Parquet files of up to 500MB in size to MindsDB.
```
ZoranPandovski, line 6: Maybe we should mention that the 500MB is a limit in our SQL Editor? Same for all other file types

ZoranPandovski, line 6: Should we add a sentence and mention that once the file is uploaded it looks like a table inside the files database?

ZoranPandovski (changes requested):
I've added few minor comments. Should we now keep the file upload under SQL https://docs.mindsdb.com/mindsdb_sql/sql/create/file ?

docs/integrations/files/parquet.mdx
martyna-mindsdb: Ok. Now I'm wondering why do we set this limit in the GUI -- is there any reason for that?

I added this to docs:
```
 Note that this limit is imposed when uploading a file via MindsDB SQL Editor. If you [upload a file via APIs](https://docs.mindsdb.com/rest/files/upload), this limit does not persist.
```

martyna-mindsdb: Ok. Added this:
```
Parquet files are stored in the form of a table inside MindsDB.
```

ZoranPandovski (approved):

PR #10275, opened 2024-12-11
docs - update no file size limit

martyna-mindsdb:
## Description

docs - update no file size limit

as discussed with the frontend, there is no file size limit when MindsDB runs locally

Fixes #issue_number

## Type of change


ZoranPandovski (changes requested):
Should we remove the note:

```
Note that this limit is imposed when uploading a file via MindsDB SQL Editor.
```

ZoranPandovski (approved):

PR #10318, opened 2024-12-23
Feat/update otel configs

lucas-koontz:
## Description

This PR updates the OpenTelemetry integration to improve configuration flexibility via environment variables. The changes allow users to define separate endpoints for logging, tracing, and metrics, making it easier to integrate with various OTEL Collectors. It also provides enhanced support for customizing attributes and enabling the SDK in local environments for development and testing purposes.

Fixes #<issue_number>

## Type of change

## Verification Process

- [ ] **Test Location**: Use a local setup with the provided `docker-compose.yml` or any custom OTEL Collector configuration.
- [ ] **Verification Steps**:
  1. Set different endpoints for `OTEL_OTLP_LOGGING_ENDPOINT`, `OTEL_OTLP_TRACING_ENDPOINT`, and `OTEL_OTLP_METRICS_ENDPOINT` in the environment variables.
  2. Start the application using `docker-compose up`.
  3. Validate that logs, metrics, and traces are correctly sent to their respective endpoints.
  4. Test the `OTEL_EXTRA_ATTRIBUTES` configuration by including additional attributes and ensuring they appear in the OTEL Collector's output.

## Additional Media:

- [ ] Attached screenshots or a Loom video showcasing the configuration changes and their effects on metrics, logging, and tracing outputs.

## Checklist:

- [x] My code follows the style guidelines (PEP 8) of MindsDB.


docker-compose.yml
```diff
@@ -54,9 +54,11 @@ services:
       LANGFUSE_TIMEOUT: "10"
       LANGFUSE_SAMPLE_RATE: "1.0"
       OTEL_EXPORTER_TYPE: "console" # or "console" # Define the exporter type (console/otlp)
-      # OTEL_OTLP_ENDPOINT: "http://jaeger:4317" # Define the endpoint for the otlp exporter. Uncomment this line if you want to use jaeger.
+      # OTEL_OTLP_ENDPOINT: "http://otel-collector:4317" # Define the endpoint for the otlp exporter.
```
perry-mitchell, line 57: Worth mentioning in the comment that this is in fact GRPC? Tripped me up the first time but I suppose the port is clear 👍

Just in terms of GRPC/HTTP support.

docker-compose.yml
```diff
@@ -69,21 +71,20 @@ services:
...
+      OTEL_PYTHON_LOGGING_AUTO_INSTRUMENTATION_ENABLED: "true"
+    volumes:
+      - ./otel-collector-config.yaml:/etc/otelcol-contrib/config.yaml
+    ports:
+      - 1888:1888 # pprof extension
+      - 8888:8888 # Prometheus metrics exposed by the Collector
+      - 8889:8889 # Prometheus exporter metrics
+      - 13133:13133 # health_check extension
+      - 4317:4317 # OTLP gRPC receiver
+      - 4318:4318 # OTLP http receiver
```
perry-mitchell, line 86: Are both supported by mindsdb in terms of exporting?

docs/integrations/open_telemetry.mdx
```diff
@@ -0,0 +1,100 @@
...
+- `OTEL_SERVICE_ENVIRONMENT`: Specifies the deployment environment (e.g., `local`, `dev`, `prod`).
+  - Default: `local`.
+
+- `OTEL_SERVICE_RELEASE`: Version or release identifier of the service.
+  - Default: `local`.
+
+### Attributes
+
+- `OTEL_EXTRA_ATTRIBUTES`: Allows passing additional attributes as a comma-separated list of key-value pairs.
+  - Example: `service.version=0.0.1,foo=bar`.
```
perry-mitchell, line 56: Beautiful!

mindsdb/utilities/otel.py
```diff
@@ -1,44 +1,152 @@
...
+    base_attributes = {
+        "service.name": OTEL_SERVICE_NAME,
+        "service.instance.id": OTEL_SERVICE_INSTANCE_ID,
+        "environment": OTEL_SERVICE_ENVIRONMENT,
+        "release": OTEL_SERVICE_RELEASE,
+    }
+
+    extra_attributes = {}
+    try:
+        extra_attributes = parse_csv_attributes(OTEL_EXTRA_ATTRIBUTES)
```
perry-mitchell, line 85: 👍

mindsdb/utilities/otel.py
```diff
@@ -1,44 +1,152 @@
...
+
+# Define OTLP tracing endpoint. If not set, the default OTLP tracing endpoint will be used
+OTEL_OTLP_TRACING_ENDPOINT = os.getenv("OTEL_OTLP_TRACING_ENDPOINT", OTEL_OTLP_ENDPOINT)
+
+# Define OTLP metrics endpoint. If not set, the default OTLP metrics endpoint will be used
+OTEL_OTLP_METRICS_ENDPOINT = os.getenv("OTEL_OTLP_METRICS_ENDPOINT", OTEL_OTLP_ENDPOINT)
+
 # Define service name
-OTEL_SERVICE_NAME = os.getenv("OTEL_SERVICE_NAME", "mindsdb")
+OTEL_SERVICE_NAME = os.getenv("OTEL_SERVICE_NAME", "mindsdb_new_test")
```
perry-mitchell, line 36: Is this intentionally test text?

mindsdb/api/http/namespaces/sql.py
```diff
@@ -24,6 +25,14 @@
 @ns_conf.route("/query")
 @ns_conf.param("query", "Execute query")
 class Query(Resource):
+    meter = otel_metrics.metrics.get_meter("query_service_meter")
+
+    query_counter = meter.create_counter(
+        name="query_post_requests",
```
perry-mitchell, line 31: Is this exported as is, or do you prefix it elsewhere that I've missed? 

Would you consider exporting it, say, as `mindsdb.query_post_requests`? Where `mindsdb` is the service name environment variable.

For instance, you see this a lot with native and some third party exporters: https://opentelemetry.io/docs/specs/semconv/general/metrics/#naming-rules-for-counters-and-updowncounters (so grouping is easier)

mindsdb/utilities/utils.py
```diff
@@ -0,0 +1,34 @@
+import csv
+import re
+import typing
+
+
+def parse_csv_attributes(csv_attributes: str = "") -> typing.Dict[str, str]:
```
ZoranPandovski, line 6: We may mark the param Optional[str]

mindsdb/utilities/utils.py
```diff
@@ -0,0 +1,34 @@
...
+        dict: Parsed key-value pairs as a dictionary.
+    """
+    attributes = {}
+
+    if not csv_attributes:
+        return attributes  # Return empty dictionary if the variable is not set
+
+    try:
+        # Use CSV reader to handle parsing the input
+        reader = csv.reader([csv_attributes])
```
ZoranPandovski, line 21: Is this the same as `str.split(',')`? if so, maybe that is simpler without the need of the reader

docs/integrations/open_telemetry.mdx
```diff
@@ -0,0 +1,100 @@
+---
+title: OpenTelemetry Integration Guide
```
ZoranPandovski, line 2: This is great. I think we should move it to another section in docs, maybe under `setup`. Keeping it in `integrations` may confuse the users with the actual data sources mindsdb works with.

mindsdb/api/http/namespaces/sql.py
```diff
@@ -24,6 +25,14 @@
 @ns_conf.route("/query")
 @ns_conf.param("query", "Execute query")
 class Query(Resource):
+    meter = otel_metrics.metrics.get_meter("query_service_meter")
```
ZoranPandovski, line 28: We need to double check this, I think it will work only when using SQL Editor and Python SDK. If the user is using the `MySQL API` this will not work.

requirements/requirements.txt
```diff
@@ -57,3 +57,4 @@ opentelemetry-sdk==1.27.0
 opentelemetry-exporter-otlp==1.27.0
 opentelemetry-instrumentation-requests==0.48b0
 opentelemetry-instrumentation-flask==0.48b0
+opentelemetry-distro==0.48b0
```
ZoranPandovski, line 60: Should we make the open telemetry optional?

ZoranPandovski (changes requested):
I've added few comments

docker-compose.yml
lucas-koontz: I added a new variable, `OTEL_EXPORTER_PROTOCOL,` and said HTTP support and GRPC.

docker-compose.yml
lucas-koontz: Now it is.

mindsdb/utilities/otel.py
lucas-koontz: Nope. It was from a test and I'm removing now

mindsdb/api/http/namespaces/sql.py
lucas-koontz: There was no prefix; I'm adding it to the code.

lucas-koontz: However, the creation of the metrics does set all the attributes.
It comes from the [resource variable](https://github.com/mindsdb/mindsdb/pull/10318/files#diff-500d0a059249ed0c41ed5793b8ecfae388a8a6ffcc759ed77f9c6c35a788655aR10), which is set [here](https://github.com/mindsdb/mindsdb/pull/10318/files#diff-75b680c4a1a5d165cf0afb3e33d0f84f054d9ca362b094cb4dc7931def504b67R164)

mindsdb/utilities/utils.py
lucas-koontz: Done.

mindsdb/utilities/utils.py
lucas-koontz: Since library csv is also native to python, I don't much the advantage in changing this logic.  It would require to refactor much of this method.

docs/integrations/open_telemetry.mdx
lucas-koontz: Agreed!

mindsdb/api/http/namespaces/sql.py
lucas-koontz: I changed where I added this metric to inside the SQL executor.

requirements/requirements.txt
lucas-koontz: It is optional but we need to install this library still.

docker-compose.yml
perry-mitchell: Amazing, thanks! Sounds like the most versatile approach.

mindsdb/utilities/utils.py
perry-mitchell: CSV is perhaps safer imho as it allows for commas to be used within values due to how CSV allows grouping using parenthesis. All in all probably a more permitting method of storing additional attributes as it'll support simple use cases as well as ones more exotic.

mindsdb/api/http/namespaces/sql.py
perry-mitchell: Having the namespaced metric name, in my experience, is an important aspect of grouping metrics. Usually when someone is viewing or consuming metrics from an application there are a number of emitters. In our case:

- Our applications (sema4ai.something.metric)
- K8s metrics (k8s.something.metric)
- Host metrics (system.)
- Otel metrics (otel.)

It makes filtering my name in UIs a lot easier, and means one can apply filtering rules without having to parse attributes. 

I'd definitely refer to this as namespacing whereas the resource label is a discriminated union. Just something I've seen done elsewhere in Otel logic.

perry-mitchell (commented):
Just one item worth discussing imo: https://github.com/mindsdb/mindsdb/pull/10318#discussion_r1896499533

mindsdb/api/http/namespaces/sql.py
lucas-koontz: Makes sense. I'm updating the logic now.

mindsdb/utilities/otel/__init__.py
```diff
@@ -0,0 +1,198 @@
...
+
+        if OTEL_EXPORTER_PROTOCOL == "grpc":
+            return OTLPLogExporterGRPC(
+                endpoint=OTEL_OTLP_LOGGING_ENDPOINT,
+                insecure=True
+            )
+
+        elif OTEL_EXPORTER_PROTOCOL == "http":
+            return OTLPLogExporterHTTP(
+                endpoint=OTEL_OTLP_LOGGING_ENDPOINT
```
perry-mitchell, line 120: Should this also be insecure, to match GRPC?

mindsdb/utilities/otel/__init__.py
```diff
@@ -0,0 +1,198 @@
...
+
+        if OTEL_EXPORTER_PROTOCOL == "grpc":
+            return OTLPSpanExporterGRPC(
+                endpoint=OTEL_OTLP_TRACING_ENDPOINT,
+                insecure=True
+            )
+
+        elif OTEL_EXPORTER_PROTOCOL == "http":
+            return OTLPSpanExporterHTTP(
+                endpoint=OTEL_OTLP_TRACING_ENDPOINT
```
perry-mitchell, line 144: Same here - `insecure=True`

mindsdb/utilities/otel/__init__.py
```diff
@@ -0,0 +1,198 @@
...
+
+        if OTEL_EXPORTER_PROTOCOL == "grpc":
+            return OTLPMetricExporterGRPC(
+                endpoint=OTEL_OTLP_METRICS_ENDPOINT,
+                insecure=True
+            )
+
+        elif OTEL_EXPORTER_PROTOCOL == "http":
+            return OTLPMetricExporterHTTP(
+                endpoint=OTEL_OTLP_METRICS_ENDPOINT
```
perry-mitchell, line 168: Same here - `insecure=True`

perry-mitchell (approved):
One (set) question regarding the HTTP exporter settings, but otherwise looks great!

mindsdb/utilities/otel/__init__.py
lucas-koontz: The HTTP exporter does not have an insecure parameter.

mindsdb/utilities/otel/__init__.py
lucas-koontz: The HTTP exporter does not have an insecure parameter.

hamishfagg (approved):

PR #10327, opened 2024-12-25
Reduce startup time and memory optimizations

StpMax:
## Description

 - import `boto3` only if it is required by config. This saves memory.
 - replace 3d-part dependency `checksumdir` to buildin `filecmp`. We need only compare dirs, `filecmp` works perfectly for it (and much faster).
 - Indirect import of APIs startup functions. This saves decent amount of memory and also reduces the overall startup time significantly. Local tests shows startup time changes in average for APIs:
   - http: 12.4s->10.5s (-15%)
   - mysql: 9.1->6.5 (-29%)
   - mongodb: 9.0->6.5 (-28%)

## Type of change

## Verification Process

## Additional Media:

## Checklist:

### DIFF
--- a/mindsdb/interfaces/storage/fs.py
+++ b/mindsdb/interfaces/storage/fs.py
@@ -1,6 +1,7 @@
 import os
 import io
 import shutil
+import filecmp
 import tarfile
 import hashlib
 from pathlib import Path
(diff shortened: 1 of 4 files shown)


mindsdb/interfaces/storage/fs.py
```diff
@@ -48,10 +47,41 @@ class RESOURCE_GROUP:
...
+    """Compare two directories
+
+    Args:
+        dir1 (str): dir to compare
+        dir2 (str): dir to compare
+
+    Returns:
+        bool: True if dirs do not differ
+    """
+    dcmp = filecmp.dircmp(dir1, dir2)
```
ea-rus, line 77: do you need to compare a content of files inside? 
As I see, it compares only time of the files by default
https://docs.python.org/3.11/library/filecmp.html#filecmp.cmpfiles

StpMax: `If shallow is true and the [os.stat()](https://docs.python.org/3.11/library/os.html#os.stat) signatures (file type, size, and modification time) of both files are identical, the files are taken to be equal.`
@ea-rus type, time and size, that should be enough

ea-rus (approved):

PR #10340, opened 2025-01-07
updates to extended config docs

### LINKED ISSUE #10295: [Docs]: Changes in app config options
### Short description of what should be added or improved

There are few changes in config options and how config works. Changes implemented in [this PR](https://github.com/mindsdb/mindsdb/pull/10287).

 -  in past, if user edit a config value through app or http request (for example, change username/password), then we edited config that the user had provided to us (with --config cmd arg). Edit user's config file is not a good practice, so we add new file for this cahanges: `config.auto.json`. The file located in `paths.root`. Any changes that app makes to the config will be saved in this file. For reference, here is doc where mentioned [mysql](https://dev.mysql.com/doc/refman/8.4/en/persisted-system-variables.html#persisted-system-variables-file-handling) and [postgres](https://www.postgresql.org/docs/current/config-setting.html) auto configs.
 - config['storage_dir'] is now deprecated, config['paths']['root'] must be used instead.
 - config['log'] is now deprecated. Instead config['logging'] should be used
 - In config['logging'] now possible to configure logging to file
 - Logging may be enabled if set env vars `MINDSDB_CONSOLE_LOG_LEVEL` and/or `MINDSDB_FILE_LOG_LEVEL` to any value of [logging level](https://docs.python.org/3/library/logging.html#logging-levels) ('INFO', 'WARNING', etc)

### Video or screenshots

_No response_

### Anything else?

_No response_

martyna-mindsdb:
## Description

updates to extended config docs

Fixes #10295 

## Type of change

### DIFF
--- a/docs/setup/custom-config.mdx
+++ b/docs/setup/custom-config.mdx
@@ -199,6 +199,14 @@ The `type` parameter specifies the type of caching mechanism to use for storing 
 * `redis`: Stores prediction results in a Redis instance. This option requires the `connection` parameter, which specifies the Redis connection string.
 
 ```bash
+    "logging": {
+        "x": "x"
+    },
+```
+
+todo
+
+```bash
     "ml_task_queue": {
         "type": "local",
         "host": "localhost", # required only when type is set to redis
@@ -234,3 +242,9 @@ If this parameter is left empty (`[]`), users can upload files from any URL with
 The `web_crawling_allowed_sites` parameter restricts web crawling operations to a specified list of allowed IPs or web addresses. This ensures that the application only accesses pre-approved and safe URLs (`"web_crawling_allowed_sites": ["https://example.com", "https://api.mysite.com"]`).
 
 If left empty (`[]`), the application allows access to all URLs by default (marked with a wildcard in the open-source version).
+
+## Modifying Config Values
+
+Users can modify config values by directly editing the `config.json` file.
+
+Users can edit the config values through HTTP requests or MindsDB Editor <todo: how to do that>. In this case, the `config.auto.json` file is created to store the updated `config.json` file. The `config.auto.json` file is stored in the location defined in the `paths.root` config parameter.


StpMax:
@martyna-mindsdb 
 1. When run locally you can change values in config['auth'] with `PUT /api/config` endpoint (I guess you can also change password on 'settings' page in GUI if `http_auth_enabled=True`)
2. Yes, MINDSDB_STORAGE_DIR still in use, `config[paths][root]` get this value
3. Default values in config['logging'] are:
```js
{
    "logging": {
        "handlers": {
            "console": {
                "enabled": True,
                "level": "INFO"     # MINDSDB_CONSOLE_LOG_LEVEL or MINDSDB_LOG_LEVEL (obsolete)
            },
            "file": {
                "enabled": False,
                "level": "INFO",    # MINDSDB_FILE_LOG_LEVEL
                "filename": "app.log",
                "maxBytes": 1 << 19,    # 0.5 Mb
                "backupCount": 3
            }
        }
    }
}
```
I added 'handlers' sub-key to make it similar to [dictSchema](https://docs.python.org/3/library/logging.config.html#logging-config-dictschema). In future we can extend it and add, for example, 'formaters' next to 'handlers'. Everything inside 'console' and 'file', except 'enabled', are the same as in dictSchema. Log level may be controlled by MINDSDB_CONSOLE_LOG_LEVEL or MINDSDB_LOG_LEVEL (first is prefer, i'll delete MINDSDB_LOG_LEVEL sometime in future) and MINDSDB_FILE_LOG_LEVEL env vars

docs/setup/custom-config.mdx
```diff
@@ -198,6 +202,35 @@ The `type` parameter specifies the type of caching mechanism to use for storing
...
+                "level": "INFO", # optional (alternatively, it can be defined in the MINDSDB_FILE_LOG_LEVEL environment variable)
+                "filename": "app.log",
+                "maxBytes": 524288, # 0.5 Mb
+                "backupCount": 3
+            }
+        }
+    },
+```
+
+The above parameters are implemented based on [Python's Dictionary Schema](https://docs.python.org/3/library/logging.config.html#logging-config-dictschema).
```
StpMax, line 225: hm, may be better `Python's logging Dictionary Schema`, or `Logging's Dictionary Schema`

docs/setup/custom-config.mdx
```diff
@@ -234,3 +267,118 @@ If this parameter is left empty (`[]`), users can upload files from any URL with
...
+Modifying config values through MindsDB Editor, provided that `"http_auth_enabled": true`, can be done in Settings.
+
+Modifying config values through an HTTP request:
+
+```bash
+curl --request PUT \
+  --url http://127.0.0.1:47334/api/config \
+  --header 'Content-Type: application/json' \
+  --data '{
+    "key": "value"
```
StpMax, line 381: Therefore, here should be 
```js
{
    "auth": {
        "key": "value"
    }
}
```
May be delete this block? I don''t think that anyone want to modify config via HTTP request. It is mostly for GUI.

docs/setup/custom-config.mdx
```diff
@@ -234,3 +267,118 @@ If this parameter is left empty (`[]`), users can upload files from any URL with
...
+</Accordion>
+
+<Accordion title="Setup">
+
+First, create a `config.json` file.
+
+```bash
+{
+    "permanent_storage": {
+        "location": "local",
```
StpMax, line 280: may be set here `"location": "absent"`?

docs/setup/custom-config.mdx
```diff
@@ -234,3 +267,118 @@ If this parameter is left empty (`[]`), users can upload files from any URL with
...
+        "bucket": "s3_bucket_name"
```
StpMax, line 281: "s3_bucket_name" useless if location is not s3

docs/setup/custom-config.mdx
```diff
@@ -234,3 +267,118 @@ If this parameter is left empty (`[]`), users can upload files from any URL with
...
+    },
+    "paths": {
+        "root": "/home/mindsdb/var",
```
StpMax, line 284: user won't have 'mindsdb' folder. Do we need all following keys in this example?

docs/setup/custom-config.mdx
```diff
@@ -234,3 +267,118 @@ If this parameter is left empty (`[]`), users can upload files from any URL with
...
+        },
+        "mongodb": {
+            "host": "127.0.0.1",
+            "port": "47336",
+            "database": "mindsdb"
+        }
+    },
+    "cache": {
+        "type": "local",
+        "connection": "redis://localhost:6379"
```
StpMax, line 326: it is useless if location is not redis

docs/setup/custom-config.mdx
```diff
@@ -234,3 +267,118 @@ If this parameter is left empty (`[]`), users can upload files from any URL with
...
+                "enabled": false,
+                "level": "INFO",
+                "filename": "app.log",
+                "maxBytes": 524288,
+                "backupCount": 3
+            }
+        }
+    },
+    "ml_task_queue": {
+        "type": "local",
```
StpMax, line 344: following keys are useless with location = 'local'

docs/setup/my_config.json
```diff
@@ -0,0 +1,76 @@
+{
+    "permanent_storage": {
+        "location": "local",
```
StpMax, line 3: The same as above :)

StpMax (changes requested):

docs/setup/custom-config.mdx
```diff
@@ -234,3 +267,118 @@ If this parameter is left empty (`[]`), users can upload files from any URL with
...
+</AccordionGroup>
+
+## Modifying Config Values
+
+Users can modify config values by directly editing the `config.json` file.
+
+Users can edit the config values through an HTTP request or MindsDB Editor. In this case, the `config.auto.json` file is created to store the updated `config.json` file. The `config.auto.json` file is stored in the location defined in the `paths.root` config parameter.
+
+<Tip>
+Modifying config values through MindsDB Editor, provided that `"http_auth_enabled": true`, can be done in Settings.
```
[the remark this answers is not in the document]
martyna-mindsdb: Only from `config['auth']`? So users cannot modify values from `config['api']` for example?

docs/setup/custom-config.mdx
martyna-mindsdb: All following keys are optional. But you're right, I removed them from this example.

martyna-mindsdb: I changed the path to `"root": "/path/to/root/location"`

StpMax (approved):

PR #10353, opened 2025-01-13
Fixed Use of Triggers with the Slack Integration

MinuraPunchihewa:
## Description

This PR fixes the use of triggers with the Slack integration by re-introducing the `table_name` parameter to the `subscribe()` function of the handler.

Fixes https://linear.app/mindsdb/issue/BE-609/fix-execution-of-slack-triggers

## Type of change

## Verification Process

## Checklist:

### DIFF
--- a/docs/mindsdb_sql/sql/create/trigger.mdx
+++ b/docs/mindsdb_sql/sql/create/trigger.mdx
@@ -106,7 +106,7 @@ ON mindsdb_slack.messages
         SELECT 'slack-bot-channel-id' AS channel_id, answer AS text
         FROM chatbot_model m
         JOIN TABLE_DELTA s
-        WHERE  s.user != 'U07J30KPAUF'
+        WHERE  s.user != 'U07J30KPAUF' AND s.channel_id = 'slack-bot-channel-id'
 );
 ```
 
--- a/mindsdb/integrations/handlers/slack_handler/slack_handler.py
+++ b/mindsdb/integrations/handlers/slack_handler/slack_handler.py
@@ -264,7 +264,7 @@ class SlackHandler(APIChatHandler):
         user_info = web_connection.auth_test().data
         return user_info['bot_id']
 
-    def subscribe(self, stop_event: threading.Event, callback: Callable, **kwargs: Any) -> None:
+    def subscribe(self, stop_event: threading.Event, callback: Callable, table_name: Text, **kwargs: Any) -> None:
         """
         Subscribes to the Slack API using the Socket Mode for real-time responses to messages.
 
@@ -274,6 +274,9 @@ class SlackHandler(APIChatHandler):
             table_name (Text): The name of the table to subscribe to.
             kwargs: Arbitrary keyword arguments.
         """
+        if table_name not in ['messages', 'threads']:
+            raise RuntimeError(f'Table {table_name} is not supported for subscription.')
+
         self._socket_connection = SocketModeClient(
             # This app-level token will be used only for establishing a connection.
             app_token=self.connection_data['app_token'],  # xapp-A111-222-xyz


MinuraPunchihewa:
Hey @ea-rus,
I was thinking that it might make sense to check if the columns for the trigger are defined by the user and then raise an error if they have done so? It doesn't really make sense to allow the columns to be defined for Slack triggers, does it? What do you think?

docs/mindsdb_sql/sql/create/trigger.mdx
```diff
@@ -106,7 +106,7 @@ ON mindsdb_slack.messages
         SELECT 'slack-bot-channel-id' AS channel_id, answer AS text
         FROM chatbot_model m
         JOIN TABLE_DELTA s
-        WHERE  s.user != 'U07J30KPAUF'
+        WHERE  s.user != 'U07J30KPAUF' AND s.channel_id = 'slack-bot-channel-id'
```
martyna-mindsdb, line 109: Today I had to use the LAST keyword in order to answer to only the most recent msgs.

We can add it in like this:
```
WHERE  s.user != 'slack-bot-id'
AND s.channel_id = 'slack-bot-channel-id'
AND s.created_at > LAST
```

MinuraPunchihewa: Done!

docs/mindsdb_sql/sql/create/trigger.mdx
```diff
@@ -107,6 +107,8 @@ ON mindsdb_slack.messages
         FROM chatbot_model m
         JOIN TABLE_DELTA s
         WHERE  s.user != 'U07J30KPAUF'
```
martyna-mindsdb, line 109: Shall we update the ID to the placeholder, like this?
```
WHERE s.user != 'slack-bot-id' # this is to prevent the bot from replying to its own messages
```
Because usually we want to prevent the bot from replying to its own messages, that is, to exclude messages sent by the bot.

MinuraPunchihewa: Done!

martyna-mindsdb: thanks!

docs/mindsdb_sql/sql/create/trigger.mdx
```diff
@@ -106,7 +106,9 @@ ON mindsdb_slack.messages
         SELECT 'slack-bot-channel-id' AS channel_id, answer AS text
         FROM chatbot_model m
         JOIN TABLE_DELTA s
-        WHERE  s.user != 'U07J30KPAUF'
+        WHERE s.user != 'slack-bot-id' # this is to prevent the bot from replying to its own messages
+        AND s.channel_id = 'slack-bot-channel-id'
+        AND s.created_at > LAST
```
ea-rus, line 111: I think this condition won't work: `s.created_at > LAST` 
- TABLE_DELTA already has only delta rows. 
- and it is not a real table, it is like a dataframe, it shouldn't work with LAST, as  I know

martyna-mindsdb: @ea-rus 
Without the LAST condition, the bot replies to all messages from the channel every time. When I added this LAST condition, then it started working as expected.
Please see this video - https://www.loom.com/share/19acff65c4484e03a6a086371283c249

martyna-mindsdb: @MinuraPunchihewa 
Let's update it as follows please:
```
JOIN TABLE_DELTA s
WHERE s.user != 'slack-bot-id' # this is to prevent the bot from replying to its own messages
AND s.channel_id = 'slack-bot-channel-id'
```

MinuraPunchihewa: Done!

martyna-mindsdb: Thanks @MinuraPunchihewa - ready to merge.

Thanks @ea-rus for catching it - all works fine with TABLE_DELTA.

ea-rus (approved):

PR #10375, opened 2025-01-16
docs: join-on updates

martyna-mindsdb:
## Description

updates to the join-on doc page

Fixes #issue_number

## Type of change

### DIFF
--- a/docs/mindsdb_sql/sql/api/join-on.mdx
+++ b/docs/mindsdb_sql/sql/api/join-on.mdx
@@ -5,36 +5,22 @@ sidebarTitle: Join Tables On
 
 ## Description
 
-The `JOIN` statement joins two tables `ON` a defined column. It is a regular `JOIN` used throughout SQL.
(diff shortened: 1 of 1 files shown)


MinuraPunchihewa (commented):
Hey @martyna-mindsdb,
This looks good to me, but I have a couple of comments:
1. This is unrelated to your changes, but when viewing the entire file, I noticed that we are talking about the other type of `JOIN` (for models) under the **Nested Queries** section. I was a little confused by this. Do you think that maybe we should move that, maybe to the top under **Description** as a tip or something?
2. Recently, I know we added support for the `ASOF JOIN` syntax. I believe it will be available with the next release. Should we include that here? Or do we take that up in a separate PR?

martyna-mindsdb:
Hey @MinuraPunchihewa 
Thanks for reviewing.

1. There is a separate [doc page that describes JOIN between tables and models](https://docs.mindsdb.com/mindsdb_sql/sql/api/join) - I have linked it in the Tip when mentioning this type of JOIN.
2. Let's make a separate PR once it is ready to test.

MinuraPunchihewa (approved):

PR #10386, opened 2025-01-20
KB: filter by distance

ea-rus:
## Description

Allow to use conditions like `distance` for KB. 

Changes in logic:
- call `dispatch_select` method of vector handler (instead of `query`)
- split conditions into:
  - handled by vector handler
  - applied to results of vector handler

Way to improve it in future:
- Update all vector handlers and make them mark conditions as applied. 
- Then we will know which conditions handler didn't use to apply them outside of handler

Fixes https://linear.app/mindsdb/issue/ML-246/kb-filter-by-distance

## Type of change

## Verification Process

## Additional Media:

## Checklist:


dusvyat:
@ea-rus sounds good - did you do any testing with specific vector db handlers for filtering by vector distance?

ea-rus:
> @ea-rus sounds good - did you do any testing with specific vector db handlers for filtering by vector distance?

tested only for chromadb, can do it for pgvector too

ea-rus:
Pgvector doesn't return `distance` column
image

Also I found that there shouldn't be such column, because select to vector db is specified [here](https://github.com/mindsdb/mindsdb/blob/c9a3835305bc8e322db3050ffab25be601d26461/mindsdb/interfaces/knowledge_base/controller.py#L94) 

It means chromadb returns column which it shouldn't

dusvyat (approved):

PR #10388, opened 2025-01-21
fix: add trace_id

lucas-koontz:
## Description

- Adds a trace ID on each chunk response.
- Fix an issue where `serialized` was being called even when it was null.

Screenshots:

Screenshot 2025-01-21 at 2 18 01 PM
Screenshot 2025-01-21 at 2 18 13 PM


QuantumPlumber (approved):
LGTM

PR #10456, opened 2025-02-12
Updated the Documentation for Using Environment Variables with Docker

MinuraPunchihewa:
## Description

This PR updates the documentation for using the environment variables supported by MindsDB with the use of Docker, which is the main distribution medium.

## Type of change

- [X] 📄 This is a documentation update


MinuraPunchihewa:
@martyna-mindsdb Since Docker is our main mechanism for distributing MindsDB, I was wondering if it would be best to update our documentation for using environment variables based on the above changes?

The current use of the `export` syntax will not work with Docker and I noticed that a few of our community members have faced issues with this. Here is an example: https://mindsdbcommunity.slack.com/archives/C01S2T35H18/p1738682791033309

martyna-mindsdb:
Hey @MinuraPunchihewa 
Thank you for bringing this up.

I agree we should add the Docker syntax for adding env vars. But I'd also keep the PyPI syntax. We can use the CodeGroup element to do that, like this:
```
<CodeGroup>

    ```bash Docker
     docker run --name mindsdb_container -e MINDSDB_USERNAME -e MINDSDB_PASSWORD -p 47334:47334 -p 47335:47335 mindsdb/mindsdb
    ```

    ```bash PyPI
     export MINDSDB_USERNAME='mindsdb_user'
     export MINDSDB_PASSWORD='mindsdb_password'
    ```

</CodeGroup>
```

This looks like this:

https://github.com/user-attachments/assets/7d1ae6f5-8dd3-4aa5-87fc-cea0588bd6a5

MinuraPunchihewa:
> Hey @MinuraPunchihewa Thank you for bringing this up.
> 
> I agree we should add the Docker syntax for adding env vars. But I'd also keep the PyPI syntax. We can use the CodeGroup element to do that, like this:
> 
> ```
> <CodeGroup>
> 
>     ```bash Docker
>      docker run --name mindsdb_container -e MINDSDB_USERNAME -e MINDSDB_PASSWORD -p 47334:47334 -p 47335:47335 mindsdb/mindsdb
>     ```
> 
>     ```bash PyPI
>      export MINDSDB_USERNAME='mindsdb_user'
>      export MINDSDB_PASSWORD='mindsdb_password'
>     ```
> 
> </CodeGroup>
> ```
> 
> This looks like this:
> 
>  Screen.Recording.2025-02-13.at.20.58.43.mov

Got it. That makes sense. I will add these commands in. I think it might be better to include them either the title `Shell` or `Bash` though.

MinuraPunchihewa:
Hey @martyna-mindsdb,
I've made the requested changes. Do you think we need to break the Docker commands into multiple lines since they are quite long?

martyna-mindsdb:
Hey @MinuraPunchihewa 
I previewed it locally and made some fixes and split some docker commands into multiple lines. The CodeGroup element has horizontal scrollbars, so I haven't split the commands in the env vars doc page.

If you find it ok, then it's ready to go. Thank you!

martyna-mindsdb (approved):

MinuraPunchihewa:
> Hey @MinuraPunchihewa I previewed it locally and made some fixes and split some docker commands into multiple lines. The CodeGroup element has horizontal scrollbars, so I haven't split the commands in the env vars doc page.
> 
> If you find it ok, then it's ready to go. Thank you!

Looks good to me, @martyna-mindsdb. Thank you.

PR #10503, opened 2025-02-24
Github handler generated from sdk

ea-rus:
## Description

Github handler tables generated from sdk:
Used inspection of `Repository` object of github sdk:
- to find methods which return tables (based on type annotation of response)
  -  they are converted to handlers tables
- input parameters of this methods are used as filter conditions for tables
- output annotation is used to define list of columns of a handler's table

Fixes https://linear.app/mindsdb/issue/BE-690/github-make-bulletproof

## Type of change

## Verification Process

## Additional Media:

## Checklist:


MinuraPunchihewa (commented):
Hey @ea-rus,
This looks good to me! Do you think there's an opportunity to refactor it at all based on what we did with the handler generator? The data classes look somewhat similar.

ea-rus:
It looks similar because openapi generator was created based on this handler and went its own way. 
If you mean objects names: in this handler another approach is used for generating tables, so are different names. 
Or what exactly did you mean by refactoring?

MinuraPunchihewa:
> It looks similar because openapi generator was created based on this handler and went its own way. If you mean objects names: in this handler another approach is used for generating tables, so are different names. Or what exactly did you mean by refactoring?

@ea-rus I was just thinking that since this is another approach for mapping APIs (in addition to parsing OpenAPI specifications), it would be good if we can standardize as many components as we can? Especially the data classes? I was thinking that `Type` looks somewhat like `APIResourceType` and `GHMethod` could be generalized as well?

MinuraPunchihewa (approved):
@ea-rus Let's look at generalizing later. Meanwhile, I have tested this and it seems to work great.

PR #10517, opened 2025-02-26
Added the Documentation for Changing the Default MindsDB Project

MinuraPunchihewa:
## Description

This PR adds the documentation for changing the default MindsDB project via the `MINDSDB_DEFAULT_PROJECT` environment variable.

Fixes https://linear.app/mindsdb/issue/BE-773/document-default-project-option

## Type of change

- [ ] 📄 This is a documentation update


MinuraPunchihewa:
Hey @ZoranPandovski,
I am leaving this in draft for now, until the PR on the front-end is merged.

MinuraPunchihewa:
Hey @ZoranPandovski,
I just got word that a new frontend release has been made as well. So, I think we can publish this now.

docs/setup/environment-vars.mdx
```diff
@@ -118,4 +118,22 @@ This environment variable defines the level of logging generated by MindsDB. You
     ```bash Shell
     export MINDSDB_LOG_LEVEL='DEBUG'
     ```
+</CodeGroup>
+
+## MindsDB Default Project
+
+By default, MindsDB creates a project named `mindsdb` where all the models and other objects are stored. You can change the default project name by setting the `MINDSDB_DEFAULT_PROJECT` environment variable.
+
+If this environment variable is set or modified after MindsDB has started, the default project will be **renamed** accordingly upon restart. To start using the new default project, a `USE` statement will also need to be executed.
```
martyna-mindsdb, line 127: @MinuraPunchihewa 
If a user sets `MINDSDB_DEFAULT_PROJECT` and then (re)starts MindsDB -- does the user still need to use the `USE` statement?

Also, shall we add a doc page that describes `USE`? That would be [somewhere here](https://docs.mindsdb.com/mindsdb_sql/sql/create/project).

MinuraPunchihewa: @martyna-mindsdb Yes, that is how it was designed. The default project will be renamed on restart, but the user still has to run a `USE` statement to change the context.
Sure, I will add the page.

martyna-mindsdb: Thank you @MinuraPunchihewa -- can you please add a line in `mint.json` to list this new doc page in the navigation?

I am not clear on what is the role of this environment variable. The user still needs to run the `USE project_name` statement, which switches the project with or without this environment variable.

MinuraPunchihewa: @martyna-mindsdb As you already know, the default project that is created with MindsDB at the moment is 'mindsdb'. Say that you did not want this and wanted to call it something else, then you would set this environment variable and run MindsDB. In such a situation where it is set from the get-go, you do have to execute a `USE` statement.

However, if you start MindsDB (without setting the environment variable), run a few queries and then decide to set the environment variable and re-run MindsDB, then you need to run a `USE` statement to change the context. There was quite a bit of debate around if this is how we ought to go about it, but we came to this decision in the end.
CC: @ZoranPandovski

MinuraPunchihewa: @martyna-mindsdb I think I was able to add it to the navigation. Let me know if I have done it right.

martyna-mindsdb: Thanks @MinuraPunchihewa 

To sum up, in order to set this env var, MindsDB needs to be stopped. And then, when starting MindsDB, this env var will be applied. Or, is it applied only before starting MindsDB for the first time? Also, can you please explain it in detail in this env var doc page?

MinuraPunchihewa: It can be done either way,
1. If we set it before MindsDB is started for the first time, the default project will be whatever we set that the environment variable to. The 'mindsdb' project will not be created at all. In this situation, there is no need to change the context or execute a `USE` statement.
2. If we want to change the default project after we have started MindsDB, we will need to stop it, set the variable and restart MindsDB. In this situation, the default project, which will be 'mindsdb' initially, will be **renamed** to the value of the environment variable. In this case only, we will need to execute a  `USE` statement to change the context of the server.

I think I have explained this to some extent in the documentation for this variable. What do you think is missing there?

MinuraPunchihewa: @martyna-mindsdb This is a great explanation of what is accomplished by the logic, but my personal opinion is that for the most part, we want to encourage users to set this at the start (before running MindsDB) rather than at the middle, especially because of the fact that context needs to be changed.

So, I don't know if this is something we should include as an option, but rather as a statement of what would happen if they did it for some reason.
CC: @ZoranPandovski

martyna-mindsdb (approved):

PR #10585, opened 2025-03-20
update table name

chandrevdw31:
update table name  from case to cases as the missing 's' is causing an error


MinuraPunchihewa (commented):
Hey @chandrevdw31,
Can you please explain the need for this? The description in the PR says something about a missing 's'?

All of the resources are already being mapped [here](https://github.com/mindsdb/mindsdb/blob/9a2f6be2fc829be4a45113b5390b5afceed30f99/mindsdb/integrations/handlers/salesforce_handler/salesforce_handler.py#L74).

chandrevdw31:
closing PR, will test with the latest branch

PR #10599, opened 2025-03-25
Run query in partitions

ea-rus:
## Description

Features added:
- run query in partitions (for specific types of queries)
  - option to split batch to separated workers (for speed up prediction)
- be able to track progress of the query
- managing of query in progress:  
  - resume query
  - cancel query
- when query is finished it is kept in queries table (with finished_at time)
  - if user runs another query: all finished queries older than 1 day are removed
- option to skip errors
  - when `error='skip'` is used, errors in partition will be skipped
- show running query info in 'describe knowledge base ...'

**1. How to run queries**

```sql
-- transfer data between tables 
INSERT INTO db2.table2
SELECT * FROM db1.table1
WHERE ws_id = 3
USING batch_size=200, track_column=id

-- add predictor in pipeline
INSERT INTO db2.table2
SELECT * FROM db1.table1
JOIN my_model
USING batch_size=200, track_column=id, threads=true;

-- load knowledge base, skip errors
INSERT INTO my_kb
SELECT * FROM db1.table1
USING batch_size=200, track_column=id, threads=10, error='skip';

-- perform prediction in threads and return data (important: result will be accumulated into one dataset and returned all at once)
SELECT * FROM db1.table1
JOIN my_model
USING batch_size=200, track_column=id, threads=10;
```

Parameters:

- batch_size - count of rows to fetch from database per iteration, optional default 1000
- threads - run partitioning in threads, bool or int, optinal, if set:
   - int value: use this as count of threads
   - true: table threads, autodetect count of thread 
   - false: disable threads even if ml task queue is enabled
- track_column - column used for creating partitions
  - query will be sorted by this column and select will be limited by batch_size
- error (default raise)
  - when `error='skip'` is used, errors in partition will be skipped
```sql
      select * from ({user query})
      where {track_column} > {previous value}
      order by track_column
      limit size {batch_size} `
```

**Speed up prediction**

If predictor is used somewhere in the query or it is load into knowledge base (it means that it also use predictor to generate embeddings) query execution can be optimized by splitting batch into chunk and send them into threads

If  [ML task queue](https://github.com/mindsdb/mindsdb/pull/7437) is enabled, threads will be used automatically. In that case prediction will be performed in different nodes
It is speed up CPU-bound prediction
If prediction uses API services (for example openai), using threads speeds up query execution (even if ML task queue is not enabled)

**2. View queries in progress:**

```sql
select * from information_schema.queries
```
Some of columns in result:
- sql
- error message (if error appeared)
- id of the query
- count of processed rows

**3. Repeat query:**

In case of error, query state will be saved
- current step num (in database)
- and step_data (in cache, it can be files or redis)

User can look at the query error, resolve the issue and then resume query from state it failed

```sql
-- use id of the query
SELECT query_resume(1)
```

**3. Stop query or remove query with error:**

```sql
SELECT query_cancel(1)
```

**DEMOS**

1. Describe knowledge base:
image

2. Partitioning
https://www.loom.com/share/ef37ed3295864bcaaab14c283e11a019

3. Keep finished query stored in 'queries table'
https://www.loom.com/share/8634d29a795b4b6abaab97572a3c7939

**Current limitations**

1. Partitioning will work well with sql datasources, which are supporting nested selects (select from select)
If it is API-handler there could be issues with running partitioning queries
    
2. Not every query can be partitioned
For example, joining between datasources 
```sql
select  * from db1.table1 t1
join db2.table2 t2 on t2.id=t1.id
``` 

3. If track_column has not unique values (for example create_time): data losses can be happened:
When end of batch is in the middle of not unique values, for example 
- batch 1: 1,1,1,2,3,3
- batch 2: 3,3,4,4,4,5
It will memorize max value of the batch as 3
And all values with 3 will be skipped in next batch
        
**Possible future improvements**
    
1. Detach query from editor (send execution to background)
    
2. Merge `select from select` into single select. It will improve work with API handlers
    
3. Use distributed task queue, for example celery
    
4. Update popular data handlers to support steaming results
- instead of 
  - `select ... limit 100` in circle
- use:
  - `yield cursor.fetch(100)`
            

Fixes #issue_number

## Type of change

## Verification Process

## Additional Media:

## Checklist:


mindsdb/api/executor/planner/query_planner.py
```diff
@@ -818,7 +821,64 @@ def from_query(self, query=None):
...
+                    steps_out.append(partition_step)
+                    # mark plan
+                    plan.is_resumable = True
+                    continue
+
+            if partition_step is not None:
+                # check and add step into partition
+
+                can_be_partitioned = False
+                if isinstance(step, (JoinStep, ApplyPredictorStep, InsertToTable)):
```
StpMax, line 854: Better to make `can_be_partitioned` as property of a step

mindsdb/api/executor/sql_query/steps/fetch_dataframe_partition.py
```diff
@@ -0,0 +1,216 @@
...
+
+        # 3 tasks per worker during 1 batch
+        partition_size = int(run_query.batch_size / thread_count / 3)
+        # min partition size
+        if partition_size < 10:
+            partition_size = 10
+
+        # create N workers pool
+        workers = []
+        for i in range(thread_count):
```
StpMax, line 148: will be good to wrap all the following with `try/except/finally: stop_event.set()`

mindsdb/api/executor/sql_query/steps/fetch_dataframe_partition.py
```diff
@@ -0,0 +1,216 @@
...
+                queue_in.put(df2)
+                sent_chunks += 1
+
+            for i in range(sent_chunks):
+                res = queue_out.get()
+                if 'error' in res:
+                    self.close_workers(workers)
+                    raise RuntimeError(res['error'])
+
+                if res['data']:
```
StpMax, line 185: should it have the same order as initial data?

mindsdb/api/executor/planner/query_planner.py
ea-rus: for partitioning fetching from database FetchDataframeStepPartition step is used
also the whole plan have to be marked to handle it in sql_query, it is done by:
`plan.is_resumable = True`

mindsdb/api/executor/sql_query/steps/fetch_dataframe_partition.py
ea-rus: wrapped

mindsdb/api/executor/sql_query/steps/fetch_dataframe_partition.py
ea-rus: it is one of the problem of using threads
we can add restriction, something like: raise exception if query has `order by` and thread are used

StpMax (changes requested):
please add typing and docstrings

mindsdb/api/executor/sql_query/steps/fetch_dataframe_partition.py
StpMax: may be send data to the queue with id of the chunk, and assemble in `concat_results` considering chunks numbers?

tests/unit/executor/test_agent.py
```diff
@@ -466,3 +466,65 @@ def test_kb_metadata(self):
...
+        # unique ids
+        df['id'] = list(map(str, range(len(df))))
+
+        # self.save_file('ral', df.reset_index())
+        self.set_handler(mock_handler, name='pg', tables={'ral': df})
+
+        ret = self.run_sql('''
+            insert into kb_part
+            SELECT id, english content FROM  pg.ral
+            using batch_size=10, track_column=id
```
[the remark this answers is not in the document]
ea-rus: added

mindsdb/api/executor/sql_query/steps/fetch_dataframe_partition.py
ea-rus: implemented in the latest commit

StpMax (approved):

hamishfagg:
@ea-rus I've added the migrations on dev, alpha-dev, and staging

PR #10644, opened 2025-04-02
Fixed Use of Caching in SQL Agent with Amazon S3

MinuraPunchihewa:
## Description

This PR sanitizes the keys used to cache content when using the SQL Agent with Amazon S3. 

Querying S3 (and other file-based integrations) can result in queries like the following:
```
SELECT * FROM s3_datasource_1.`data/iris/iris.csv`
````

Using this table (file) name (s3_datasource_1.`data/iris/iris.csv`) as a key, especially when using the `FileCache` can fail due to the fact that it is treated as a path with intermediary directories and because it contains back-ticks.

Fixes https://linear.app/mindsdb/issue/BE-855/caching-issues-related-to-minds-with-s3

## Type of change

## Verification Process

## Checklist:

### CODE REVIEW

**Review by ZoranPandovski — APPROVED** [review]

File: mindsdb/interfaces/skills/sql_agent.py [hunk 1]
```diff
@@ -251,11 +251,13 @@ def get_table_info(self, table_names: Optional[List[str]] = None) -> str:
         tables_info = []
         for table in all_tables:
             key = f"{ctx.company_id}_{table}_info"
-            table_info = self._cache.get(key) if self._cache else None
+            # Sanitize the key to avoid table (file) names with backticks and slashes.
+            sanitized_key = re.sub(r'[^\w\-.]', '_', key)
```
ea-rus, line 255: As I understand the real problem is located [here](https://github.com/mindsdb/mindsdb/blob/4ea0d3f57fff448a9ce12155dd146c2b26027a1d/mindsdb/utilities/cache.py#L157) - file cache wrongly makes file name, it mustn't create subderictories if key name has `/`
Maybe do this update more globally - in fcache module?

MinuraPunchihewa: @ea-rus Are you suggesting that we add sanitization to the cache module? To replace slashes, back-ticks and the like there? Do you think there can be a situation where someone wants to place the file in an actual path (within directories) somewhere? I guess it is unlikely.

MinuraPunchihewa: @ea-rus I've moved the sanitization to the file cache.

MinuraPunchihewa:
@ZoranPandovski @hamishfagg I've tried to solve this dependency issue a few ways now, but none of them have worked. Any ideas?

ea-rus (approved):

PR #10653, opened 2025-04-03
Added filtering_threshold in the query

ala12326571:
Added filtering_threshold to the query conditions, and it can be tested using this query:

```

SELECT *
FROM kb_custom_model_ada
WHERE content = "Kindle Voyage" AND reranking_threshold =0.6 LIMIT 5;

```

## Description

Fixes #issue_number

## Type of change

## Verification Process

## Additional Media:

## Checklist:

### DIFF
--- a/mindsdb/interfaces/knowledge_base/controller.py
+++ b/mindsdb/interfaces/knowledge_base/controller.py
@@ -35,6 +35,7 @@ from mindsdb.interfaces.knowledge_base.preprocessing.models import Preprocessing
 from mindsdb.interfaces.knowledge_base.preprocessing.document_preprocessor import PreprocessorFactory
 from mindsdb.interfaces.model.functions import PredictorRecordNotFound
 from mindsdb.utilities.exception import EntityExistsError, EntityNotExistsError
+from mindsdb.integrations.utilities.sql_utils import FilterCondition, FilterOperator
 
 from mindsdb.api.executor.command_executor import ExecuteCommands
 from mindsdb.utilities import log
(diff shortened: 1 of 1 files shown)


mindsdb/interfaces/knowledge_base/controller.py
```diff
@@ -701,6 +723,41 @@ def _convert_metadata_value(self, value):
         # Convert everything else to string
         return str(value)
 
+    def _filter_out_threshold_condition(self, where_clause):
```
MinuraPunchihewa, line 726: @ala12326571 Can you explain why this is done this way? There's only a single re-reranking threshold condition, right?

ala12326571: @MinuraPunchihewa 
Because there might be other conditions (with 'AND') and also various spacing schema between conditions (depending on the user's query). Also user might pass threshold value as str or float, ... , and all these make lots of edge cases if want to address it using Regex.
Defining conditions tree is a more robust solution.

MinuraPunchihewa (approved):
LGTM. I think we could definitely improve on how these conditions are being parsed. It looks like we are using the extracted conditions in certain places, but also the where clause directly. But, we can address these later.

mindsdb/interfaces/knowledge_base/controller.py
ea-rus: We don't have to modify AST tree if we already extacted condiitons and can update this structure
Here I create example how to extract conditions once and modify them before passing to vectorstore
https://github.com/mindsdb/mindsdb/pull/10666

Important: Not everything was copied there: content of add_relevance functiton was untoched

PR #10689, opened 2025-04-10
Run integration tests on deployed environment

hamishfagg:
- Remove docker as a requirement for integration tests
- Perform testing against a deployed dev environment in github actions
- Make tests run in parallel 
- Remove as much test inter-dependency as possible


mindsdb-devops (approved):
lgtm

PR #10733, opened 2025-04-17
Fixed DESCRIBE Commands for Objects in Projects

MinuraPunchihewa:
## Description

This PR fixes running `DESCRIBE` queries on objects like Knowledge Bases when the name has been prefixed by the project. For example:
`DESCRIBE KNOWLEDGE_BASE kb_proj.test_proj_kb;`

Fixes https://linear.app/mindsdb/issue/BE-903/fixed-describe-commands-for-knowledge-bases-and-other-objects-in

## Type of change

## Verification Process

## Additional Media:

image

## Checklist:

### DIFF
--- a/mindsdb/api/executor/command_executor.py
+++ b/mindsdb/api/executor/command_executor.py
@@ -869,6 +869,12 @@ class ExecuteCommands:
                 raise WrongArgumentError(f'Unknown describe type: {obj_type}')
 
         parts = obj_name.parts
+        if len(parts) > 2:
+            raise WrongArgumentError(
+                f"Invalid object name: {obj_name.to_string()}.\n"
+                "Only models support three-part namespaces."
+            )
+
         name = parts[-1]
         where = BinaryOperation(op='=', args=[
             Identifier('name'),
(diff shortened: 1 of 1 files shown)


mindsdb/api/executor/command_executor.py
```diff
@@ -868,13 +868,21 @@ def answer_describe_object(self, obj_type: str, obj_name: Identifier, database_n
             else:
                 raise WrongArgumentError(f'Unknown describe type: {obj_type}')
 
-        name = obj_name.parts[-1]
+        parts = obj_name.parts
```
StpMax, line 871: since min python version changed to 3.10, it is possible to use match/case for such cases. I found it more easy to understand in the code.
```python
match obj_name.parts:
    case [name,]:
        database_name = None
    case [database, name]:
        pass
    case _:
        raise Exception('?!')
```

StpMax (approved):

mindsdb/api/executor/command_executor.py
MinuraPunchihewa: @StpMax I agree that this is easier to read. Should we do it like this though?
```
parts = obj_name.parts

match len(parts):
    case 1:
        pass
    case 2:
        database_name = parts[0]
    case _:
        raise WrongArgumentError(
            f"Invalid object name: {obj_name.to_string()}.\n"
            "Only models support three-part namespaces."
        )
```

I was thinking that extracting the `database` might not be very accurate, especially if someone were to pass something like a three-part namespace?

PR #10779, opened 2025-04-30
Updated the TO_MARKDOWN() Function to Use AIPDF

MinuraPunchihewa:
## Description

This PR updates the `TO_MARKDOWN()` function to run via the [aipdf](https://github.com/mindsdb/aipdf) library. This also limits the documents that can be processed to PDF files.

Fixes https://linear.app/mindsdb/issue/BE-943/improve-the-performance-of-the-to-markdown-function

## Type of change

- [X] ⚡ New feature (non-breaking change which adds functionality)

## Verification Process

To ensure the changes are working as expected:

 - [ ]   Test Location: Specify the URL or path for testing.
 - [ ]   Verification Steps: Outline the steps or queries needed to validate the change. Include any data, configurations, or actions required to reproduce or see the new functionality.

## Checklist:

- [X] My code follows the style guidelines(PEP 8) of MindsDB.
- [X] I have appropriately commented on my code, especially in complex areas.
- [ ] Necessary documentation updates are either made or tracked in issues.
- [ ] Relevant unit and integration tests are updated or added.

PR #10795, opened 2025-05-02
Updated the Documentation for the TO_MARKDOWN() Function

MinuraPunchihewa:
## Description

This PR updates the documentation for the `TO_MARKDOWN()` function based on the recent changes that have been made [here](https://github.com/mindsdb/mindsdb/pull/10779).

Fixes https://linear.app/mindsdb/issue/BE-948/update-the-documentation-of-the-to-markdown-function-based-on-recent


MinuraPunchihewa:
@martyna-mindsdb Can you please help me finish up this PR? Will the markdown content that I have included in the table be rendered correctly? If so, how can I fix it?

martyna-mindsdb (approved):

MinuraPunchihewa:
@martyna-mindsdb Since we made IDs optional again, do you think we should remove the use of the HASH() function to add  IDs here?

martyna-mindsdb:
@MinuraPunchihewa 
No -- let's keep it. With KBs, there are 3 ways that users can provide (or not) the id column:

1. They can use one of the existing columns from their data source.
2. They can generate the id column with HASH() or ROW_NUMBER()
3. They can decide not to provide id column and use the default one.

PR #10798, opened 2025-05-05
Added Support for Complex Query Execution in Knowledge Bases

MinuraPunchihewa:
## Description

This PR allows `SELECT` queries with complex clauses such as DISTINCT, GROUP BY etc.

Fixes https://linear.app/mindsdb/issue/PRD-122/distinct-and-group-by-operations

## Type of change

## Verification Process

## Checklist:


dusvyat (approved):

mindsdb/interfaces/knowledge_base/controller.py
```diff
@@ -223,6 +233,11 @@ def select_query(self, query: Select) -> pd.DataFrame:
         # filter by targets
         if requested_kb_columns is not None:
             df = df[requested_kb_columns]
+
+        # apply distinct if needed
+        if is_distinct:
+            df = df.drop_duplicates()
```
ea-rus, line 239: I think it is better to do it with duckdb. Otherwise we have to implement every feature in custom way: distinct, group, order, functions.
If we use [query_df](https://github.com/mindsdb/mindsdb/blob/4ea0d3f57fff448a9ce12155dd146c2b26027a1d/mindsdb/api/executor/utilities/sql.py#L93) function instead we can do better feature coverage. To use it or not - we can check by query like [here](https://github.com/mindsdb/mindsdb/blob/a7526f39ddcf8c0683a8302e938cf0ac4b80c493/mindsdb/api/executor/planner/plan_join.py#L107)

ea-rus (changes requested):

mindsdb/interfaces/knowledge_base/controller.py
MinuraPunchihewa: @ea-rus You are absolutely right. This is much better. I just pushed some changes.

One question I have is if we should restrict the kind of operations that can be performed on certain columns like `chunk_content` or `metadata`, for example?
@dusvyat What do you think?

ea-rus (approved):

PR #10808, opened 2025-05-07
Generate KB row ID if not provided 

dusvyat:
## Description

Either take provided row ID from user on insert or generate a row_id using following format:

` <first 16 char of md5 hash of row content>`

Fixes UNI-25

## Type of change

## Verification Process

verified in SQL UI

## Additional Media:

Screenshot 2025-05-07 at 12 37 16

## Checklist:


dusvyat:
cc @martyna-mindsdb

fshabashev (approved):

mindsdb/interfaces/knowledge_base/controller.py
```diff
@@ -47,7 +47,7 @@
 logger = log.getLogger(__name__)
 
 KB_TO_VECTORDB_COLUMNS = {
-    'id': 'original_row_id',
+    'id': 'original_doc_id',
```
ea-rus, line 50: it makes `id` not the same as original id (from source table). 
after it we can't join KB with source table
How it used to be before: 
image

dusvyat: fair point, I moved content column name to chunk id

martyna-mindsdb: Thanks for catching this @ea-rus 

Here is how it is supposed to work:

1. If user provides id_column when creating KB, then values from this id_column are used for the KB's id, which is not the case atm (@dusvyat FYI -- https://www.loom.com/share/3eaed64b8bb144f19a06974862f341ea)

2. Only if user does not provide id_column, then we use this format of id: <first 16 char of md5 hash of row content>_<content_column_name>

ea-rus (approved):

martyna-mindsdb:
@dusvyat 
The `_<content column>` part was removed from the default id. Shall we bring it back?

When querying KB without id_column provided (this should have hashValue_contentColumnName, but it has only hashValue):
image

</br></br>

When querying KB with id_column provided (this is ok):
image

martyna-mindsdb:
> > The _ part was removed from the default id. Shall we bring it back?
> 
> @martyna-mindsdb, what the reason to have column name in generated id?

In fact, chunk_id contains the content column name. So, let's keep the id column the way it is now.

martyna-mindsdb (approved):

MinuraPunchihewa (approved):

PR #10816, opened 2025-05-08
docs for agents and chatbots updates

martyna-mindsdb:
## Description

docs for agents and chatbots updates

Fixes #issue_number

## Type of change

### DIFF
--- a/docs/mindsdb_sql/agents/agent.mdx
+++ b/docs/mindsdb_sql/agents/agent.mdx
@@ -93,9 +93,9 @@ DROP SKILL my_skill;
 
 ### `provider`
 
-Agents can be assigned different model available from [LangChain integration](/integrations/ai-engines/langchain).
+Agents can be assigned different models available from [LangChain integration](/integrations/ai-engines/langchain).
 
-The `provider` parameter defined the provider of the model. Its available values include `openai`, `anthropic`, `anyscale`, `ollama`, `litellm`, and `mindsdb`.
+The supported providers include `openai`, `anthropic`, `anyscale`, `ollama`, `litellm`, and `mindsdb`.
 
 ### `model`
 
@@ -115,6 +115,10 @@ USING
 The `prompt_template` parameter is used to provide additional instructions to the agent.
 
 Alternatively, this instruction can be provided when creating a model within MindsDB with the `CREATE MODEL` command by assigning this model to an agent.
+
+### `verbose`
+
+The `verbose` parameter accepts two values: `True` to generate a more detailed and expansive response, or `False` to generate a shorter, more concise response.
 
 ### `max_tokens`
 
--- a/docs/mindsdb_sql/agents/agent.mdx
+++ b/docs/mindsdb_sql/agents/agent.mdx
@@ -3,7 +3,7 @@ title: Agent
 sidebarTitle: Agent
 ---
 
-With MindsDB, you can create and deploy AI agents that comprise AI models and customizable skills such as knowledge bases and text-to-SQL.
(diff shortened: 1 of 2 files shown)


docs/mindsdb_sql/agents/agent.mdx
```diff
@@ -3,145 +3,173 @@ title: Agent
...
+    skills = ['text_to_sql_skill'],
+    provider = 'openai',
+    model = 'gpt-4',
+    prompt_template = 'Answer the user input in a helpful way using tools',
+    verbose = True,
+    max_tokens = 100;
+```
+
+<Note>
+Users can define a default model in the [MindsDB configuration under `default_llm`](/setup/custom-config) to be used with agents.
```
ZoranPandovski, line 32: Should we mention this? If users use docker they will need to build new image for this to take effect? Maybe mention that we use this default model

docs/mindsdb_sql/agents/agent.mdx
```diff
@@ -3,145 +3,173 @@ title: Agent
...
 
 ```sql
-SHOW KNOWLEDGE_BASES;
+DROP SKILL my_skill;
+```
 
-SHOW SKILLS;
+### `provider`
+
+Agents can be assigned different model available from [LangChain integration](/integrations/ai-engines/langchain).
```
ZoranPandovski, line 96: models

docs/mindsdb_sql/agents/agent.mdx
```diff
@@ -3,145 +3,173 @@ title: Agent
...
+
+The `provider` parameter defined the provider of the model. Its available values include `openai`, `anthropic`, `anyscale`, `ollama`, `litellm`, and `mindsdb`.
```
ZoranPandovski, line 98: Maybe change this to `The supported providers include `openai`, `anthropic`, `anyscale`, `ollama`, `litellm`, and `mindsdb`.

docs/mindsdb_sql/agents/agent.mdx
```diff
@@ -263,14 +259,23 @@ USING
...
-UPDATE AGENT text_to_sql_agent
-SET
-    skills_to_add = ['kb_skill'];
+CREATE AGENT my_agent
+USING 
+    skills = ['kb_skill'],
+    provider = 'openai',
+    model = 'gpt-4',
+    prompt_template = 'Answer the user input in a helpful way using tools',
+    verbose = True,
```
ZoranPandovski, line 271: Not required but we may need to document the `verbose` too

ZoranPandovski (changes requested):
Few minor comments

docs/mindsdb_sql/agents/agent.mdx
martyna-mindsdb: @MinuraPunchihewa has implemented the default model such that it must be passed into config.json. I can add info that users need to rebuild Docker image. But otherwise, they cannot use default models if these are not defined.

@MinuraPunchihewa 
Is there any easier way to do it for MindsDB in Docker?

MinuraPunchihewa: @martyna-mindsdb I believe there is going to be a GUI page created for setting the default models? They can also do it by calling the back-end API (PUT /api/config) directly.

martyna-mindsdb: @MinuraPunchihewa 
Can you please give an example of how users would call this endpoint specifically to add default models? Something like:
```
curl PUT /api/config
...
{ default_model: ...}
```

PR #10839, opened 2025-05-15
Added Minor Improvements to the Agents Execution Flow

MinuraPunchihewa:
## Description

Fixes #issue_number

## Type of change

## Verification Process

## Additional Media:

## Checklist:

### DIFF
--- a/mindsdb/interfaces/skills/sql_agent.py
+++ b/mindsdb/interfaces/skills/sql_agent.py
@@ -232,8 +232,18 @@ class SQLAgent:
             table_identifier = tables_idx.get(tuple(table_parts))
 
             if table_identifier is None:
-                # If the entire table name is enclosed in backticks, raise an error by mentioning this.
-                if len(table_parts) == 1 and (table_name.startswith('`') and table_name.endswith('`') and table_name.count('`') == 2):
+                # If the full table name (along with the database and schema) is enclosed in backticks
+                # raise an error for the agent to correct it
+                supported_file_formats = ['csv', 'tsv', 'json', 'parquet']
+                if (
+                    table_name.startswith('`')
+                    and table_name.endswith('`')
+                    and table_name.count('`') == 2
+                    and len(table_parts) == 1
+                    # Ensure the table name has at least one dot and it is not in relation to a file e.g. `file.csv`
+                    and '.' in table_name
+                    and table_name.split('.')[-1] not in supported_file_formats
+                ):
                     raise ValueError(
                         "Only the table name should be enclosed in backticks.\n"
                         "Do not include the database or schema inside the backticks."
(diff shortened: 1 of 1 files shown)


mindsdb/interfaces/skills/sql_agent.py
```diff
@@ -232,7 +232,23 @@ def _resolve_table_names(self, table_names: List[str], all_tables: List[Identifi
...
+                    and table_name.count('`') == 2
+                    and len(table_parts) == 1
+                    # Ensure the table name has at least one dot and it is not in relation to a file e.g. `file.csv`
+                    and '.' in table_name
+                    and table_name.split('.')[-1] not in supported_file_formats
+                ):
+                    raise ValueError(
+                        "Only the table name should be enclosed in backticks.\n"
+                        "Do not include the database or schema inside the backticks."
+                    )
```
ea-rus, line 250: why it is prohibited to add backticks to database or schema?

MinuraPunchihewa: @ea-rus In many situations, the LLM returns full database name in the following format:
```
`database_name.table_name`
```

That is, with the database + table name within back-ticks. We are not handling such strings and it is difficult to because we can't exactly split by the periods (.) in case there is something like the following: 
```
`s3_datasource.iris.csv`
```

This results in agent seeing the error that the table cannot be found, which is not correct.
Do you have an idea for how else this could be handled?

ea-rus: I think if we remove databases from tables names it helps llm to compose better queries:
First step: map tables to simplified names for llm
- `database_name.table_name` -> table_name
- `s3_datasource.iris.csv` -> iris
Say LLM that we have tables: table_name, iris
After llm compose query: replace names to original ones

MinuraPunchihewa: @ea-rus There are several places where the agent needs the database name. For instance, [here](https://github.com/mindsdb/mindsdb/blob/8f78eaa9155d13ae14d99e7fd0a662ba69b3faa4/mindsdb/interfaces/skills/sql_agent.py#L316) to retrieve sample rows and even within [caching](https://github.com/mindsdb/mindsdb/blob/8f78eaa9155d13ae14d99e7fd0a662ba69b3faa4/mindsdb/interfaces/skills/sql_agent.py#L254) process, I feel, because if we remove the table name and other data sources contain tables with the same name, this can be a problem.

MinuraPunchihewa: @ea-rus Just to understand this better; are you suggesting that we change the description of the tools, so that table names generated by the agent don't include database names? Or are we going to keep it as is and remove the database name after the agent generates it?

martyna-mindsdb:
@ea-rus @MinuraPunchihewa 
Is this PR still in progress?

MinuraPunchihewa:
@martyna-mindsdb I think there was a different line of work done on handling the issues addressed here by Andrey. We can close this.

PR #10848, opened 2025-05-16
Enabled Default Re-ranking Model for Knowledge Bases

MinuraPunchihewa:
## Description

This PR enables users to define a default re-ranking model (`default_reranking_model`) in the configuration for use when creating knowledge bases.

Fixes https://linear.app/mindsdb/issue/UNI-76/add-default-model-configuration-for-re-ranking

## Type of change

## Verification Process

## Checklist:

### DIFF
--- a/mindsdb/api/http/namespaces/config.py
+++ b/mindsdb/api/http/namespaces/config.py
@@ -32,7 +32,7 @@ class GetConfig(Resource):
                 'http_auth_enabled': config['auth']['http_auth_enabled']
             }
         }
-        for key in ['default_llm', 'default_embedding_model']:
+        for key in ['default_llm', 'default_embedding_model', 'default_reranking_model']:
             value = config.get(key)
             if value is not None:
                 resp[key] = value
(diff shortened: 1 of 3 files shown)


mindsdb/api/http/namespaces/config.py
```diff
@@ -32,7 +32,7 @@ def get(self):
                 'http_auth_enabled': config['auth']['http_auth_enabled']
             }
         }
-        for key in ['default_llm', 'default_embedding_model']:
+        for key in ['default_llm', 'default_embedding_model', 'default_reranking_model']:
```
ea-rus, line 35: it is for getting reranking llm from 'default_reranking_model' instead of 'default_llm' and 'default_llm' will be use only by functions, right?

ea-rus (approved):

mindsdb/api/http/namespaces/config.py
MinuraPunchihewa: I think the idea is that it is possible that the user might want to use a different model for re-ranking. That is what I understood, but I will double-check with Ale.

MinuraPunchihewa: @ea-rus You were right about this. I spoke to Ale and he said that only the default reranking model will be used for re-ranking. Can you please have a quick look and we can merge this?

PR #10918, opened 2025-05-27
Don't send slack messages for cancelled workflow runs

hamishfagg:
Cancelled runs are usually due to a concurrent run having priority, so we don't need to alert for it being cancelled.


mindsdb-devops (approved):

mindsdb-devops (approved):
lgtm

PR #10986, opened 2025-06-02
kb docs - create index on kb

martyna-mindsdb:
## Description

kb docs - create index on kb

image

Fixes #issue_number

## Type of change

### DIFF
--- a/docs/mindsdb_sql/knowledge-bases.mdx
+++ b/docs/mindsdb_sql/knowledge-bases.mdx
@@ -678,7 +678,7 @@ CREATE INDEX ON KNOWLEDGE_BASE my_kb;
 ```
 
 <Warning>
-Note that this feature works only when PGVector is used as the [storage of the knowledge base](/mindsdb_sql/knowledge-bases#storage).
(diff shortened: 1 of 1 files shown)


docs/mindsdb_sql/knowledge-bases.mdx
```diff
@@ -669,6 +669,22 @@ FROM sample_data.orders;
...
+```
+
+<Warning>
+Note that this feature works only when PGVector is used as the [storage of the knowledge base](/mindsdb_sql/knowledge-bases#storage).
+</Warning>
+
+Upon executing this statement, an index is created on the knowledge base's underlying vector store. This is essentially a database index created on the vector database.
+
+Note that having an index on the knowledge base may reduce the speed of the insert operations. Therefore, it is recommended to insert bulk data into the knowledge base before creating an index. The index improves performance of querying the knowledge base, while it may slow down subsequent data inserts.
+
```
fshabashev: looks good, but let's clarify that is it only for postgres (pgvector database)

martyna-mindsdb: @fshabashev 
It is already clarified in the warning block. I added a note about ChromaDB.

PR #10997, opened 2025-06-03
Add support to agent for default LLM parameters from config

dusvyat:
## Description

Updated agents_controller to utilize `default_llm` from config file or via UI config when model or provider is not explicitly provided. This is done at run time and not on agent creation.

Introduced unit tests to validate the behavior, ensuring defaults are properly applied and can be overridden when specified.

see config docs: https://docs.mindsdb.com/setup/custom-config#overview-of-config-parameters

Supports minimalist syntax:

```sql
CREATE AGENT amazon_reviews_agent 
USING 
 include_tables = ['files.enriched_reviews', 'files.enriched_products'];


```

closes UNI-109

## Type of change

## Verification Process

## Additional Media:
https://www.loom.com/share/778ec540839e450aa7bd79cbbc33a1e5

## Checklist:

### DIFF
--- a/mindsdb/utilities/render/sqlalchemy_render.py
+++ b/mindsdb/utilities/render/sqlalchemy_render.py
@@ -14,24 +14,22 @@ from sqlalchemy.sql import functions as sa_fnc
 from mindsdb_sql_parser import ast
 
 
-RESERVED_WORDS = {
-    "collation"
-}
+RESERVED_WORDS = {"collation"}
 
 sa_type_names = [
-    key for key, val in sa.types.__dict__.items() if hasattr(val, '__module__')
-    and val.__module__ in ('sqlalchemy.sql.sqltypes', 'sqlalchemy.sql.type_api')
+    key
+    for key, val in sa.types.__dict__.items()
+    if hasattr(val, "__module__") and val.__module__ in ("sqlalchemy.sql.sqltypes", "sqlalchemy.sql.type_api")
 ]
 
 types_map = {}
 for type_name in sa_type_names:
     types_map[type_name.upper()] = getattr(sa.types, type_name)
-types_map['BOOL'] = types_map['BOOLEAN']
-types_map['DEC'] = types_map['DECIMAL']
+types_map["BOOL"] = types_map["BOOLEAN"]
+types_map["DEC"] = types_map["DECIMAL"]
 
 
-class RenderError(Exception):
-    ...
+class RenderError(Exception): ...
 
 
 # https://github.com/sqlalchemy/sqlalchemy/discussions/9483?sort=old#discussioncomment-5312979
@@ -43,13 +41,13 @@ class INTERVAL(ColumnElement):
 
 @compiles(INTERVAL)
 def _compile_interval(element, compiler, **kw):
-    items = element.info.split(' ', maxsplit=1)
-    if compiler.dialect.name == 'oracle' and len(items) == 2:
+    items = element.info.split(" ", maxsplit=1)
+    if compiler.dialect.name == "oracle" and len(items) == 2:
         # replace to singular names (remove leading S if exists)
-        if items[1].upper().endswith('S'):
+        if items[1].upper().endswith("S"):
             items[1] = items[1][:-1]
 
-    if compiler.dialect.driver in ['snowflake']:
+    if compiler.dialect.driver in ["snowflake"]:
         # quote all
         args = " ".join(map(str, items))
         args = f"'{args}'"
@@ -62,8 +60,9 @@ def _compile_interval(element, compiler, **kw):
 
 class AttributedStr(str):
     """
-       Custom str-like object to pass it to `_requires_quotes` method with `is_quoted` flag
+    Custom str-like object to pass it to `_requires_quotes` method with `is_quoted` flag
     """
+
     def __new__(cls, string, is_quoted: bool):
         obj = str.__new__(cls, string)
         obj.is_quoted = is_quoted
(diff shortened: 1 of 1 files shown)


martyna-mindsdb:
@dusvyat 
I tried this out in the editor, and I get this error re API key when querying the agent:

https://www.loom.com/share/4507a67b0fbc4ecc88223ceef3351f95

dusvyat:
> @dusvyat I tried this out in the editor, and I get this error re API key when querying the agent:
> 
> https://www.loom.com/share/4507a67b0fbc4ecc88223ceef3351f95
@martyna-mindsdb  please try again for me it works

mindsdb/interfaces/agents/agents_controller.py
```diff
@@ -180,8 +180,34 @@ def add_agent(
         if agent is not None:
             raise ValueError(f"Agent with name already exists: {name}")
 
+        # Get default LLM parameters from config if they exist
+        default_llm_params = config.get("default_llm", {})
```
MinuraPunchihewa: Hey @dusvyat,
Currently, when using Knowledge Bases, we allow users to set the default model parameters, but they have the option of [overriding them](https://github.com/mindsdb/mindsdb/blob/f12e82a4a6a883b9fad40c9e4a87bfe2f10a6f5d/mindsdb/interfaces/knowledge_base/controller.py#L58) by explicitly passing them with the CREATE statement. I think it would be best if we allowed the same for agents?

dusvyat: yes this is how it works here too, see the lines below this one.

Basically create params will override default_llm config here as well

MinuraPunchihewa: Yes, sorry I missed that, but why does the provider need to be the same? I think there be a situation where the default model is set to given provider, but the user wants to try an entirely new provider when creating a particular agent?

I think we can simply override the parameters in the default config with what is explicitly passed here.

MinuraPunchihewa (commented):
@dusvyat I don't think we should store the default LLM parameters as part of the record for the agent in the database.

We should just access them separately and use them at each point the agent is accessed.

The reason we should not store them in the database is as follows; imagine that we create an agent after setting the default model and we store those parameters in the database. Then, at the point of executing the agent, we want to try out a different and therefore, change the values of the default configuration. These will not take because they are not accessed when the agent is called.

I think how default models are handled in KBs will be a good reference point.

dusvyat:
> @dusvyat I don't think we should store the default LLM parameters as part of the record for the agent in the database.
> 
> We should just access them separately and use them at each point the agent is accessed.
> 
> The reason we should not store them in the database is as follows; imagine that we create an agent after setting the default model and we store those parameters in the database. Then, at the point of executing the agent, we want to try out a different and therefore, change the values of the default configuration. These will not take because they are not accessed when the agent is called.
> 
> I think how default models are handled in KBs will be a good reference point.

got it I didn't realise this was the design decision, fixed

mindsdb/interfaces/agents/agents_controller.py
```diff
@@ -180,8 +179,30 @@ def add_agent(
...
+            provider = default_llm_params.get("provider")
+
+        # Check if model_name is still None after trying to get it from default_llm
+        if model_name is None:
+            raise ValueError("Model name must be provided either directly or through default_llm configuration")
+
         _, provider = self.check_model_provider(model_name, provider)
 
+        # Initialize params with a copy to avoid modifying the input
+        params = (params or {}).copy()
```
MinuraPunchihewa: Do we need this line? The copy is just overriding the original variable. It should either be changed to a different variable name or just removed.

mindsdb/interfaces/agents/langchain_agent.py
```diff
@@ -239,14 +239,31 @@ def __init__(self, agent: db.Agents, model: dict = None):
...
+
+        Args:
+            params: Parameters for agent execution (already merged with defaults)
+
+        Returns:
+            dict: Final parameters for agent execution
+        """
+        # Start with parameters passed to the method (already merged with defaults)
+        args = params.copy() if params else {}
+        if not args and self.agent.params:
```
MinuraPunchihewa: We are already accounting for the agent parameters in `get_agent_llm_params()`, yes? I don't see a reason to do this again here?

mindsdb/interfaces/agents/langchain_agent.py
```diff
@@ -226,7 +226,7 @@ def process_chunk(chunk):
 
 
 class LangchainAgent:
```
MinuraPunchihewa, line 228: I see that `MCPLangchainAgent` inherits from this class. Will introducing this new parameter (params) have an effect there?

dusvyat: should be resolved ptal

mindsdb/interfaces/agents/mcp_client_agent.py
```diff
@@ -245,8 +250,11 @@ def create_mcp_agent(agent_name: str, project_name: str, mcp_host: str = "127.0.
     if agent_db is None:
         raise ValueError(f"Agent {agent_name} not found in project {project_name}")
 
-    # Create MCP agent
-    mcp_agent = MCPLangchainAgent(agent_db, mcp_host=mcp_host, mcp_port=mcp_port)
+    # Get merged parameters (defaults + agent params)
+    merged_params = agent_controller.get_agent_llm_params(agent_db.params)
```
dusvyat, line 254: @MinuraPunchihewa to handle default llm config for mcpagent

PR #11014, opened 2025-06-04
Enable Data Catalog Support for the New Agent Implementation

MinuraPunchihewa:
## Description

This PR enables the new implementation of agents to work with the data catalog, i.e., to extract table information from the catalog for generating SQL queries.

This also resolves the issues given below by accounting for the include_tables parameter that is now passed when creating a new agent and also for tables from multiple data sources. 

Fixes https://linear.app/mindsdb/issue/BE-1044/setting-specific-tables-fails-with-new-agent-syntax
Fixes https://linear.app/mindsdb/issue/BE-1048/data-catalogue-from-multiple-datasources

## Type of change

## Verification Process

## Checklist:


mindsdb/interfaces/skills/skills_controller.py
```diff
@@ -96,11 +100,38 @@ def add_skill(self, name: str, project_name: str, type: str, params: Dict[str, s
...
-            data_catalog_loader = DataCatalogLoader(
-                database_name=params["database"], table_names=params["tables"] if "tables" in params else None
-            )
-            data_catalog_loader.load_metadata()
+        if config.get("data_catalog", {}).get("enabled", False):
+            if type == SkillType.TEXT2SQL.value:
+                # TODO: Is it possible to create a skill with complete access to the database with the new agent syntax?
+                # TODO: Handle the case where `ignore_tables` is provided. Is this a valid parameter?
+                # TODO: Knowledge Bases?
+                if "include_tables" not in params:
```
dusvyat: is there a plan to also make data_catalog also include knowledge bases? so shouldn't we also consider `include_knowledge_base` param?

dusvyat (approved):
one question otherwise looks good

mindsdb/interfaces/skills/skills_controller.py
MinuraPunchihewa: Yes, we need to at some point, but I think we will be tackling that later.

PR #11132, opened 2025-06-19
KB algebra

ea-rus:
## Description

Implementation of KB algebra. 

**Part 1. Processing 'Content' column conditions.** 
 
Step 1: remove unary NOT from condition related to 'Content' column and try to apply it to related operator.
Supported cases:
- `NOT content <op> value` => `content <!op> value`
- `content <op> NOT value` => `content <!op> value`

Where `<!op>` is inverted operator of `<op>` (for example `LIKE` => `NOT LIKE`)

Step 2: Flat conditions tree. 
Group same operators from the same levels into the same condition blocks:
Examples
`OR (a=1 OR (b=2, c=3))` => `Block(OR, [a=1, b=2, c=3])`

`OR (a=1 OR (b=2, AND(c=3, AND(d=4, e=5))))` => `Block(OR, [a=1, b=2, Block(AND, [c=3, d=4, e=5])])`

Step 3: Recursively execute condition blocks
In OR block:
- execute all items independently and UNION them

In AND block:
- split conditions into 'content' and 'other' conditions
- in 'content' conditions find negative conditions (e.g. '!=', 'NOT IN')
  - convert into positive and call KB with this condition and limit=N  (currently is 100) 
     - it gives sample of IDs with size N that we have to exclude
     - reranking is disabled for getting this sample
     - added filter by relevance to exclude not relate items, current value is 0.5
  - convert them into conditions like: ID NOT (list)
  - add into  'other' conditions
- for every positive condition:
  - simplify it to `content='value' ` and call KB together with 'other' conditions
- INTERSECT all items of data received in AND block

**Part 2. Processing metadata columns**

It is done using verctordb handler logic

Known issues:
- PGVector 
   - stores metadata in json columns, even if they store as int/float - the selected value is always string.
      - workaround: get data type from filtered value and cast json column
   - date, datetime are stored as formates string. but it allows to filter by this column
- Chromadb 
   - limitation on using operators:
      - like, not like are not supported
      - '>=', '>', '<=', '<' - support only int and float types (not string)
   - no option to limit by ids and embedding at the same time:
      - workaround - use embedding filter with greater limit and filter ids after it

**Part 3. Post process query**

Post processing: if query has other cluases than just WHERE (for example GROUP, ORDER):
- Use duckdb to apply addition sql features

Fixes https://linear.app/mindsdb/issue/UNI-105/sql-algebra-for-filtering-metadata-and-content-columns
Fixes https://linear.app/mindsdb/issue/UNI-133/is-not-null-not-working-on-metadata-fields

## Type of change

## Verification Process

## Additional Media:

## Checklist:


mindsdb/interfaces/knowledge_base/executor.py
```diff
@@ -0,0 +1,309 @@
...
+            if isinstance(node.args[0], Identifier):
+                parts = node.args[0].parts
+                if len(parts) == 1 and parts[0].lower() == self.content_column:
+                    return True
+        return False
+
+    @staticmethod
+    def invert_content_op(node: BinaryOperation) -> BinaryOperation:
+        # Change operator of binary operation to opposite one
+        op_map = {"=": "!=", "!=": "=", "LIKE": "!=", "NOT LIKE": "=", "IN": "NOT IN"}
```
StpMax: should {"NOT IN": "IN"}  be here?

mindsdb/interfaces/knowledge_base/executor.py
```diff
@@ -0,0 +1,309 @@
...
+            return results[0]
+
+        item = results[0]
+        for item2 in results[1:]:
+            item = item[item[self.id_column].isin(item2[self.id_column])]
+
+        df = item
+        return df
+
+    def flatten_conditions(self, node: ASTNode) -> Union[ASTNode, ConditionBlock]:
```
StpMax: it should be classmethod, or separate function

StpMax (commented):
tests?

mindsdb/interfaces/knowledge_base/executor.py
ea-rus: changed

mindsdb/interfaces/knowledge_base/executor.py
ea-rus: added

ea-rus:
@StpMax, added set of tests

StpMax (approved):

PR #11138, opened 2025-06-20
Don't replace backticks

ZoranPandovski:
## Description

This PR removes the logic that replaces backticks. Note that this bug is reproducible when the data catalog is not enabled; that's why it worked on the cloud.

Fixes https://linear.app/mindsdb/issue/BE-1149/querys-cant-be-parsed

## Type of change

## Checklist:

### DIFF
--- a/mindsdb/interfaces/skills/sql_agent.py
+++ b/mindsdb/interfaces/skills/sql_agent.py
@@ -403,8 +403,6 @@ class SQLAgent:
         if config.get("data_catalog", {}).get("enabled", False):
             database_table_map = {}
             for name in table_names or self.get_usable_table_names():
-                name = name.replace("`", "")
-
                 parts = name.split(".", 1)
                 # TODO: Will there be situations where parts has more than 2 elements? Like a schema?
                 # This is unlikely given that we default to a single schema per database.


MinuraPunchihewa (commented):
@ZoranPandovski Can you give me an example of what kind of failures this causes? If these back-ticks around the names of tables are introduced by the agent, the data catalog will not be able to handle it, i.e., because the actual table name does not contain back-ticks.

ZoranPandovski:
No, actually, this happens if the data catalogue is not enabled, which means that we are using https://github.com/mindsdb/mindsdb/blob/ed76795e960ecb8837cc5aaeee19d4e822c7d9c2/mindsdb/interfaces/skills/sql_agent.py#L422.

MinuraPunchihewa:
> No, actually, this happens if the data catalogue is not enabled, which means that we are using
> 
> https://github.com/mindsdb/mindsdb/blob/ed76795e960ecb8837cc5aaeee19d4e822c7d9c2/mindsdb/interfaces/skills/sql_agent.py#L422
> 
> .

Understood, but the line has been removed in block where the data catalog is enabled?

PR #11173, opened 2025-06-25
Fix `InterfaceError` for Snowflake Data Catalog stats

ZoranPandovski:
## Description

Cast `min_val` and `max_val` to supported types before saving.  This ensures that all values being inserted into the database are of a type that SQLite can handle, preventing the InterfaceError.

Fixes https://linear.app/mindsdb/issue/BE-1166/fix-snowflake-dc-stats

## Type of change

## Checklist:

### DIFF
--- a/mindsdb/interfaces/data_catalog/data_catalog_loader.py
+++ b/mindsdb/interfaces/data_catalog/data_catalog_loader.py
@@ -1,9 +1,7 @@
 from typing import List, Union
 import pandas as pd
 import json
-from decimal import Decimal
 import datetime
-import uuid
 from mindsdb.integrations.libs.response import RESPONSE_TYPE
 from mindsdb.interfaces.data_catalog.base_data_catalog import BaseDataCatalog
 from mindsdb.interfaces.storage import db
@@ -384,14 +382,8 @@ class DataCatalogLoader(BaseDataCatalog):
         """
         if val is None:
             return None
-        if isinstance(val, (int, float, str, bool)):
-            return str(val)
-        if isinstance(val, Decimal):
-            return str(val)
         if isinstance(val, (datetime.datetime, datetime.date)):
             return val.isoformat()
-        if isinstance(val, uuid.UUID):
-            return str(val)
         if isinstance(val, (list, dict, set, tuple)):
             return json.dumps(val, default=str)
         return str(val)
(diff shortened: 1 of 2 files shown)


MinuraPunchihewa (commented):
@ZoranPandovski This looks good to me, but do you think that maybe we should handle this more globally, so that this error does not occur with other integrations either?
We've made a few such fixes already like this [one](https://github.com/mindsdb/mindsdb/blob/4330dea3f351bbb33f1187e44cd38ad7035299e7/mindsdb/interfaces/data_catalog/data_catalog_loader.py#L206).
Or do you think that this is specific to Snowflake?

ZoranPandovski:
@MinuraPunchihewa 100%, I thought about this but I was working on another thing in parallel so didn't improve this :) Now fixed

mindsdb/interfaces/data_catalog/data_catalog_loader.py
```diff
@@ -204,6 +204,12 @@ def _add_column_statistics(self, df: pd.DataFrame, tables: db.MetaTables, column
                 # Convert the distinct_values_count to an integer if it is not NaN, otherwise set it to None.
                 val = row.get("distinct_values_count")
                 distinct_values_count = int(val) if pd.notna(val) else None
+                min_val = row.get("minimum_value")
+                max_val = row.get("maximum_value")
+                if min_val is not None and not isinstance(min_val, (str, int, float)):
```
MinuraPunchihewa: @ZoranPandovski I understand that storing int and float values in a string column will be allowed in SQLite, but I don't think will work in Postgres?

MinuraPunchihewa:
> @MinuraPunchihewa 100%, I thought about this but I was working on another thing in parallel so didn't improve this :) Now fixed

Awesome. Thanks, @ZoranPandovski. There was one other concern that I left a comment regarding.

mindsdb/interfaces/data_catalog/data_catalog_loader.py
```diff
@@ -373,3 +377,21 @@ def unload_metadata(self) -> None:
             db.session.delete(table)
         db.session.commit()
         self.logger.info(f"Metadata for {self.database_name} removed successfully.")
+
+    def to_str(self, val) -> str:
```
MinuraPunchihewa, line 379: @ZoranPandovski I think we might be able to simply this as follows?
```
def to_str(self, val) -> str:
    """
    Convert a value to a string representation.
    """
    if val is None:
        return None
    if isinstance(val, (datetime.datetime, datetime.date)):
        return val.isoformat()
    if isinstance(val, (list, dict, set, tuple)):
        return json.dumps(val, default=str)
    return str(val)
```

MinuraPunchihewa (approved):

martyna-mindsdb:
@ZoranPandovski 
Can this be merged now?

ZoranPandovski:
@martyna-mindsdb Done

PR #11178, opened 2025-06-25
Added Structured JSON response output from LLMs

### LINKED ISSUE #11172: [Bug]: [KB Evalute] Unable to parse generated answer from the LLM

KB Evaluate is throwing an error while parsing the response from LLM. JSON fails because it does not sanitize the response
It can have stuff like  ```json makes it fail

### Video or screenshots

Image

Image

### Expected behavior

It should parse it smoothly

EVALUATE KNOWLEDGE_BASE my_kb

### Anything else?

It can be fixed in this file

https://github.com/mindsdb/mindsdb/blob/80082cc3d7856a9a6087725004769ab10e85ecf4/mindsdb/interfaces/knowledge_base/evaluate.py#L233-L243

PriyanshuPz:
## Description
Fixed JSON response parsing problem due to backticks in the response.
Added sanitization function. also added id in dataframe so, when metrics is inserted into data source it will have defined unique ID.

Test it by running 
```sql
EVALUATE KNOWLEDGE_BASE kb
USING
    test_table = ds.evaluation_test_data,
    version = 'llm_relevancy',
    evaluate = true,
    generate_data = {
        'from_sql': 'SELECT content FROM ds.evaluation_test_data',
        'count': 100
    }, 
    llm = {
        'provider': 'groq',
        'api_key':'gsk_...',
        'model_name':'qwen/qwen3-32b'
    },
    save_to = ds.evaluations_metrics

```
CHANGE THE VALUES ACCOUDING MAIN THING IS USING GENERATE DATA

Fixes #11172 

## Type of change

## Verification Process

 - [x]   Test Location: Specify the URL or path for testing.

## Additional Media:
image

## Checklist:


PriyanshuPz:
@martyna-mindsdb Please Review the PR.
Thanks

mindsdb/interfaces/knowledge_base/evaluate.py
```diff
@@ -43,6 +44,19 @@ def calc_entropy(values: List[float]) -> float:
     return -sum([pk * math.log(pk) for pk in values])
 
 
+def sanitize_json_response(response: str) -> str:
+    """Remove markdown code block formatting from JSON response."""
+    # Remove leading/trailing whitespace
+    # Try to find the first JSON object using regex (this assumes a flat or nested dict)
+    json_match = re.search(r"\{.*?\}", response, re.DOTALL)
```
ea-rus: for json like `{"a": {"b": 1}}` this regexp will return `{"a": {"b": 1}`
maybe better to look for first `{` and last `}` symbols?

PriyanshuPz: This is going to be the edge case as we are using json output already still fixed.

ea-rus: if llm add extra key with object, the parsing json will be failed evey if the json is correct.
I think this regexp will handle it better `r"\{.*\}"`

PriyanshuPz: The regex r"\{.*\}" uses greedy matching, which means it will match from the first { to the last } in the string. This could include extra text after the JSON object.

For example, if the LLM response is:
```
Here's your JSON: {"query": "What is X?", "reference_answer": "Y"} and some extra text with } characters.
```

mindsdb/interfaces/knowledge_base/evaluate.py
```diff
@@ -43,6 +44,28 @@ def calc_entropy(values: List[float]) -> float:
...
+
+    # Find the first opening brace
+    # Remove markdown code block markers if present
+    response = re.sub(r"^```(?:json)?\s*", "", response, flags=re.MULTILINE)
+    response = re.sub(r"\s*```$", "", response, flags=re.MULTILINE)
+    response = response.strip()
+
+    # Find JSON object using regex - matches from first { to its corresponding }
+    # This pattern matches balanced braces
+    json_pattern = r"\{(?:[^{}]|{(?:[^{}]|{[^{}]*})*})*\}"
```
ea-rus: counting braces is not reliable, if brace is part of string (more probable than having nested {}) it makes json invalid 
`{"a": 1, "b": "}"}` => `{"a": 1, "b": "}`
`{"a": 1, "b": "{"}` => `{"}`

PriyanshuPz:
Covers these cases.
As we have defined a clear prompt so, it won't have nested JSON objects and using structured output also takes care that only JSON is thrown out.
Still sanitising the output to cover these edge cases.

### 1. **Braces in JSON Strings** (Your main concern)
```python
# These would break brace counting:
'{"query": "What is }?", "reference_answer": "Answer"}'
'{"query": "What is {?", "reference_answer": "Answer"}'
'{"query": "Use {} for formatting", "reference_answer": "Braces"}'
```

### 2. **Multiple JSON Objects**
```python
# LLM might return multiple objects:
'{"query": "Q1", "reference_answer": "A1"} {"query": "Q2", "reference_answer": "A2"}'
```

### 3. **JSON with Extra Text**
```python
# Common LLM responses:
'Here is your JSON: {"query": "What?", "reference_answer": "Answer"} Hope this helps!'
'The answer is {"query": "What?", "reference_answer": "Answer"}.'
```

### 4. **Escaped Characters**
```python
# JSON with escaped quotes and braces:
'{"query": "What is \"JSON\"?", "reference_answer": "JavaScript Object Notation"}'
'{"query": "Use \\{ and \\}", "reference_answer": "Escaped braces"}'
```

### 5. **Nested Objects** (Less common but possible)
```python
'{"query": "What?", "metadata": {"type": "question"}, "reference_answer": "Answer"}'
```

mindsdb/interfaces/knowledge_base/llm_client.py
```diff
@@ -69,6 +80,10 @@ def completion(self, messages: List[dict]) -> str:
         else:
             kwargs = params.copy()
             model = kwargs.pop("model_name")
+            kwargs.pop("provider", None)
+            base_url = params.pop("base_url", None)
```
ea-rus: this block was moved in litellm handler in [this pr](https://github.com/mindsdb/mindsdb/pull/11229/files#diff-fa489cc748534ecd30448850061322ecc3dadd96db76855773f76fa1c65f5436L73)

mindsdb/interfaces/knowledge_base/evaluate.py
```diff
@@ -43,6 +44,49 @@ def calc_entropy(values: List[float]) -> float:
...
+        try:
+            parsed = json.loads(candidate)
+            # Ensure it's a dictionary (object) not just any valid JSON
+            if isinstance(parsed, dict):
+                return candidate
+        except json.JSONDecodeError:
+            continue
+
+    # Fallback: try smaller substrings if the above fails
+    for end_idx in range(start_idx + 2, len(response) + 1):
```
ea-rus: what difference between this block and above? looks like they do the same but in different direction

mindsdb/interfaces/knowledge_base/llm_client.py
PriyanshuPz: This PR removes the logic to set `api_base` by passing `base_url` in parameters. Removing it will remove the option for example using openai as base SDK for the providers like groq that is openai compatible.
(after 2 seconds) Okay, It is migrated to LiteLLM client, Let me remove the duplicate code. 

What I am doing is removing out provider from `kwargs` as it is already being passed with `self.provider` passing it again is causing LiteLLM error.

mindsdb/interfaces/knowledge_base/evaluate.py
PriyanshuPz: in first block it starts from the end of the response and works backwards and tries to find the longest possible valid JSON first, Prioritizes completeness over speed
in second block it starts from just after the opening brace and works forwards and tries to find the shortest possible valid JSON first, Prioritizes speed over completeness

example: ```{"name": "John", "age": 30} some extra text here```

First loop: Will try `{"name": "John", "age": 30}` some extra text here first (fails), then `{"name": "John", "age": 30}` some extra text (fails), eventually finding `{"name": "John", "age": 30}` (succeeds)
Second loop: Will try `{"` first (fails), then `{"n` (fails), eventually finding `{"name": "John", "age": 30}` much sooner

ea-rus: but if first block doesn't find: the secon't won't find too 
and in oposite situation: if first block find something - the second wont' be  used

PriyanshuPz: Hmm. correct, second loop is completely redundant so let me remove that.

ea-rus (approved):

martyna-mindsdb:
@PriyanshuPz 
Please fix the `Test on Push` so we can merge this PR.

martyna-mindsdb:
@hamishfagg 
Test on Push fail due to the litellm dependency. Can you please take a look?

hamishfagg:
@martyna-mindsdb its failing because litellm is being imported in this PR. Litellm seems like a big dependency to add for how it's being used, @PriyanshuPz is there a way to do this without litellm?

PriyanshuPz:
@martyna-mindsdb I have fixed the problem.
@hamishfagg I have added the changes in litellm handler and tested it is working fine

martyna-mindsdb:
@PriyanshuPz 
There is still one error left in Test on Push -- https://github.com/mindsdb/mindsdb/actions/runs/16166173075/job/45628569163?pr=11178

PriyanshuPz:
@martyna-mindsdb I have fixed the tests.

martyna-mindsdb:
Thanks @PriyanshuPz 

@hamishfagg 
Now this requires manual approval from you.

martyna-mindsdb:
Thanks @PriyanshuPz 
Looks like Test on Push is failing now. Can you please take a look?

PriyanshuPz:
@martyna-mindsdb This error is not related to the changes done in this PR. You will need to resolve the SQLite FTS5 and ChromaDB schema issues separately from this PR.

Same error in other PR: https://github.com/mindsdb/mindsdb/actions/runs/16204779108/job/45752389611?pr=11281

PR #11399, opened 2025-08-04
docs -- kb hybrid search

martyna-mindsdb:
## Description

docs -- kb hybrid search

https://github.com/user-attachments/assets/3c3fc8bb-dd01-436a-9f5c-2e6b7570d13f

Fixes #issue_number

## Type of change

### DIFF
--- a/mindsdb/integrations/handlers/huggingface_handler/__init__.py
+++ b/mindsdb/integrations/handlers/huggingface_handler/__init__.py
@@ -11,9 +11,9 @@ from .__about__ import __version__ as version, __description__ as description
 # NOTE: security vulnerability is in `pytorch` v2.7.1, revert changes here and in
 # requirements.txt/requirements_cpu.txt when new version is released
 Handler = None
-import_error = """
-    The `huggingface_handler` is temporary disabled in current version of MindsDB due to security vulnerability.
-"""
+import_error = ImportError(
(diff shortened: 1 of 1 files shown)


fshabashev (approved):
LGTM

martyna-mindsdb:
Thanks @fshabashev 
Question: when using the `hybrid_search_alpha` parameter, is it required to set `reranking=false`?

fshabashev:
> Question: when using the hybrid_search_alpha parameter, is it required to set reranking=false?


when reranking is set to false hybrid_search_alpha is ignored

martyna-mindsdb:
@fshabashev 
When reranking is set to TRUE, then hybrid_search_alpha is ignored. And when reranking is set to false, then hybrid_search_alpha is used.

So this query uses alpha reranking:
```
SELECT * from my_kb
WHERE
	content = ”ACME-213”
	AND hybrid_search_alpha = 0.8 -- optional, this is 0.5 by default
	AND hybrid_search = true
	AND reranking=false -- this needs to be added when KB has its reranking model defined
```

And this query uses KB's reranking:
```
SELECT * from my_kb
WHERE
	content = ”ACME-213”
	AND hybrid_search = true
```

fshabashev:
yes, correct

PR #11428, opened 2025-08-12
Auth a2a

torrmal:
Description

This PR implements a unified authentication token management system for MindsDB that provides secure token-based authentication for external services like A2A (Agent-to-Agent) and MCP servers. The system ties authentication requirements to the HTTP API configuration, ensuring consistent security across all services.

Key Features:
Generic authentication token management with /api/auth_tokens/* endpoints
Automatic token validation for external services via HTTP API calls
Support for both A2A and MCP server authentication
Backward compatibility with existing MCP token configurations
Conditional authentication based on HTTP auth configuration

Problem Solved:
Previously, external services like A2A and MCP servers had inconsistent authentication mechanisms. A2A servers had no authentication, while MCP servers only supported environment-based tokens. This created security gaps and configuration complexity.

Type of change
[x] ⚡ New feature (non-breaking change which adds functionality)
[x] 📄 This change requires a documentation update
Verification Process
Test Location:
HTTP API: http://localhost:47334/api/auth_tokens/*
A2A Server: http://localhost:10002/*
MCP Server: http://localhost:47337/*
Verification Steps:
1. Test Token Management (HTTP Auth Disabled)
status
2. Test A2A Server Authentication
"
3. Test MCP Server Authentication
'
4. Test HTTP Auth Integration
txt
5. Test MCP Environment Token Priority
'
Additional Media:
[ ] I have attached a brief loom video or screenshots showcasing the new functionality or change.
Checklist:
[x] My code follows the style guidelines(PEP 8) of MindsDB.
[x] I have appropriately commented on my code, especially in complex areas.
[x] Necessary documentation updates are either made or tracked in issues.
[x] Relevant unit and integration tests are updated or added.
Files Changed
New Files:
mindsdb/utilities/a2a_auth.py - Shared authentication utilities
mindsdb/api/http/namespaces/configs/auth_tokens.py - Auth tokens namespace config
mindsdb/api/http/namespaces/auth_tokens.py - HTTP API endpoints
mindsdb/api/a2a/common/server/auth_middleware.py - A2A authentication middleware
docs/a2a_authentication.md - Comprehensive documentation
Modified Files:
mindsdb/api/a2a/common/server/server.py - Added auth middleware
mindsdb/api/a2a/__main__.py - Updated A2A server initialization
mindsdb/api/mcp/start.py - Enhanced MCP authentication
mindsdb/api/http/initialize.py - Added auth_tokens namespace
Architecture
The implementation uses a separate server architecture where:
[...]

PR #11441, opened 2025-08-14
Unify http, mcp, a2a servers

hamishfagg:
Creates a single HTTP server that serves MCP, A2A, and regular HTTP API traffic:

- Removes a lot of MCP and A2A config - these now use the same host and server as mindsdb
- Removes `server_type` config - we need to use uvicorn for this and so can't support the other servers.
- Creates similar functions for API app instantiation - `get_x_app()`
- Constructs a unified app in the http API `start()` function
- Removes uvicorn-specific logging config and replaces with `log_config=None` on uvicorn commands, which instruct it to use mindsdb's logging config.
- Generalizes auth for HTTP based APIs with a single PAT based approach.

Solves #11448


torrmal:
@hamish if we could combine efforts with this one:

https://github.com/mindsdb/mindsdb/pull/11428

hamishfagg:
@dusvyat is the a2a server run separately for anything (via `__main__.py`)? If not can we remove the cmdline options stuff ?

docs/setup/environment-vars.mdx
```diff
@@ -90,28 +90,6 @@ To use the A2A API when running MindsDB at a remote host, use the `MINDSDB_A2A_H
     ```
 </CodeGroup>
 
-## MindsDB Server
```
lucas-koontz, line 93: Is there new documentation for MindsDB Server?

mindsdb/api/a2a/common/server/task_manager.py
```diff
@@ -31,10 +31,10 @@
     InternalError,
 )
 from ...common.server.utils import new_not_implemented_error
+from mindsdb.utilities import log
 import asyncio
-import logging
 
-logger = logging.getLogger(__name__)
+logger = log.getLogger(__name__)
```
lucas-koontz, line 37: 👏

mindsdb/api/a2a/common/server/server.py
```diff
@@ -107,13 +81,20 @@ async def _process_request(self, request: Request):
             body = await request.json()
             json_rpc_request = A2ARequest.validate_python(body)
 
+            user_info = {
+                "user-id": request.headers.get("user-id", None),
+                "company-id": request.headers.get("company-id", None),
+                "user-class": request.headers.get("user-class", None),
+                "authorization": request.headers.get("Authorization", None),
+            }
```
lucas-koontz, line 89: Do we have a Pydantic Model for this class? If so, I would prefer using that. If we don't, keep this as is.

Also,

💅 A method to extract the headers would also be beneficial to reuse elsewhere.

hamishfagg: We don't. I added this just to pass headers through from the gateway to get gateway auth working on a2a

mindsdb/api/a2a/agent.py
```diff
@@ -116,14 +116,12 @@ async def stream(
...
         """Stream responses from the MindsDB agent (uses streaming API endpoint)."""
         try:
-            logger.info(f"Using streaming API for query: {query[:100]}...")
             # Create A2A message structure with history and current query
             a2a_message = {"role": "user", "parts": [{"text": query}]}
             if history:
                 a2a_message["history"] = history
             # Convert to Q&A format using centralized utility
             formatted_messages = convert_a2a_message_to_qa_format(a2a_message)
-            logger.debug(f"Formatted messages for agent: {formatted_messages}")
```
lucas-koontz: Was this removal intentional?

hamishfagg: nope, good catch

mindsdb/api/common/middleware.py
```diff
@@ -0,0 +1,106 @@
...
+import os
+import traceback
+
+from mindsdb.utilities import log
+from mindsdb.utilities.config import config
+
+logger = log.getLogger(__name__)
+
+SECRET_KEY = os.environ.get("AUTH_SECRET_KEY") or secrets.token_urlsafe(32)
+# We store token (fingerprints) in memory, which means everyone is logged out if the process restarts
```
lucas-koontz, line 18: Do you know if this is intentional behavior? Why don't we preserve the sessions?

hamishfagg: Just trying to avoid complexity if we don't need it. Is there a use case for keeping sessions between restarts?

docs/setup/environment-vars.mdx
hamishfagg: There aren't any env-vars now that control server setup, so I removed this section

lucas-koontz (approved):

mindsdb/api/common/middleware.py
lucas-koontz: It was more of a genuine question to understand the system. I'm not sure what the intended behavior.

PR #11492, opened 2025-08-29
Added API Metadata to Config Endpoint

MinuraPunchihewa:
## Description

This PR updates the GET /config endpoint to provide information regarding the various APIs within MindsDB including the host, port and whether or not the service is actually running.

Fixes https://linear.app/mindsdb/issue/CONN-1320/provide-api-metadata-for-front-end-dashboard

## Type of change

## Verification Process

## Checklist:


mindsdb/utilities/api_status.py
```diff
@@ -0,0 +1,49 @@
...
+
+_api_status_file = None
+
+
+def get_api_status_file():
+    global _api_status_file
+    if _api_status_file is None:
+        # Use a temporary file that can be shared across processes.
+        temp_dir = tempfile.gettempdir()
+        _api_status_file = os.path.join(temp_dir, "mindsdb_api_status.json")
```
ea-rus, line 17: mindsdb server creates `mindsdb` folder in temp directory. this file can be also located there

MinuraPunchihewa: @ea-rus You are right. I just updated the logic to create the file in the to store this file in `tmp` directory created by the server.
I also added an exit handler to delete the contents of this folder when the application is stopped. It looks like we are currently maintaining the contents across startups.

PR #11495, opened 2025-08-30
added datasource integration testing

sejubar:
## Description
This pull request integrates the new Data Source Integration (DSI) testing framework into the main MindsDB repository. The framework is designed to provide automated, configuration-driven, end-to-end tests for all data handlers.

The core of this framework is a test generation script that:

Connects to a MindsDB instance to get a list of installed handlers.

For each handler specified in the .env configuration, it creates a temporary database connection.

It then runs SHOW TABLES to perform live discovery of the handler's tables or endpoints.

Finally, it generates a static pytest test file with a suite of health-check and custom tests for each discovered table.

This entire process is now seamlessly integrated into the pytest lifecycle.

Fixes #issue_number

## Type of change

## Verification Process

provide the correct values for variables in the .env file and set variable for list of handlers we want to test e.g 
HANDLERS_TO_TEST=postgres,s3,github,databricks,bigquery,mariadb,mysql,mssql
for the above configuration we should see two failure provided handlers github,databricks are not installed

## Additional Media:

## Checklist:


sejubar:

sejubar:

hamishfagg:
I don't think writing python code into a file with python is the right approach.
We can take the generated test parameters and pass them into a parametized test, we don't need to write each test out individually: https://docs.pytest.org/en/stable/example/parametrize.html

sejubar:
> I don't think writing python code into a file with python is the right approach. We can take the generated test parameters and pass them into a parametized test, we don't need to write each test out individually: https://docs.pytest.org/en/stable/example/parametrize.html

Right, 
this has been addressed, removed file generation and parametrized the tests. @hamishfagg

sejubar:

.env.example
```diff
@@ -0,0 +1,100 @@
...
+# Comma-separated list of handler names.
+# -------------------------------------
+HANDLERS_TO_TEST=postgres,s3,github,databricks,bigquery,mariadb,mysql
+
+# =====================================
+# Data Source Credentials
+# =====================================
+
+# --- PostgreSQL ---
+PG_SOURCE_HOST=samples.mindsdb.com
```
hamishfagg, line 29: We should make as many of these DB params as possible just work out of the box. As it stands users have to manually copy this file

Makefile
```diff
@@ -56,4 +65,4 @@ unit_tests_debug:
 	env PYTHONPATH=./ pytest $(PYTEST_ARGS_DEBUG) tests/unit/executor/  
 	pytest $(PYTEST_ARGS_DEBUG) --ignore=tests/unit/executor tests/unit/
 
-.PHONY: install_mindsdb install_handler precommit format run_mindsdb check build_docker run_docker integration_tests integration_tests_slow integration_tests_debug unit_tests unit_tests_slow unit_tests_debug
+.PHONY: install_mindsdb install_handler precommit format run_mindsby check build_docker run_docker integration_tests integration_tests_slow integration_tests_debug datasource_integration_tests datasource_integration_tests_debug unit_tests unit_tests_slow unit_tests_debug
```
hamishfagg: typo

.gitignore
```diff
@@ -64,3 +64,6 @@ node_modules
 
 mindsdb/**/pyproject.toml
 mindsdb/**/uv.lock
+
+reports/
+test_generated_integrations.py
```
hamishfagg: this can be removed as we don't generate it anymore right?

tests/README.md
```diff
@@ -32,6 +32,16 @@ To execute unit or integration tests, use the following pytest command:
 pytest -vx <test_dir>
 ```
 
+To run ONLY the DSI tests:
+```
+pytest -v --run-dsi-tests tests/integration/handlers/
```
hamishfagg, line 37: Should list the makefile command for this

tests/conftest.py
```diff
@@ -1,19 +1,144 @@
...
+import sys
+import json
+import logging
+from pathlib import Path
+from dotenv import load_dotenv
+from typing import Generator, Any
 
+# DSI: Add project root to path for script imports
+project_root = Path(__file__).parent.parent
+sys.path.insert(0, str(project_root))
```
hamishfagg: I'm not sure this is a good idea. Messing around with the path has caused issues in the past. And the supported method of running this should me using the makefile, in which case the user's current dir is known.

tests/conftest.py
```diff
@@ -1,19 +1,144 @@
...
     config.addinivalue_line("markers", "slow: mark test as slow to run")
+    config.addinivalue_line("markers", "dsi: mark test as part of the DSI framework")
+
+    if config.getoption("--run-dsi-tests"):
+        logging.basicConfig(
+            level=logging.INFO,
+            format='%(asctime)s - %(levelname)s - %(message)s',
+            datefmt='%Y-%m-%d %H:%M:%S',
+            force=True
+        )
```
hamishfagg: Why are we configuring logging here?

tests/conftest.py
```diff
@@ -1,19 +1,144 @@
...
+        logging.info("--- DSI: Loading environment variables ---")
+        try:
+            dotenv_path = project_root / '.env'
+            if load_dotenv(dotenv_path=dotenv_path, override=True):
+                logging.info(f"DSI: Successfully loaded environment variables from: {dotenv_path}")
+            else:
+                logging.warning(f"DSI: Could not find .env file at {dotenv_path}. Using system variables.")
+        except Exception as e:
```
hamishfagg: Don't do this. We need to either catch exceptions we want to do something about, or let them raise

tests/conftest.py
```diff
@@ -1,19 +1,144 @@
...
+            pytest.fail(f"DSI: A critical error occurred while loading the .env file: {e}", pytrace=False)
+        
+        logging.info("--- DSI: Test generation will be handled by pytest parametrization ---")
 
 
 def pytest_collection_modifyitems(config, items):
-    if config.getoption("--runslow"):
-        # --runslow given in cli: do not skip slow tests
+    if not config.getoption("--runslow"):
```
hamishfagg, line 16: The logic has been inverted here? Won't think make slow tests run by default?

tests/conftest.py
```diff
@@ -1,19 +1,144 @@
...
+            if "slow" in item.keywords:
+                item.add_marker(skip_slow)
+
+    if not config.getoption("--run-dsi-tests"):
+        items[:] = [item for item in items if "dsi" not in item.keywords]
+
+
+def pytest_sessionfinish(session: Any, exitstatus: int) -> None:
+    if session.config.getoption("--run-dsi-tests"):
+        logging.info("--- DSI: Test session finished. No generated files to clean up. ---")
```
hamishfagg: Can probably just delete this?

tests/conftest.py
```diff
@@ -1,19 +1,144 @@
...
-    for item in items:
-        if "slow" in item.keywords:
-            item.add_marker(skip_slow)
+    
+    from tests.integration.handlers.utils.helpers import connect_to_mindsdb
+        
+    try:
+        server = connect_to_mindsdb()
+        yield server
+    except Exception as e:
```
hamishfagg: Don't catch all

tests/conftest.py
```diff
@@ -1,19 +1,144 @@
...
+        logging.info("--- DSI: Test session finished. No generated files to clean up. ---")
+
+
+# --- DSI Framework Fixtures with Improved Error Handling ---
+
+@pytest.fixture(scope="session")
+def mindsdb_server(pytestconfig) -> Generator[Any, None, None]:
+    if not pytestconfig.getoption("--run-dsi-tests"):
+        yield None
         return
```
hamishfagg: This makes the behaviour of this confusing. It should just always return a connection when called
You can probably also just move `connect_to_mindsdb()` to here

tests/conftest.py
```diff
@@ -1,19 +1,144 @@
...
+        yield server
+    except Exception as e:
+        pytest.fail(
+            f"DSI: Failed to connect to MindsDB via SDK. Please ensure MindsDB is running and accessible. Error: {e}", 
+            pytrace=False
+        )
+
+
+@pytest.fixture(scope="session")
+def query_logger(pytestconfig):
```
hamishfagg: it might be better to subclass or patch the SDK client so that it logs queries automatically. That way we don't have to remember to log each one separately.

tests/conftest.py
```diff
@@ -1,19 +1,144 @@
...
+    if not pytestconfig.getoption("--run-dsi-tests"):
+        yield None
+        return
```
hamishfagg: I'd remove this, it makes behaviour confusing

tests/conftest.py
```diff
@@ -1,19 +1,144 @@
...
+        logging.info(f"DSI: Full query log saved to {log_filepath}")
+    except Exception as e:
+        logging.error(f"DSI: Failed to save the query log file: {e}")
+
+
+@pytest.fixture(scope="session")
+def session_databases(mindsdb_server, pytestconfig):
+    if not pytestconfig.getoption("--run-dsi-tests") or mindsdb_server is None:
+        yield None
+        return
```
hamishfagg: same as above

tests/conftest.py
```diff
@@ -1,19 +1,144 @@
...
+            logging.warning(f"DSI: Skipping database creation for {handler_name}: {skip_reason}")
+            continue
+            
+        try:
+            mindsdb_server.query(f"DROP DATABASE IF EXISTS {db_name};").fetch()
+            create_query = f"CREATE DATABASE {db_name} WITH ENGINE = '{handler_name}', PARAMETERS = {params_clause};"
+            mindsdb_server.query(create_query).fetch()
+            created_dbs[handler_name] = db_name
+            logging.info(f"DSI: Successfully created database '{db_name}' for handler '{handler_name}'.")
+        except Exception as e:
```
hamishfagg: don't catch all

tests/integration/handlers/README.md
hamishfagg, line 1: There's a lot of overlap between this and tests/integration/handlers/DevGuide.md
Both are also out of date in some places

tests/integration/handlers/test_data_sources.py
```diff
@@ -0,0 +1,172 @@
...
+def generate_test_cases_for_parametrization() -> List[Dict[str, Any]]:
+    """
+    Generates test case definitions for all handlers to be used by pytest.parametrize.
+    Connects to MindsDB temporarily just for test case discovery.
+    """
+    try:
+        # We need a temporary server connection to discover handlers and tables.
+        # The actual tests will use the 'mindsdb_server' fixture from conftest.
+        server = connect_to_mindsdb()
+    except Exception as e:
```
hamishfagg: dont catch all

tests/integration/handlers/test_data_sources.py
```diff
@@ -0,0 +1,172 @@
...
+            test_cases.append({
+                'handler_name': handler_name,
+                'test_type': 'skipped_test',
+                'reason': error_message
+            })
+            continue
+        finally:
+            server.query(f"DROP DATABASE IF EXISTS {temp_db_name};").fetch()
+
+        config_path = Path(__file__).parent / 'configs' / f'{handler_name}.json'
```
hamishfagg: This is important, put it at the top

tests/integration/handlers/test_data_sources.py
```diff
@@ -0,0 +1,172 @@
...
+    try:
+        if test_type == 'negative':
+            details = test_case['details']
+            try:
+                mindsdb_server.query(query).fetch()
+                pytest.fail('Negative test failed. Query was expected to raise an error, but it succeeded.')
+            except Exception as e:
+                error_str = str(e)
+                assert details['expected_error'] in error_str, f'Negative test failed. Expected error substring "{details["expected_error"]}" not found in actual error: {error_str}'
+                error = error_str
```
hamishfagg: Use:
```
with pytest.raises(<ExceptionType>):
    mindsdb_server.query(query).fetch()
```

tests/integration/handlers/utils/config.py
```diff
@@ -0,0 +1,105 @@
+import os
+import logging
+from dotenv import load_dotenv
+from typing import Dict, Optional, Any
+from pathlib import Path
+
+# --- Corrected: Load .env from the project root ---
+# Navigates up from tests/integration/handlers/utils to the main project root
+project_root = Path(__file__).parent.parent.parent.parent.parent
```
hamishfagg: same as above

tests/integration/handlers/utils/helpers.py
```diff
@@ -0,0 +1,108 @@
...
+        if config.MINDSDB_USER and config.MINDSDB_PASSWORD:
+            url = f"{config.MINDSDB_PROTOCOL}://{config.MINDSDB_HOST}:{config.MINDSDB_PORT}"
+            logging.info(f"DSI: Connecting SDK with credentials to server at: {url}")
+            server = mindsdb_sdk.connect(
+                url=url,
+                login=config.MINDSDB_USER,
+                password=config.MINDSDB_PASSWORD
+            )
+        else:
+            url = f"{config.MINDSDB_PROTOCOL}://{config.MINDSDB_HOST}:{config.MINDSDB_PORT}"
```
hamishfagg: Duplicate of above

hamishfagg (changes requested):

tests/README.md
sejubar: yes

tests/conftest.py
sejubar: updated but no, slow tests are skipped else `--runslow` flag is provided.

```
# "If the user did NOT provide the --runslow flag..."
    if not config.getoption("--runslow"):
        # "...then find the slow tests and mark them to be skipped."
        skip_slow = pytest.mark.skip(reason="need --runslow option to run")
        for item in items:
            if "slow" in item.keywords:
                item.add_marker(skip_slow)
```

.env.example
sejubar: Out of the box will mean ? 
I believe users will need to test 1 or two handlers otherwise we will have set of priotized handlers to execute in ci pipeline. 

could you provide example of out of the box you mean ?  are we already doing this somewhere ?

hamishfagg: Out of the box = without the user having to set anything up.

So for DB types where we have a sample DB, we can default to the sample DB's user/pass/host etc. If there is a sane default value we shouldn't force the user to copy an env file etc.

sejubar: Absolutely right, can you provide those defaults for the missing ones in the list s3,github? I actually like the idea of default target datasource
When this task was give the request was that this can be configured by users, providing credentials for a target datasource should not be seen as forcing the user. What if the user needs to test against different target datasource? 

**What I will do**:
keep the configuration options and have the test use default credentials if not config is provided.

sejubar: the test now uses default credentials if no config is provided. I am resolving this comment

tests/integration/handlers/README.md
```diff
@@ -0,0 +1,125 @@
...
+
+1.  **Create and activate a virtual environment:**
+    ```bash
+    uv venv
+    source .venv/bin/activate  # On Windows, use `.venv\Scripts\activate`
+    ```
+
+2.  **Install dependencies:**
+    ```bash
+    uv pip install -r requirements.txt
```
hamishfagg: should this be `requirements/requirements-test.txt` ?

tests/integration/handlers/README.md
```diff
@@ -0,0 +1,125 @@
...
+    * Copy the `.env.example` to a new file named `.env`.
+    * Fill in the .env file with your MindsDB server details and the credentials for any data sources you wish to test.
+    * **(Optional)**: Fill in the PG_LOG_* variables if you wish to use the script to ingest results into a PostgreSQL database.
+
+## Running Tests
+
+To run the full test suite, execute the following command from the project's root directory:
+
+```bash
+pytest -v --run-dsi-tests tests/integration/handlers/
```
hamishfagg, line 39: Can we use the makefile targets here please?
`make datasource_integration_tests`

tests/integration/handlers/conftest.py
```diff
@@ -0,0 +1,138 @@
...
+    def patched_query_constructor(sql_query: str):
+        query_object = original_query_method(sql_query)
+        query_object.__class__ = PatchedQuery
+        return query_object
+
+    # Replace the original server.query with our new constructor
+    server.query = patched_query_constructor
+    # --- END OF PATCHING LOGIC ---
+
+    yield server
```
hamishfagg: Eh? You can do all of this with:

```
from functools import wraps

def mindsdb_server(...
    original_query_method = server.query
    @wraps(original_query_method)
    def logged_query(*args, **kwargs):
        # Include your timing etc
        result = original_query_method(*args, **kwargs)
        query_log_data[handler_name].append(...)
        
    server.query = logged_query
    yield server
```

tests/integration/handlers/utils/config.py
```diff
@@ -0,0 +1,148 @@
...
+MINDSDB_USER: Optional[str] = os.getenv("MINDSDB_USER")
+MINDSDB_PASSWORD: Optional[str] = os.getenv("MINDSDB_PASSWORD")
+
+# --- Test Execution Configuration ---
+# Default to handlers that have public credentials for a better out-of-the-box experience.
+HANDLERS_TO_TEST: str = os.getenv("HANDLERS_TO_TEST", "postgres,mariadb,mysql,mssql")
+
+# --- Data Source Credentials (Convention: HANDLERNAME_CREDS) ---
+POSTGRES_CREDS: Dict[str, Any] = {
+    "host": get_creds('postgres', "PG_SOURCE_HOST", "host"),
```
hamishfagg: Instead of DEFAULT_CREDS being a dictionary and passing two separate lookups for a variable here (`postgres + host` and `PG_SOURCE_HOST`), then using get_creds, can't we just list the default here? with:

```
os.getenv("PG_SOURCE_HOST", "samples.mindsdb.com")
```

hamishfagg (changes requested):

hamishfagg:
You still are catching some generic `Exception`s  (or casting more specific exceptions to `Exception`).
The general wisdom is to only catch an exception if you have something useful to do with it, or to handle the exception (i.e. it shouldn't be shown to the user).

Logging the exception and then raising it isn't particularly useful either, since it's logged by python anyway. At the very least we need to include traceback info with the log message like:

```
except <some_specific_exception>:
    logger.exception()
    raise
```

tests/integration/handlers/conftest.py
sejubar: resolved

tests/integration/handlers/utils/config.py
sejubar: :-) resolved

sejubar:
> You still are catching some generic `Exception`s (or casting more specific exceptions to `Exception`). The general wisdom is to only catch an exception if you have something useful to do with it, or to handle the exception (i.e. it shouldn't be shown to the user).
> 
> Logging the exception and then raising it isn't particularly useful either, since it's logged by python anyway. At the very least we need to include traceback info with the log message like:
> 
> ```
> except <some_specific_exception>:
>     logger.exception()
>     raise
> ```
All your comments should be addressed now. Let me know what you think!
in files (conftest.py, helpers.py, and test_data_sources.py) based on your review,  updated, all the try...except blocks to catch the specific RuntimeError from the SDK. I also adopted your suggestion to use logging.exception() in the setup fixtures to get the full traceback in the logs.

tests/integration/handlers/README.md
sejubar: added

PR #11519, opened 2025-09-04
docs -- mindsdb objects naming

martyna-mindsdb:
## Description

docs -- mindsdb objects naming

Fixes #issue_number

## Type of change

### DIFF
--- a/docs/mindsdb-objects.mdx
+++ b/docs/mindsdb-objects.mdx
@@ -1,0 +1,148 @@
+---
+title: Naming Standards for MindsDB Objects
+sidebarTitle: MindsDB Objects
+icon: "puzzle-piece-simple"
+---
+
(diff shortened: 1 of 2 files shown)


docs/mindsdb-objects.mdx
```diff
@@ -0,0 +1,148 @@
+---
+title: Naming Standards for MindsDB Objects
+sidebarTitle: MindsDB Objects
+icon: "puzzle-piece-simple"
+---
+
+MindsDB allows you to create and manage a variety of entities within its ecosystem. All MindsDB objects follow the same naming conventions to ensure consistency and compatibility across the platform.
+
+## MindsDB Entities
```
StpMax, line 9: should we also mention  `create skill` and `create model`?

martyna-mindsdb: These are not exposed to users at this point.

We have a new syntax for agents that does not use skills. Also, we have hidden the CREATE MODEL feature from docs because we focus on default models setup instead.

PR #11554, opened 2025-09-11
Update mssql_handler.py

gauiPPP:
Adding brackets [] around schema name, table name and column name so that names with spaces would work in queries "SELECT * FROM data base.dbo 1.table name" leads to an error while "SELECT * [data base].[dbo 1].[table name]" will work so after these changes.

## Description

Deleting simple SQL query rows and replacing with CONCAT. Tested to be working.

Fixes #issue_number

1 or 5 lines

- [x ] 🐛 Bug fix (non-breaking change which fixes an issue)

## Verification Process

 - [x ]   Test Location: Specify the URL or path for testing.
 On premises MS SQL server
 - [ x]   Verification Steps: Outline the steps or queries needed to validate the change. Include any data, configurations, or actions required to reproduce or see the new functionality.
 Running queries with the changes.

## Checklist:

- [? ] My code follows the style guidelines(PEP 8) of MindsDB.
- [x ] I have appropriately commented on my code, especially in complex areas.
- [? ] Necessary documentation updates are either made or tracked in issues.
- [? ] Relevant unit and integration tests are updated or added.

### DIFF
--- a/mindsdb/integrations/handlers/mssql_handler/mssql_handler.py
+++ b/mindsdb/integrations/handlers/mssql_handler/mssql_handler.py
@@ -283,8 +283,8 @@ class SqlServerHandler(DatabaseHandler):
 
         query = f"""
             SELECT
-                table_schema,
-                table_name,
+                CONCAT('[',table_schema,']') AS schema_name,
+                CONCAT('[',table_name,']') AS table_name,
                 table_type
             FROM {self.database}.INFORMATION_SCHEMA.TABLES
             WHERE TABLE_TYPE in ('BASE TABLE', 'VIEW');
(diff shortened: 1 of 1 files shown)


gauiPPP:

gauiPPP:
The check seems to to failing with ruff format check, code 1; so according to ruff that would be pass but with exceptions.
Unit tests are not being run.

Ruff probably explains why I encounter this issue with most product trying to connect to MS SQL Server.

If Ruff is concerned with SQL queries specifically, then the combining [ and ] with the results should then happen after the query results have been found.

mindsdb/integrations/handlers/mssql_handler/mssql_handler.py
```diff
@@ -283,8 +283,8 @@ def get_tables(self) -> Response:
 
         query = f"""
             SELECT
-                table_schema,
-                table_name,
+                CONCAT('[',table_schema,']') AS schema_name,
```
gauiPPP, line 286: Should be 
 CONCAT('[',table_schema,']') AS table_schema,

StpMax:
Wrapping identifier names with square brackets via CONCAT in the SELECT list embeds the brackets into the data values returned by the query, not into the SQL identifiers themselves — so this does nothing to fix quoting issues in subsequent queries.

PR #11581, opened 2025-09-17
Use bigger limit for reranker 

ea-rus:
## Description

If reranker is used - extend limit of the query:
 - get more records before reranking usage:  twice size of input but not greater than 30

Examples:
- limit = 10, sample size = 20
- limit = 20, sample size = 40
- limit = 50, sample size = 80

The problem to solve: different limit can lead to different results. Reranker gets the input from semantic search that might not include more relevant item if it is out of scope of limit. 
For example if the most suitable record is A and it is at the 5th place of semanic search:
- if query without limit is used: default limit is 10 and reranker gets 10 record and puts A in the first place
- if query with limit 4 is used: reranger gets only 4 records (and no A record there). It means that wrong record will be at the first place and record A will absent in result

Side-update: removed logging with secrets (reranking_model_params)

Fixes https://linear.app/mindsdb/issue/FQE-652/bug-limit-doesnt-consider-relevance-when-querying-knowledge-base

## Type of change

## Verification Process

## Additional Media:

## Checklist:

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -40,7 +40,7 @@ jobs:
     # We only want to run one deploy job for an env at a time
     # Don't cancel in progress jobs because it may be for a different PR
     env:
-      MINDSDB_DB_CON: ${{ secrets.MINDS_DB_URI }}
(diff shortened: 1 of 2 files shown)


alecantu7 (approved):
Really smart solution!

PR #11616, opened 2025-09-25
Support `ollama` in knowledge bases

StpMax:
## Description

Added support of `ollama` provider in knowledge bases.

Example:
1. run ollama
2. pull models:
```bash
ollama pull nomic-embed-text
ollama pull gemma3:270m
```
3. create and use knowledge base:
```sql
CREATE KNOWLEDGE_BASE my_kb
USING
    embedding_model = {
        "provider": "ollama",
        "model_name" : "nomic-embed-text",
        "base_url": "http://127.0.0.1:11434/v1",
        "api_key": "n/a"
    },
    reranking_model = {
        "provider": "ollama",
        "model_name" : "gemma3:270m",
        "base_url": "http://127.0.0.1:11434/v1",
        "api_key": "n/a"
    },
    content_columns = ['product'],
    id_column = 'order_id';

INSERT INTO my_kb SELECT order_id, product, notes FROM example_db.orders;

SELECT chunk_content, distance, relevance FROM my_kb WHERE content = 'mouse';
```

Fixes #FQE-594

## Type of change

## Verification Process

## Additional Media:

## Checklist:

### DIFF
--- a/.github/workflows/build_deploy_dev.yml
+++ b/.github/workflows/build_deploy_dev.yml
@@ -13,7 +13,7 @@ on:
 # Cancel any existing runs of this workflow on the same branch/pr
 # We always want to build/deploy/test a new commit over an older one
 concurrency:
-  group: ${{ github.workflow_ref }}
+  group: ${{ github.workflow }}-${{ github.head_ref }}
   cancel-in-progress: true
 
 jobs:
--- a/.github/workflows/build_deploy_staging.yml
+++ b/.github/workflows/build_deploy_staging.yml
@@ -11,6 +11,7 @@ on:
       - closed
     branches:
       - 'main'
+      - 'develop'
       - 'release/*'
 
 concurrency:
@@ -116,7 +117,7 @@ jobs:
                       "type": "header",
                       "text": {
                         "type": "plain_text",
-                        "text": "TEST RUN FAILED ON ${{ github.head_ref }}",
+                        "text": "TEST RUN FAILED ON ${{ github.base_ref }}",
                         "emoji": true
                       }
                     },
--- a/mindsdb/api/http/initialize.py
+++ b/mindsdb/api/http/initialize.py
@@ -95,7 +95,7 @@ def custom_output_json(data, code, headers=None):
 
 
 def get_last_compatible_gui_version() -> Version | bool:
-    logger.debug("Getting last compatible frontend..")
+    logger.debug("Getting last compatible frontend...")
     try:
         res = requests.get(
             "https://mindsdb-web-builds.s3.amazonaws.com/compatible-config.json",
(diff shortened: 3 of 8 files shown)


mindsdb/interfaces/knowledge_base/controller.py
```diff
@@ -1260,13 +1260,13 @@ def _create_embedding_model(self, project_name, engine="openai", params: dict =
             raise ValueError("'provider' parameter is required for embedding model")
 
         # check available providers
-        avail_providers = ("openai", "azure_openai", "bedrock", "gemini", "google")
+        avail_providers = ("openai", "azure_openai", "bedrock", "gemini", "google", "ollama")
```
tino097, line 1263: there  is still a `custom_openai` that can be set from the UI which will make this check to fail

StpMax: @tino097 At the moment, if select 'custom openai' on the form, then in this method params['provider'] is 'openai'. I think it would be better to add the 'ollama' option to the GUI, or add optional parameter for 'no-logprobs' if 'openai' is used.

mindsdb/integrations/utilities/rag/rerankers/base_reranker.py
```diff
@@ -74,7 +84,10 @@ def _init_client(self):
                     timeout=self.request_timeout,
                     max_retries=2,
                 )
-            elif self.provider == "openai":
+            elif self.provider in ("openai", "ollama"):
+                if self.provider == "ollama":
+                    self.method = "no-logprobs"
+                    self.logprobs = False
```
ea-rus, line 90: self.logprobs is not used when `method = "no-logprobs"`

mindsdb/interfaces/knowledge_base/controller.py
tino097: image
If `custom_openai` is selected in the UI, this will raise an error at line 1264 because `params["provider"]` gets set to `custom_openai`. That breaks the current validation and also causes downstream issues in the flow.

I see two possible approaches:

1. Add `custom_openai` to the list of available providers so it’s treated as a option in the backend.
2. Or, adjust the UI form so that users cannot select `custom_openai` unless explicitly supported. In that case, it should be made clear in the UI that the only “custom” provider supported at the moment is `ollama`.

Either way, keeping UI and backend consistent will prevent these errors and avoid user confusion.

@StpMax @ea-rus

StpMax: @tino097 Hm, as i see, if select 'Custom OpenAI' in current version of gui, then it send PUT request with request body containing custom base_url and 'openai' as provider. Therefore it will pass check in line 1264, but it will fail later, because ollama does not have logprobs.
Also, it is not possible to select nomic-embed-text as embedding model, because it does not support 'completion' method.
I'll create a ticket for frontend team to:
 - add 'ollama' to the 'provider' dropdown list
 - use another method to validate embedding model

PR #11626, opened 2025-09-26
Fix boot error messages

torrmal:
## Description

When I boot Mindsdb, I was getting some warnings and errors, this should be fixed, some of them are external dependencies pedantic warnings, we can't control those, but else should have a clean start

### DIFF
--- a/mindsdb/integrations/handlers/web_handler/requirements.txt
+++ b/mindsdb/integrations/handlers/web_handler/requirements.txt
@@ -1,2 +1,2 @@
 html2text
-bs4
+beautifulsoup4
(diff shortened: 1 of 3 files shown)


ZoranPandovski (dismissed):

ZoranPandovski:
@torrmal Should we make any changes to https://github.com/mindsdb/mindsdb/blob/346c11ebfeec24bf91ca9a33ecd97817c6219587/mindsdb/integrations/handlers/email_handler/email_ingestor.py#L3C1-L3C30. This import will fail?

torrmal:
maybe, not sure if it does as the library is an alias

martyna-mindsdb:
@StpMax @ea-rus 
Can you please review?

PR #11636, opened 2025-09-29
removed documentdb handler

martyna-mindsdb:
## Description

removed documentdb handler -- because it uses mongo api syntax, which has been deprecated

Fixes #issue_number

## Type of change


ZoranPandovski (changes requested):
@martyna-mindsdb we shouldn't remove the documentdb since it still works. We depricated our Mongo API not the MongoDB Handler

martyna-mindsdb:
@ZoranPandovski 
Ok. Then we need to rewrite its [README](https://github.com/mindsdb/mindsdb/tree/main/mindsdb/integrations/handlers/documentdb_handler#readme) into SQL syntax, because the MongoDB syntax is not available now.

PR #11666, opened 2025-10-02
Fix Zendesk Integration

### LINKED ISSUE #11665: [Bug]: Zendesk Integration is not working with WHERE and LIMIT clause
## WHERE condition bug
  
```
SELECT *
FROM zendesk_datasource.users
where email="example@zendesk.com";
```

When executing queries with a where condition on zendesk tables, there is an exception - `too many values to unpack (expected 3)`

## ITERATION bug with LIMIT
When the number of records is >= LIMIT value, there is a `StopIteration` exception

## WHERE condition bug
Image

## ITERATION bug with LIMIT
Image

### Expected behavior

Expected behavior would be for no exceptions while using WHERE and LIMIT 

 ```
CREATE DATABASE zendesk_datasource
WITH
    ENGINE = 'zendesk',
    PARAMETERS = {
      "api_key":{API_KEY},
      "sub_domain": {DOMAIN},
      "email":{EMAIL}
    };
```

Then execute

```
SELECT *
FROM zendesk_datasource.users
where email="example@zendesk.com";
```
Above query will fail for the where condition

Also execute
```
SELECT *
FROM zendesk_datasource.triggers
LIMIT 10;
```
Above query will  fail if you have >= 10 triggers in zendesk

### Anything else?

_No response_

### PR DESCRIPTION
**Pull request by vigbav36**
## Description

Fixes #11665

Zendesk integration was failing due to below reasons

### WHERE Condition Parsing Issue
`mindsdb/integrations/utilities/sql_utils.py:197` was expecting `op, arg1, arg2` on unpacking `conditions` which was passed through `subset_where_conditions` in `mindsdb/integrations/handlers/zendesk_handler/zendesk_tables.py` 

But  subset_where_conditions was passed as a dictionary in the current implemention, so to fix this issue following changes were done in this PR

- Introduce a new variable `api_filters`, which will be a dictionary (as` subset_where_conditions` was previously) to be passed to the Zendesk API calls.
- Convert `subset_where_conditions` to the expected format.

### Iteration over results Issue
When the number of records where more 

```
if isinstance(result, zenpy.lib.generator.BaseResultGenerator):
            while count <= result_limit:
                response.append(result.next().to_dict())
                count += 1
```

Here in the parsing logic, if number of results is less than the result_limit, there will be a StopIteration exception. To solve this I have refactored into using `islice` from `itertools` which does safe slicing

## Type of change

## Verification Process

 - [x]   Verification Steps: 
        - Add a zendesk datasource
        - Query for users with filter on any column
        - Query for activities with a limit which is greater than or equal to total number of records
 

## Additional Media:

- Filtering users by email

image

- Using limit which is greater than number of records

image

## Checklist:


vigbav36:

vigbav36:
recheck

PR #11734, opened 2025-10-13
Fix Groq integration

vigbav36:
## Description

Fixes issue #11582

In groq handler we are re-using open AI handler based on groq reverse compatibility - https://console.groq.com/docs/openai

The root cause was the check in `mindsdb/integrations/handlers/openai_handler/openai_handler.py:496` 

https://github.com/mindsdb/mindsdb/blob/f900128d7f93296741512c6e0f3175232a994da5/mindsdb/integrations/handlers/openai_handler/openai_handler.py#L519

where we are checking if it is a chat model or non chat model. By default all groq models are chat models but in open ai handler only following below models were considered chat models

`"gpt-3.5", "gpt-3.5", "gpt-3.5", "gpt-4", "o3-mini", "o1-mini"` 

This in turn wrongly called non chat model endpoint for groq.

The fix makes an override to the method `is_chat_model(model_name)` which is overriden in groq handler and will always returns true since all Groq models are chat models, referred from here - https://console.groq.com/docs/models

```
@staticmethod
def is_chat_model(model_name):
    """
    All Groq models use the chat completions endpoint, hence every model is a chat model
    """
    return True
```

## Type of change

## Verification Process

 Execute below queries
 
```
CREATE ML_ENGINE groq_engine_1
FROM groq
USING  
    groq_api_key='gsk_smthg';

CREATE MODEL groq_model
PREDICT sentiment
USING
   engine = 'groq_engine_1',
   model_name = 'llama-3.3-70b-versatile',
   prompt_template = 'Classify the sentiment of the following text as one of `positive`, `neutral` or `negative`: {{text}}. Give sentiment as result only.';

SELECT text, sentiment
FROM groq_model
WHERE text = 'I love machine learning!'
```

## Additional Media:

### Before fix
image (4)

### After fix
image (5)

## Checklist:


mindsdb/integrations/handlers/groq_handler/groq_handler.py
```diff
@@ -62,10 +62,10 @@ def create_engine(self, connection_args):
             None
         """
         connection_args = {k.lower(): v for k, v in connection_args.items()}
-        api_key = connection_args.get('groq_api_key')
+        api_key = connection_args.get("groq_api_key")
         if api_key is not None:
-            org = connection_args.get('api_organization')
-            api_base = connection_args.get('api_base') or os.environ.get('groq_BASE', groq_handler_config.BASE_URL)
+            org = connection_args.get("api_organization")
+            api_base = connection_args.get("api_base") or os.environ.get("groq_BASE", groq_handler_config.BASE_URL)
```
tino097: this should be `GROQ_BASE` ?

vigbav36: Hi @tino097 yes, looks like an existing inconsistency !
I have fixed it

mindsdb/integrations/handlers/groq_handler/groq_handler.py
```diff
@@ -44,9 +44,9 @@ def _check_client_connection(client: OpenAI):
         except NotFoundError:
             pass
         except AuthenticationError as e:
-            if e.body['code'] == 401:
-                raise Exception('Invalid api key')
-            raise Exception(f'Something went wrong: {e}')
+            if e.body["code"] == 401:
```
tino097: i know this was from before, but isn't safer to check for `if e.get('body')`first ?

vigbav36: Yes @tino097 , also looks like that logic was wrong anyways, `e.body["code"]` will be `invalid_api_key` on invalid api key. Have fixed that including none check

tino097: @vigbav36 yes its wrong but as its using https://github.com/openai/openai-python/blob/main/src/openai/_exceptions.py#L109
i think the correct check should be 
`if e.status_code == 401`

vigbav36: I think since they are catching `openai._exceptions.AuthenticationError` which already has defined the status as literal of 401, my understanding is it would be redundant to check here for the status_code again, please correct me if I am wrong !

tino097: @vigbav36 yes, you are right, it wont need to check as the code is already catching the AuthenticationError, so returning only the error log message would be enough.

PR #11754, opened 2025-10-16
Removed Data Catalog Persistence from MindsDB

MinuraPunchihewa:
## Description

This PR removes the persistence of the data catalog from MindsDB as it has now been moved to Minds. It is still possible to query the information schema tables, but each query will now make a request via the relevant handler. This is in contrast to before, where when an agent was created all of the relevant metadata will be loaded into the internal database and queries against the information schema will query the database.

Fixes https://linear.app/mindsdb/issue/STRC-134/remove-data-catalog-persistence-from-mindsdb

## Type of change

## Verification Process

## Additional Media:

## Checklist:

### DIFF
--- a/mindsdb/integrations/handlers/mysql_handler/__init__.py
+++ b/mindsdb/integrations/handlers/mysql_handler/__init__.py
@@ -1,7 +1,7 @@
 from mindsdb.integrations.libs.const import HANDLER_TYPE
 
 from .__about__ import __version__ as version, __description__ as description
-from .connection_args import connection_args, connection_args_example
+from .connection_args import connection_args_example, connection_args
 
 try:
     from .mysql_handler import MySQLHandler as Handler
--- a/mindsdb/integrations/handlers/mysql_handler/mysql_handler.py
+++ b/mindsdb/integrations/handlers/mysql_handler/mysql_handler.py
@@ -1,4 +1,4 @@
-from typing import Optional
+from typing import Optional, List, Dict, Any
 
 import pandas as pd
 import mysql.connector
@@ -37,7 +37,7 @@ def _map_type(mysql_type_text: str) -> MYSQL_DATA_TYPE:
         return MYSQL_DATA_TYPE.TEXT
 
 
-def _make_table_response(result: list[dict], cursor: mysql.connector.cursor.MySQLCursor) -> Response:
+def _make_table_response(result: List[Dict[str, Any]], cursor: mysql.connector.cursor.MySQLCursor) -> Response:
     """Build response from result and cursor.
 
     Args:
@@ -113,20 +113,20 @@ class MySQLHandler(MetaDatabaseHandler):
 
     name = "mysql"
 
-    def __init__(self, name, **kwargs):
+    def __init__(self, name: str, **kwargs: Any) -> None:
         super().__init__(name)
         self.parser = parse_sql
         self.dialect = "mysql"
         self.connection_data = kwargs.get("connection_data", {})
         self.database = self.connection_data.get("database")
 
-        self.connection = None
+        self.connection: Optional[mysql.connector.MySQLConnection] = None
 
-    def __del__(self):
+    def __del__(self) -> None:
         if self.is_connected:
             self.disconnect()
 
-    def _unpack_config(self):
+    def _unpack_config(self) -> Dict[str, Any]:
         """
         Unpacks the config from the connection_data by validation all parameters.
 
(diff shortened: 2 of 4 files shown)


mindsdb/interfaces/data_catalog/data_catalog_retriever.py
```diff
@@ -0,0 +1,363 @@
...
+        """
+        Construct a formatted string representation of the column statistics for a single column.
+        """
+        inner_pad = pad + " " * 4
+        inner_inner_pad = inner_pad + " " * 4
+        stats_str = f"\n{pad}- Column Statistics:"
+
+        most_common_values = stats_row.get("MOST_COMMON_VALUES")
+        most_common_frequencies = stats_row.get("MOST_COMMON_FREQUENCIES")
+        if not most_common_values.empty and most_common_values.iloc[0]:
```
ZoranPandovski: We may got AttributeError since MOST_COMMON_VALUES may be none

mindsdb/interfaces/data_catalog/data_catalog_retriever.py
```diff
@@ -0,0 +1,363 @@
...
+        Retrieve the column metadata from the handler.
+        """
+        self.logger.info(
+            f"Retrieving columns for {', '.join(self.table_names) if self.table_names else 'all'} tables for {self.database_name}"
+        )
+        response = self.data_handler.meta_get_columns(self.table_names)
+        if response.resp_type == RESPONSE_TYPE.ERROR:
+            self.logger.error(f"Failed to retrieve columns for {self.database_name}: {response.error_message}")
+            return pd.DataFrame()
+        elif response.resp_type == RESPONSE_TYPE.OK:
```
ZoranPandovski, line 342: We always return OK if we got an empty result set, right?

mindsdb/interfaces/data_catalog/data_catalog_retriever.py
```diff
@@ -0,0 +1,363 @@
...
+                    table_primary_keys_df = pd.DataFrame()
+                if not foreign_keys_df.empty:
+                    table_foreign_keys_df = foreign_keys_df[foreign_keys_df["TABLE_NAME"] == table_row["TABLE_NAME"]]
+                else:
+                    table_foreign_keys_df = pd.DataFrame()
+
+            tables_metadata_str += self._construct_metadata_string_for_table(
+                table_row,
+                table_columns_df,
+                table_column_stats_df,
```
ZoranPandovski, line 112: This may be undefined if table_columns_df is empty

mindsdb/interfaces/data_catalog/data_catalog_retriever.py
```diff
@@ -0,0 +1,363 @@
...
+    def _construct_metadata_string_for_foreign_keys(
+        self,
+        foreign_keys_df: pd.DataFrame,
+        table_name: str,
+    ) -> str:
+        """
+        Construct a formatted string representation of the foreign keys for a single table.
+        """
+        pad = " " * 4
+        foreign_keys_str = "\n\nKey Relationships:"
```
ZoranPandovski, line 306: This will always get prineted even if the foreign_keys_df get skipped

MinuraPunchihewa: @ZoranPandovski Since there is a check for an empty DataFrame (`if not foreign_keys_df.empty`) at line 147, I think this will not happen?

mindsdb/interfaces/data_catalog/data_catalog_retriever.py
MinuraPunchihewa: Yes, @ZoranPandovski. Do you see an issue with this?

mindsdb/interfaces/data_catalog/data_catalog_retriever.py
MinuraPunchihewa: You are right. I fixed this by initializing empty DataFrames first.

mindsdb/interfaces/data_catalog/data_catalog_retriever.py
MinuraPunchihewa: You are right, @ZoranPandovski. I just added more reliable parsing for both MOST_COMMON_VALUES and MOST_COMMON_FREQUENCIES.

mindsdb/interfaces/data_catalog/data_catalog_retriever.py
ZoranPandovski: Just wanted to double check in case we need to return something instead of empty df. Thanks

PR #11773, opened 2025-10-21
Tests/files nested queries

tino097:
@ea-rus added tests
Tested in the UI, i didnt find any issues

### DIFF
--- a/tests/unit/executor/test_files.py
+++ b/tests/unit/executor/test_files.py
@@ -11,6 +11,7 @@ from tests.unit.executor_test_base import BaseExecutorDummyML
 
 
 class TestFiles(BaseExecutorDummyML):
+
     def test_create_table(self):
         df = pd.DataFrame(
             [
@@ -20,37 +21,48 @@ class TestFiles(BaseExecutorDummyML):
         )
         self.set_data("table1", df)
 
-        self.run_sql("""
+        self.run_sql(
+            """
             create table files.myfile
             select * from dummy_data.table1
-        """)
+        """
+        )
 
-        self.run_sql("""
+        self.run_sql(
+            """
             create or replace table files.myfile
             select * from dummy_data.table1
-        """)
+        """
+        )
 
         ret = self.run_sql("select count(*) c from files.myfile")
         assert ret["c"][0] == 2
 
-        self.run_sql("""
+        self.run_sql(
+            """
             insert into files.myfile (
               select * from dummy_data.table1
             )
-        """)
+        """
+        )
 
         ret = self.run_sql("select count(*) c from files.myfile")
         assert ret["c"][0] == 4
 
-        self.run_sql("""
+        self.run_sql(
+            """
             insert into files.myfile (a)
             values (9)
-        """)
+        """
+        )
 
         ret = self.run_sql("select count(*) c from files.myfile")
         assert ret["c"][0] == 5
 
-    @pytest.mark.skipif(sys.platform == "win32", reason="Fixme: Open file handle somewhere makes this fail on Windows.")
+    @pytest.mark.skipif(
+        sys.platform == "win32",
+        reason="Fixme: Open file handle somewhere makes this fail on Windows.",
+    )
     def test_multipage(self):
         # copy test file because source will be removed after uloading
         source_path = Path(__file__).parent / "data" / "test.xlsx"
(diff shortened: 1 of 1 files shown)


tests/unit/executor/test_files.py
```diff
@@ -8,9 +8,88 @@
...
+            for file_name in list(self.file_controller.list_files()):
+                try:
+                    self.file_controller.delete_file(file_name)
+                except Exception:
+                    pass
+        except Exception:
+            pass
+
+    @pytest.fixture(autouse=True)
+    def create_tables_and_data(self):
```
ea-rus: these data are not used by every test, it is more test specific data
maybe is better to create separated class (with all tests that use these files) and create these files once in `setup_class`?

tests/unit/executor/test_files.py
```diff
@@ -8,9 +8,88 @@
 import pandas as pd
 
 from tests.unit.executor_test_base import BaseExecutorDummyML
+from mindsdb.utilities.exception import QueryError
 
 
 class TestFiles(BaseExecutorDummyML):
+
+    @pytest.fixture(autouse=True)
```
ea-rus: when data are created once, this method can be removed

tests/unit/executor/test_files.py
tino097: Yes, im looking into that but I've created the PR to ask if these scenarios are all valid and if i need to add or remove some. Also im looking for a option to refactor `query_dfs` into smaller helper functions.

ea-rus: ok, I will check scenarios now

tests/unit/executor/test_files.py
```diff
@@ -20,54 +99,267 @@ def test_create_table(self):
...
+        """Test dropping a non-existent table without IF EXISTS"""
+        with pytest.raises(Exception) as excinfo:
+            self.run_sql(
+                """
+                DROP TABLE files.nonexistent_table
+            """
+            )
+        assert "Can't delete table 'nonexistent_table'" in str(excinfo.value)
+
+    def test_drop_table_from_different_database(self):
```
ea-rus: it could be removed, it doesn't test files.

tests/unit/executor/test_files.py
```diff
@@ -20,54 +99,267 @@ def test_create_table(self):
...
+        assert len(result1) == 3
+
+        result2 = self.run_sql(
+            """
+            SELECT DISTINCT a1.tab_num FROM files.A5 a1
+        """
+        )
+        assert len(result2) == 3
+
+    def test_simple_select_with_database_prefix(self):
```
ea-rus: this is a simple select from file, it should be covered already (several times)

tests/unit/executor/test_files.py
```diff
@@ -20,54 +99,267 @@ def test_create_table(self):
...
+        assert len(result) == 3
+        assert sorted(result["tab_num"].tolist()) == [1, 2, 3]
+        assert sorted(result["fio"].tolist()) == ["Baratheon", "Stark", "Targaryen"]
+        assert sorted(result["city"].tolist()) == [
+            "Dragonstone",
+            "King's Landing",
+            "Winterfell",
+        ]
+
+    def test_alias_without_database_prefix(self):
```
ea-rus: what without_database_prefix means?

tests/unit/executor/test_files.py
```diff
@@ -20,54 +99,267 @@ def test_create_table(self):
...
+        result = self.run_sql(
+            """
+            SELECT * FROM files.B5 b
+        """
+        )
+
+        assert len(result) == 3
+        assert sorted(result["shop"].tolist()) == [1, 2, 3]
+
+    def test_self_join_with_aliases(self):
```
ea-rus: is it almost the same as `test_multi_table_join_with_aliases`? what is key difference?

tests/unit/executor/test_files.py
ea-rus: I made couple more comments related to scenarios themselves

tests/unit/executor/test_files.py
tino097: i was trying this in the UI
```
CREATE TABLE files.C4(id INT);
DROP TABLE files.C4;
```
and because it was valid scenario ive added the test

tests/unit/executor/test_files.py
tino097: as i understand, even we are using `files` we actually have `database` and because there was example of usage
`SELECT files.a1.tab_num` Im checking if we can get the result also as `SELECT a1.tab_num` and the database prefix i assume is the datasource/database name. Maybe i didnt get the terminology correct

tests/unit/executor/test_files.py
tino097: yes, having the more complex query working, this more simple ones can be omitted

tests/unit/executor/test_files.py
tino097: i havent used `autouse` before and i think it can be handy but i will revisit this

tests/unit/executor/test_files.py
tino097: Thanks @ea-rus i will take a look now

PR #11794, opened 2025-10-27
docs -- query kb's chunk_content column

martyna-mindsdb:
## Description

docs -- query kb's chunk_content column

related ticket: https://linear.app/mindsdb/issue/FQE-1574/add-chunk-content-as-a-possible-filter-in-kb-queries

Fixes #issue_number

## Type of change


martyna-mindsdb:
@ea-rus 
I get this error when I use `chunk_content` with one of my existing KBs. What may be the issue here?

image

ea-rus:
@martyna-mindsdb, I think you had embedding model (mindsdb openai model for example) for KB, but then if was removed
10 - is id of predictor in database

PR #11889, opened 2025-11-14
Fix selecting mixed-case objects

ea-rus:
If database was created with quotes (it is stored in mixed case):
- the name should be matched when used in queries (quoted or not) 

If database was create without quotes
 - can be accessed by mixed case if name is not quoted

Fixes https://linear.app/mindsdb/issue/FQE-1745/datasource-created-with-camelcase-cant-be-joined

## Type of change

## Verification Process

## Additional Media:

## Checklist:


mindsdb/api/executor/planner/query_planner.py
```diff
@@ -258,8 +261,14 @@ def resolve_database_table(self, node: Identifier):
...
-            if parts[0].lower() in self.databases:
-                database = parts.pop(0).lower()
+            # if not quoted  check in lower case
+            part = parts[0]
+            if part not in self.databases and not node.is_quoted[0]:
+                part = part.lower()
+
+            if part in self.databases:
+                database = part
+                parts.pop(0)
```
StpMax, line 271: will be good also pop node.is_quoted, to keep node consistent

ea-rus: thanks, added

ea-rus: no, parts is copy of node.parts, we can't change `node.is_quoted`
rolled it back

PR #11891, opened 2025-11-14
Feature/cassandra handler update

### LINKED ISSUE #11879: Integrate with python-driver for DataStax Enterprise, HCD, Astra, and open-source Apache Cassandra

Based on conversation in #11873: `scylla-driver` doesn't officially support Astra.

[scylla-driver](https://github.com/scylladb/python-driver/) is forked from [datastax/python-driver](https://github.com/datastax/python-driver), which is the [recommended Python driver](https://docs.datastax.com/en/datastax-drivers/compatibility/python-drivers.html) for Astra as well as HCD, DSE, and [OSS Cassandra](https://cassandra.apache.org/doc/latest/cassandra/getting-started/drivers.html#python).

My suggestion is that you keep `scylla-driver` for your Scylla integration, but use `python-driver` for Astra, HCD, DSE, and open-source Cassandra. 

### Anything else?

`python-driver` might also be preferred for other NoSQL/Cassandra-based databases, but I am not sure which.

tino097:
## Description

Refactor `CassandraHandler` to be parent handler and `DataStax` can inherit from it. Currently `ScyllaHandler` is not changed and in the future we can check if we can make it to inherit from `Cassandra` too.

Fixes #11879 

## Type of change

- [x] 📄 This change requires a documentation update

## Verification Process

 - [x]   Test Location: `tests/unit/handlers/test_cassandradb.py`

## Additional Media:

## Checklist:


tino097:
@aimurphy if you can please check this PR

aimurphy:
@tino097 let me ask one of the engineers who works on python-driver.

mindsdb/integrations/handlers/cassandra_handler/cassandra_handler.py
```diff
@@ -1,27 +1,261 @@
...
+        connection_props["protocol_version"] = self.connection_args.get("protocol_version", 4)
+        secure_connect_bundle = self.connection_args.get("secure_connect_bundle")
+
+        if secure_connect_bundle:
+            if secure_connect_bundle.startswith(("http://", "https://")):
+                secure_connect_bundle = self.download_secure_bundle(secure_connect_bundle)
+            connection_props["cloud"] = {"secure_connect_bundle": secure_connect_bundle}
+        else:
+            connection_props["contact_points"] = [self.connection_args["host"]]
+            connection_props["port"] = int(self.connection_args["port"])
```
absurdfarce: Semi-nit: it seems a bit strange to have Astra-specific functionality (specifically the handling of secure connect bundles) in cassandra_handler.py rather than the more specific datastax_handler.py.  I fully acknowledge I'm completely unfamiliar with the structure of this project generally so there might be some reason or set of reasons that explain this... but it wasn't clear to me on first read.

mindsdb/integrations/handlers/cassandra_handler/cassandra_handler.py
```diff
@@ -1,27 +1,261 @@
...
+
+        cluster = Cluster(**connection_props)
+        session = cluster.connect(self.connection_args.get("keyspace"))
+
+        self.is_connected = True
```
absurdfarce: This seems fine, but how do you handle a case in which the session is disconnected due to, say, network failures?  I don't see any way for this value to be set to False once it's been set to true in connect().  Again I may have missed something (or be missing something from other parts of the code).

Sometimes code like this will track instances of errors coming off of the various query methods and make educated guesses about a sessions health based on the frequency of such errors... you may want to consider something like that here.

mindsdb/integrations/handlers/cassandra_handler/cassandra_handler.py
```diff
@@ -1,27 +1,261 @@
...
         return result
+
+    def get_columns(self, table_name: str) -> Response:
+        """
+        Return columns for a single table, minimal shape:
+        - column_name
+        - type
+        """
+        q = f"DESCRIBE {table_name};"
+        return self.native_query(q)
```
absurdfarce: "DESCRIBE" actually isn't a query that'll work against Cassandra directly; when you execute "DESCRIBE" in cqlsh you're actually exercising functionality [built into cqlsh itself](https://github.com/apache/cassandra/blob/trunk/pylib/cqlshlib/cqlshmain.py#L1104).  Since you're going through the Python driver directly here (and not cqlsh) this query won't be understood.

mindsdb/integrations/handlers/cassandra_handler/cassandra_handler.py
```diff
@@ -1,27 +1,261 @@
...
-        Get the list of tables in the connected Cassandra database.
+        Get a list of all tables in the current keyspace.
+        """
+
+        query = """
+            SELECT
+                keyspace_name AS table_schema,
+                table_name
+            FROM system_schema.tables
+        """
```
absurdfarce: Note that the Python driver queries this information when it establishes a connection to a cluster and stores it in [cassandra.metadata](https://github.com/datastax/python-driver/blob/3.29.3/cassandra/metadata.py) (by way of [this code](https://github.com/datastax/python-driver/blob/3.29.3/cassandra/cluster.py#L3743)).  That looks to be true for several other queries that are executed here... you should be able to get the same information directly from your Cluster's metadata.

absurdfarce (commented):
Review provided at the request of @aimurphy

mindsdb/integrations/handlers/cassandra_handler/cassandra_handler.py
tino097: Thanks, this is really helpful, i was not sure if this is only Astra specific or it would apply everywhere. I've done some inheritance changes and left this part here. I will move it into [DataStax](mindsdb/integrations/handlers/datastax_handler/datastax_handler.py) handler as  it is more suitable

mindsdb/integrations/handlers/cassandra_handler/cassandra_handler.py
tino097: As the datasource is used to read (in most cases) data, keeping simple reconnects on demand are just fine.

mindsdb/integrations/handlers/cassandra_handler/cassandra_handler.py
tino097: Thanks again, i will revisit this again

mindsdb/integrations/handlers/cassandra_handler/cassandra_handler.py
tino097: Thanks for your comments, really helpful @absurdfarce

mindsdb/integrations/handlers/cassandra_handler/cassandra_handler.py
```diff
@@ -1,27 +1,282 @@
...
+        :return: success status and error message if error occurs
+        """
+        try:
+            session = self.connect()
+            session.execute("SELECT release_version FROM system.local")
+            return StatusResponse(success=True)
+        except Exception as e:
+            return StatusResponse(success=False, error_message=str(e))
+
+    def prepare_response(self, resp):
```
StpMax, line 97: it can be static method

mindsdb/integrations/handlers/cassandra_handler/cassandra_handler.py
```diff
@@ -1,27 +1,282 @@
...
+        q = "SELECT table_name FROM system_schema.tables WHERE keyspace_name = '{}';".format(
+            self.connection_args["keyspace"]
+        )
+        return self.native_query(q)
+
+    def meta_get_columns(self, table_name) -> Response:
+        """
+        Show details about the table
+        """
+        q = "SELECT column_name, kind, position, type FROM system_schema.columns WHERE keyspace_name = '{}' AND table_name = '{}';".format(
```
StpMax, line 254: The column list in the response is not that expected from `meta_get_columns`. In the docstring there are:
```
            - TABLE_NAME (str): Name of the table.
            - COLUMN_NAME (str): Name of the column.
            - DATA_TYPE (str): Data type of the column, e.g. 'VARCHAR', 'INT', etc.
            - COLUMN_DESCRIPTION (str): Description of the column (optional).
            - IS_NULLABLE (bool): Whether the column can contain NULL values (optional).
            - COLUMN_DEFAULT (str): Default value of the column (optional).
```

mindsdb/integrations/handlers/datastax_handler/datastax_handler.py
```diff
@@ -10,3 +20,101 @@ class DatastaxHandler(ScyllaHandler):
...
+            if secure_connect_bundle.startswith(("http://", "https://")):
+                secure_connect_bundle = self.download_secure_bundle(
+                    secure_connect_bundle
+                )
+            connection_props["cloud"] = {"secure_connect_bundle": secure_connect_bundle}
+        else:
+            connection_props["contact_points"] = [self.connection_args["host"]]
+            connection_props["port"] = int(self.connection_args["port"])
+
+        cluster = Cluster(**connection_props)
```
StpMax, line 115: `cluster` is not saved to `self`, therefore `get_tables` and `get_columns` will fail

tests/unit/handlers/test_cassandradb.py
```diff
@@ -0,0 +1,408 @@
...
+
+    @property
+    def get_columns_query(self):
+        pass
+
+    def test_connect_with_secure_bundle(self):
+        """Test connection using secure connect bundle (Astra DB)"""
+        connection_data = self.dummy_connection_data.copy()
+        connection_data["secure_connect_bundle"] = "/path/to/bundle.zip"
+        handler = CassandraHandler("cassandra", connection_data=connection_data)
```
StpMax, line 388: should it he datastax handler?

mindsdb/integrations/handlers/cassandra_handler/cassandra_handler.py
```diff
@@ -1,27 +1,282 @@
-from mindsdb.integrations.handlers.scylla_handler import Handler as ScyllaHandler
+import tempfile
```
StpMax, line 1: it is not used

mindsdb/integrations/handlers/cassandra_handler/cassandra_handler.py
```diff
@@ -1,27 +1,282 @@
...
+                if isinstance(v, Date):
+                    v = v.date()
+                row2[k] = v
+            data.append(row2)
+        return data
+
+    def native_query(self, query: str) -> Response:
+        """
+        Receive SQL query and runs it
+        :param query: The SQL query to run in MySQL
```
StpMax, line 112: `cassandra`, not `MySQL`

tino097:
I will revisit this, lets keep this PR open for some time

PR #11904, opened 2025-11-18
Optimize inner join

ea-rus:
## Description

Optimization for inner join when the query has limit but we don't know how many rows to fetch from first table to have enought record in results. 
Sample query:
```sql
    SELECT * FROM db1.statuses as t1
     join db2.tasks as t2 on t1.id=t2.status_id
    limit 2
```

The Optimization algorithm:
- Fetch limit*2 records from the first table (statuses)
- Get list of values for filtering second table (list of statuses.id)
- Fetch from the second table (tasks) using IN filter with values from previous step
- Do the join
- If we don't have enough records: increase limit and repeat the circle above

Limit calculation:
- the first run: limit  * 2
After it we do a 3 tries with limit:
- if we have X < limit results from the previous try and add an extra records depending on number of try (to reduce a chance of having the one more circle):
   - forecast required value: limit = prev_limit * required_records / X * try_num + 10**try_num
- if previous result is empty:
  -  limit = prev_limit*10

When total tries is 4 (1 + 3): do the last try with limit = null

Mechanizm 
FetchDataframePartition is used with `condition` attribute. It contains the first table query in `query` attribute and the subsequent queries and joins in `steps` attrubute. 
When `limit` is set in `condition`: we will execute `steps` several times with different result on input 

Fixes https://linear.app/mindsdb/issue/FQE-1749/left-join-optimization

## Type of change

## Verification Process

## Additional Media:

## Checklist:


mindsdb/api/executor/sql_query/steps/fetch_dataframe_partition.py
```diff
@@ -87,6 +95,56 @@ def call(self, step: FetchDataframeStepPartition) -> ResultSet:
...
+
+            if try_num > 3:
+                # the last try without the limit
+                first_table_limit = None
+                continue
+
+            # no enough results
+            if len(result) > 0:
+                # forecast the required limit (depending on how much row we don't have)
+                first_table_limit = first_table_limit * limit / len(result) * try_num + 10**try_num
```
StpMax: first_table_limit must be rounded here, instead of line 113
Otherwise check in line 122 may be wrong (for example `len([1,2,3]) < 3.1` )

mindsdb/api/executor/sql_query/steps/fetch_dataframe_partition.py
```diff
@@ -87,6 +95,56 @@ def call(self, step: FetchDataframeStepPartition) -> ResultSet:
...
+                query2.limit = Constant(int(first_table_limit))
+            else:
+                query2.limit = None
+
+            response = dn.query(query=query2, session=self.session)
+            df = response.data_frame
+
+            result = self.exec_sub_steps(df)
+
+            if len(result) >= limit or len(df) < first_table_limit or first_table_limit is None:
```
StpMax: in last try here will be check `len(df) < None`

mindsdb/api/executor/sql_query/steps/fetch_dataframe_partition.py
ea-rus: yes, it can be, I will update

mindsdb/api/executor/sql_query/steps/fetch_dataframe_partition.py
ea-rus: nice catch

PR #11923, opened 2025-11-21
ChromaDB: clear storage on KB deletion

StpMax:
## Description

There was a bug: if create/delete KB, then data keep exists on the disk. The fix is - delete persistent storage on 'table' deletion

Fixes #FQE-1786

## Type of change

## Verification Process

## Additional Media:

## Checklist:


mindsdb/integrations/handlers/chromadb_handler/chromadb_handler.py
```diff
@@ -471,6 +472,8 @@ def drop_table(self, table_name: str, if_exists=True):
         self.connect()
         try:
             self._client.delete_collection(table_name)
+            if "persist_directory" in self._client_config:
```
ea-rus: persist_directory is always in self._client_config. is better to check if it is none

also, I think we can do only for folders that are in self.handler_storage
because you can use local folder and create two collection (tables) there
after deletion of the first table the second shouln't be deleted

PR #11973, opened 2025-12-05
fix: Optimize get_integration by fetching directly from DB

SyedaAnshrahGillani:
The previous implementation of 'get_integration' loaded all databases into memory and iterated through them to find the matching integration. This was inefficient and could lead to performance issues on systems with a large number of databases.

This commit refactors 'get_integration' to use the 'integration_controller.get_by_id' method, which directly queries the database for the integration. This significantly improves performance by avoiding unnecessary data loading and iteration.


SyedaAnshrahGillani:

PR #11996, opened 2025-12-10
Test coverage for integrations and  action for verified handlers 

ZoranPandovski:
## Description

This PR enables us to measure test coverage for the verified handlers. For now, this is enabled for following handlers:
1. MySQL
2. PostgreSQL
3. Snowflake
4. TimescaleDB
5. Salesforce
6. Microsoft SQL Server
7. Slack
8. File
9. BigQuey
10. Redshift
 and adds more tests for them. In the subsequent PRs, we will increase this to all verified handlers we have. If the percentage fails under 80% for any handler, the action will fail.

>Note: Statsforecast failing test was removed, we are deprecating it in the next release.

## Type of change

## Checklist:

### DIFF
--- a/.github/workflows/tests_unit.yml
+++ b/.github/workflows/tests_unit.yml
@@ -151,7 +151,9 @@ jobs:
           for handler in "${INSTALL_HANDLERS[@]}"; do
             HANDLER_EXTRAS+=(".[${handler}]")
           done
-          uv pip install .[agents, kb] -r requirements/requirements-test.txt "${HANDLER_EXTRAS[@]}"
+          uv pip install ".[agents,kb]" \
+            -r requirements/requirements-test.txt \
+            "${HANDLER_EXTRAS[@]}"
           # Onuxruntime is required for ChromaDB, once we have default pgvector we can remove it
           uv pip install --force-reinstall onnxruntime==1.20.1
           git clone --branch v$(uv pip show mindsdb_sql_parser | grep Version | cut -d ' ' -f 2) https://github.com/mindsdb/mindsdb_sql_parser.git parser_tests


.github/workflows/tests_unit.yml
```diff
@@ -24,15 +55,13 @@ jobs:
...
-          python-version: ${{ vars.CI_PYTHON_VERSION || '3.11' }}      # Default to 3.11 where vars aren't available (PRs from forks)
-
+          cache-local-path: "/home/runner/_work/_tool/uv-local-cache" # Place cache in the tool dir because we mount this in our runnners
+          prune-cache: false # We want to save all cache because it's in the mount^
+          python-version: ${{ vars.CI_PYTHON_VERSION || '3.11' }} # Default to 3.11 where vars aren't available (PRs from forks)
       # Checks the codebase for print() statements and fails if any are found
       # We should be using loggers instead
       - name: Check for print statements
         run: |
-          # The pyproject file confuses uv: https://github.com/astral-sh/uv/issues/6838
```
hamishfagg, line 35: We need to keep this comment - It's explaining the next line which is not self-explanatory

.github/workflows/tests_unit.yml
```diff
@@ -86,57 +113,61 @@ jobs:
...
+            INSTALL_HANDLERS+=("${handler}")
+          done <<< "${HANDLERS_TO_INSTALL}"
+          HANDLER_EXTRAS=()
+          for handler in "${INSTALL_HANDLERS[@]}"; do
+            HANDLER_EXTRAS+=(".[${handler}]")
+          done
+          uv pip install . -r requirements/requirements-test.txt "${HANDLER_EXTRAS[@]}"
+          uv pip install -r requirements/requirements-agents.txt
+          uv pip install -r requirements/requirements-kb.txt
+          uv pip install --force-reinstall onnxruntime==1.20.1
```
hamishfagg: We need to install all of these in one command, because otherwise pip will uninstall things from previous commands to satisfy the current one.
```suggestion
uv pip install . .[agents] .[kb] "${HANDLER_EXTRAS[@]}" -r requirements/requirements-test.txt
```
Or we could add `agents` and `kb` to `HANDLERS_TO_INSTALL`

What's the onnxruntime install about? It seems like a hack

mindsdb/integrations/handlers/mssql_handler/mssql_handler.py
```diff
@@ -300,7 +300,12 @@ def disconnect(self):
 
         if not self.is_connected:
             return
-        self.connection.close()
+        if self.connection is not None:
+            try:
+                self.connection.close()
+            except Exception:
+                pass
```
hamishfagg, line 308:
```suggestion
            try:
                self.connection.close()
            except Exception:
                logger.exception("Failed to close connection:")
                pass
```

hamishfagg (changes requested):

.github/workflows/tests_unit.yml
ZoranPandovski: I will make sure we have all comments

.github/workflows/tests_unit.yml
ZoranPandovski: Good point, adding one liner. About onuxruntime I will add comment, this is required for ChromaDB. In the next release we will migrate to pgvector as default KB so we can remove this

PR #11962, opened 2025-12-03
MindsDB Lite version

ZoranPandovski:
## 🚧 Migration Notes

No breaking API changes at runtime for users of the Docker or Cloud editions.
Standalone users installing via pip install mindsdb may need to update requirements if they rely on features now behind extras. Check the description for more info :point_down: 

## Description

Initial version of the `MindsDB Lite` a slimmer distribution intended to support the core Federated Query Engine while allowing optional features to be installed via Python extras.. Note once this PR is merged it should be merged into #11971 
1. New lightweight mindsdb package layout
 * The default installation now includes only dependencies required for the Federated Query Engine.
 * Additional capabilities (Agents, Knowledge Base, MCP, A2A protocol, etc.) are now available via extras:
     ```
     pip install "mindsdb[mcp,a2a]"
     pip install "mindsdb[agents]"
     ```
2. The file handler is now the only pre-installed handler in the base package. In Docker images, MySQL and PostgreSQL handlers remain bundled out of the box.
3. Non-default handlers (web, evaluator, agents, etc.) are now lazily imported to avoid unnecessary dependency loading. The RAG handler underwent significant import restructuring, but it is unused due to KB .It will be fully removed/deprecated in the follow-up PR (#11971).
5. Official Docker images remain fully functional: all dependencies required for Agents, Knowledge Base, and integrations are still installed.
6.Removed Hugging Face–specific Docker images and related tests. We now only provide the handler, without shipping Hugging Face–powered images by default.
7. MCP and A2A servers are now conditionally enabled if the required dependencies are installed via extras (examples above).
8. Cloud images remain unchanged functionally. Unused dependencies (OneDrive, RAG, DynamoDB, Hugging Face, etc.) have been removed to reduce footprint.
9. Shared libraries previously bundled globally (e.g., forecasting libs, scikit-learn, transformers) have been moved into the handlers/features that actually require them.

Fixes: https://linear.app/mindsdb/issue/FQE-1796/mindsdb-lite-pypi-package

## Type of change

(Please delete options that are not relevant)

- [ ] 🐛 Bug fix (non-breaking change which fixes an issue)
- [X] ⚡ New feature (non-breaking change which adds functionality)
- [X] 📢 Breaking change (fix or feature that would cause existing functionality not to work as expected)
- [X] 📄 This change requires a documentation update

## Verification Process

To ensure the changes are working as expected:

 - [ ]   Test Location: Specify the URL or path for testing.
[...]

PR #12005, opened 2025-12-12
Release 25.13.0

ea-rus:
Release 25.13.0

### DIFF
--- a/docker/mindsdb.Dockerfile
+++ b/docker/mindsdb.Dockerfile
@@ -94,7 +94,7 @@ EXPOSE 47334/tcp
 EXPOSE 47335/tcp
 
 # Pre-load tokenizer from Huggingface, and UI
-#RUN python -m mindsdb --config=/root/mindsdb_config.json --load-tokenizer --update-gui
+RUN python -m mindsdb --config=/root/mindsdb_config.json --load-tokenizer --update-gui
 
 # Same as extras image, but with dev dependencies installed.
 # This image is used in our docker-compose


.github/workflows/build_deploy_dev.yml
```diff
@@ -2,6 +2,7 @@ name: Build and deploy to dev
 
 permissions:
   contents: read
+  pull-requests: write
```
mindsdb-devops, line 5: What is this for?

docker/mindsdb.Dockerfile
```diff
@@ -94,7 +94,7 @@ EXPOSE 47334/tcp
 EXPOSE 47335/tcp
 
 # Pre-load tokenizer from Huggingface, and UI
-RUN python -m mindsdb --config=/root/mindsdb_config.json --load-tokenizer --update-gui
+#RUN python -m mindsdb --config=/root/mindsdb_config.json --load-tokenizer --update-gui
```
mindsdb-devops: This also I'm curious this means the docker image is smaller, but then the initial startup time is longer if we're always downloading these by default on first run.  Maybe we want to accept this, but in the past, the startup time was important.  Maybe we changed some logic I'm unaware of and don't need to pre-load these aka we don't use them at all any more?

mindsdb-devops (commented):
Review comments please

.github/workflows/build_deploy_dev.yml
ea-rus: It was in this PR:
https://github.com/mindsdb/mindsdb/pull/11996/files#diff-b721cbc7a14e2466d10171c10fb0dde7134fe2fec48f24e19cf80d44a3482d18R5

docker/mindsdb.Dockerfile
ea-rus: The source PR is: [MindsDB Lite version ](https://github.com/mindsdb/mindsdb/pull/11962/files#diff-982eb86076d1a97116d39c22566f4b91f59f8b15ed49470f1a193be560936784R97) 
@ZoranPandovski is this update required

mindsdb-devops: Yeah, that's what I suspected, but I don't see anyone commenting/thinking about this other issue, about the effect on startup-time.  Maybe no one has noticed, or maybe the "time savings" you'd get from the initial download being smaller, is then lost by having to download things on-the-fly during first start.  I'd like someone to closely look at this before we merge and at least think about and answer this.

Basically a few questions to answer:
* Ignoring the time for the initial download of an image, does the startup time of the container now take longer due to having to download some resources on-the-fly during startup?
* If yes, do we want to accept this, or improve this before we merge/release this? 

I ask because I know certain customers are very particular about speed and run time, and this has a potential effect to increase the runtime in favor of a smaller download time.  But, once an image is downloaded, if you need to re-run the container multiple times you have to keep re-paying this initial download time-savings over and over as a runtime download.  Make sense?

Also, if we refactored enough to never need to download things initially on startup then, why is this just commented out and the line not removed?  That's what I'd expect a developer to do if this was completely not necessary any more.  That's why my instinct here is to hold up and ask for review/insight into this process.

.github/workflows/build_deploy_dev.yml
mindsdb-devops: Understood, needed for the test coverage comment logic to work, thanks.  Resolving...

docker/mindsdb.Dockerfile
hamishfagg: I left a comment on the original PR that this should be left uncommented.

PR #11855, opened 2025-11-05
Files API: [PUT] Check the content of the json payload

tino097:
## Description

Small refactor of the code with adding helpers 
Added `safe_extract` for zipfiles
Updated the test scenarios 

(Please delete options that are not relevant)

- [x] 🐛 Bug fix (non-breaking change which fixes an issue)
- [x] 📄 This change requires a documentation update

## Verification Process

To ensure the changes are working as expected:

 - [x]   Test Location: `tests/unit/api/http/files_test.py`

## Checklist:

- [x] My code follows the style guidelines(PEP 8) of MindsDB.
- [x] I have appropriately commented on my code, especially in complex areas.
- [x] Relevant unit and integration tests are updated or added.

PR #12010, opened 2025-12-15
Bugfix/files safe extract

tino097:
This PR is followup of the #11855


mindsdb/utilities/fs.py
```diff
@@ -181,15 +184,24 @@ def __is_within_directory(directory, target):
...
+        if hasattr(archivefile, "data_filter"):
+            archivefile.extractall(path, members=members, numeric_owner=numeric_owner, filter="data")
+            return
+
+        # for py < 3.12
+        for member in archivefile.getmembers():
+            member_path = os.path.join(path, member.name)
+            if not __is_within_directory(path, member_path):
+                raise Exception("Attempted Path Traversal in Tar File")
+        archivefile.extractall(path, members=members, numeric_owner=numeric_owner)
```
StpMax, line 317: will be good to raise an error if it is not tar or zip

tests/unit/api/http/files_test.py
```diff
@@ -129,3 +130,25 @@ def test_archive_file_with_extension_upload(client):
     assert response.status_code == 400
     data = response.get_json()
     assert "File name cannot contain extension." in data["detail"]
+
+
+def test_zipfile_traversal(client):
+    """Test uploading a zip archive with path traversal filenames"""
+    import zipfile
+    import io
```
StpMax, line 179: better to import at the top

mindsdb/utilities/fs.py
tino097: as i have checked, we are already doing it before calling this method

hamishfagg (approved):
Same as other PRs, comment in check_requirements.py

PR #12038, opened 2025-12-23
fix(s3_handler): handle None location constraint as us-east-1

kelvinvelasquez-SDE:
### Description
Fixes issue #11886 where S3 handler fails for buckets in `us-east-1` because `get_bucket_location` returns `None`.

### Changes
- Added check in `s3_handler.py` to default to `us-east-1` if location is `None`.

### Verification
- Reproduced the issue with a test case mocking `None` response.
- Verified that the fix correctly sets `s3_region` to `us-east-1`.


kelvinvelasquez-SDE:
addressed feedback: added explicit comment clarifying that AWS returns None for us-east-1.

kelvinvelasquez-SDE:
Closing this PR due to inactivity. I'll revisit this if it becomes relevant again. Thanks! 🚀

PR #12100, opened 2026-01-14
Make `track_column` parameter case-insensitive

StpMax:
## Description

If table has `id` column in lower case, and try to run query with `track_column` in differ case, then error appear:
```sql
INSERT INTO kb
SELECT id, name
FROM local_pg.data
USING batch_size=2, track_column=ID;  
```

Fixes #FQE-1965

## Type of change

## Verification Process

## Additional Media:

## Checklist:


mindsdb/interfaces/query_context/context_controller.py
```diff
@@ -18,6 +18,30 @@
...
+    Args:
+        columns: list of column names
+        name: name of the column to search for (case-insensitive)
+    Returns:
+        str: the actual column name from the list
+    Raises:
+        ValueError: if column is not found or multiple columns match
+    """
+    lower_names = [x.lower() for x in columns]
+    count = [x.lower() for x in columns].count(name.lower())
```
ea-rus: `count = lower_names.count(name.lower())`

ea-rus: also should we use logic like?:
- search case-dependent.
- if not found: search case-independent

StpMax: In that case, the behavior may not be obvious. Assuming there is a table with two columns with the same name but in different cases "test" and "TEST", and the query is as follows:
```sql
insert into kb select * from tbl using track_column=TEST
```
i would assume that "test" in lowercase would be used, as the column name in the query is not enclosed in quotes. But, in fact, in proposed logic uppercase column will be used. This is not obvious.
Anyway, having two columns with the same name but in different cases is a very rare case.

PR #12113, opened 2026-01-18
fix(ollama): auto-detect input column when prompt_template is missing

Sriram-B-Srivatsa:
### Description
Currently, the Ollama handler relies on a strict contract: the input column must be named `text`, or a `prompt_template` must be explicitly provided. If a user queries with a different column name (e.g., `question`) and no template, the handler crashes with a `KeyError` inside `get_completed_prompts`.

This PR improves the UX by adding a smart fallback.

### Changes
- Updated `predict()` in `ollama_handler.py`.
- Added logic to check if `prompt_template` is missing AND `text` column is missing.
- If the input DataFrame has exactly one column, the handler now automatically uses that column as the input variable.

### Verification
**Environment:** Local MindsDB + Ollama (llama3.2)

**SQL Used:**
```sql
CREATE MODEL my_local_llama
PREDICT response
USING
    engine = 'ollama',
    model_name = 'llama3.2';

-- Query with 'question' column (previously caused crash)
SELECT response
FROM my_local_llama
WHERE question = 'Explain quantum physics in 5 words';
```
### Result
**Before:** ```KeyError: "None of [Index(['text'], dtype='object')] are in the [columns]"```
**After:** `Successfully` returns generated `text` (e.g., "Particles exist in multiple states")

### DIFF
--- a/.github/workflows/build_deploy_staging.yml
+++ b/.github/workflows/build_deploy_staging.yml
@@ -10,7 +10,7 @@ on:
     types:
       - closed
     branches:
-      - 'develop'
+      - 'main'
       - 'releases/*'
 
 concurrency:
--- a/docker/mindsdb.Dockerfile
+++ b/docker/mindsdb.Dockerfile
@@ -93,6 +93,8 @@ ENV PATH=/venv/bin:$PATH
 EXPOSE 47334/tcp
 EXPOSE 47335/tcp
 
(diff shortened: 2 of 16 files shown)


Sriram-B-Srivatsa:

Sriram-B-Srivatsa:
recheck

ea-rus:
@Sriram-B-Srivatsa, I merged #12114, now this PR has conflicts. Could you solve them?

PR #12114, opened 2026-01-18
feat(ollama): add support for temperature parameter 

Sriram-B-Srivatsa:
### Description
Previously, the Ollama handler did not support passing the `temperature` parameter, locking all generations to the default model creativity. This PR extracts the `temperature` argument from `pred_args` (or model `args`) and injects it into the Ollama API options.

This allows users to control the determinism vs. creativity of the LLM directly from SQL.

### Changes
- Modified `predict()` in `ollama_handler.py`.
- Added logic to extract `temperature` from `pred_args`.
- Validates that `temperature` is a float before passing it to the `options` dictionary in the API payload.

### Verification
**Environment:** Local MindsDB + Ollama (llama3.2)

**SQL Used:**
```sql
-- 1. Deterministic (Robot)
SELECT response
FROM my_smart_test
WHERE text = 'Write a poem about Python'
USING temperature = 0.0;

-- 2. Creative (Poet)
SELECT response
FROM my_smart_test
WHERE text = 'Write a poem about Python'
USING temperature = 0.9;
```

### Result
- Validated that the request payload to Ollama includes `"options": {"temperature": 0.9}`
- Model generates different outputs based on the parameter.

### DIFF
--- a/mindsdb/integrations/handlers/ollama_handler/__about__.py
+++ b/mindsdb/integrations/handlers/ollama_handler/__about__.py
@@ -1,9 +1,9 @@
-__title__ = 'MindsDB Ollama handler'
-__package_name__ = 'mindsdb_ollama_handler'
-__version__ = '0.0.1'
+__title__ = "MindsDB Ollama handler"
+__package_name__ = "mindsdb_ollama_handler"
+__version__ = "0.0.1"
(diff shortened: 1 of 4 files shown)


ea-rus:
@Sriram-B-Srivatsa, can you fix static code errors?

PR #12133, opened 2026-01-22
feat(chromadb): enable native semantic search by text content

Sriram-B-Srivatsa:
### Description
Currently, the ChromaDB handler supports vector search if `search_vector` is provided, but it does not support native text-to-vector search (Semantic Search). If a user queries `WHERE content = 'some text'`, the handler defaults to an exact match lookup (`.get()`), which is often not the intended behavior for a vector database.

This PR upgrades the `select` method to utilize ChromaDB's native `query_texts` feature when filtering by content.

### Changes
- Modified `select()` in `chromadb_handler.py`.
- Added logic to detect `content` filters in the SQL condition.
- If a content filter is found, it triggers `collection.query(query_texts=...)` instead of `collection.get()`.
- Added unit tests to verify that `upsert` handles embeddings correctly and `select` triggers the correct query mode.

### Verification
**Unit Tests:**
Added `test_chromadb_handler.py` covering:
1. `test_insert_calls_upsert`: Verifies embeddings are passed to the database.
2. `test_select_semantic_search`: Verifies that querying by content triggers semantic search.

### DIFF
--- a/.github/workflows/build_deploy_staging.yml
+++ b/.github/workflows/build_deploy_staging.yml
@@ -10,7 +10,7 @@ on:
     types:
       - closed
     branches:
-      - 'develop'
+      - 'main'
       - 'releases/*'
 
 concurrency:
--- a/docs/package-lock.json
+++ b/docs/package-lock.json
@@ -5043,9 +5043,9 @@
       "license": "MIT"
     },
     "node_modules/brace-expansion": {
-      "version": "1.1.12",
-      "resolved": "https://registry.npmjs.org/brace-expansion/-/brace-expansion-1.1.12.tgz",
-      "integrity": "sha512-9T9UjW3r0UW5c1Q7GTwllptXwhvYmEzFhzMfZ9H7FQWt+uZePjZPjBP/W1ZEyZ1twGWom5/56TF4lPcqjnDHcg==",
+      "version": "1.1.13",
+      "resolved": "https://registry.npmjs.org/brace-expansion/-/brace-expansion-1.1.13.tgz",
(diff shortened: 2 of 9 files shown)


Sriram-B-Srivatsa:
Hi team, just a quick update: I've verified this locally against the latest main branch and the new semantic search logic is working as expected with the updated unit tests. Standing by for feedback whenever you have a moment during the GSoC org application season. Thanks!

ea-rus:
@Sriram-B-Srivatsa, can you fix static code errors?

ea-rus:
The error is:
```
ruff format..............................................................Failed
Would reformat: mindsdb/integrations/handlers/chromadb_handler/chromadb_handler.py
Would reformat: mindsdb/integrations/handlers/chromadb_handler/tests/test_chromadb_handler.py
```

mindsdb/integrations/handlers/ollama_handler/__about__.py
```diff
@@ -1,9 +1,9 @@
-__title__ = 'MindsDB Ollama handler'
-__package_name__ = 'mindsdb_ollama_handler'
-__version__ = '0.0.1'
+__title__ = "MindsDB Ollama handler"
```
ea-rus, line 1: by some reason ollama handler was included in this pr. should it be here?

ea-rus (changes requested):
also still code style problems
you can solve the by running 
`ruff format  mindsdb/integrations/handlers/chromadb_handler/tests/test_chromadb_handler.py`
locally

Sriram-B-Srivatsa:
I tried running ruff format command, it says files unchanged, still I will try once again

mindsdb/integrations/handlers/ollama_handler/__about__.py
Sriram-B-Srivatsa: My bad, this was supposed to be a fix for the PR #12114 which was done here unknowingly

PR #12140, opened 2026-01-23
Adds user id scope to resources

lucas-koontz:
## Description

This pull request adds a `user_id` scope to various database tables and resources, ensuring data isolation and access control at the user level. It also migrates all current data to have an associated `user_id`. Finally, it makes the `company_id` non-nullable in the database.

Fixes STRC-539

## Type of change

## Verification Process

 - [ ]   Test Location: After deployment, check data isolation across different users within the same company.
 - [ ]   Verification Steps:
   1.  Create resources (e.g., integrations, models, files) as one user and verify that they are not visible or accessible to other users within the same company.
   2.  Run SQL queries that target these resources and validate that only the user who created them can access them.

## Additional Media:

## Checklist:

### DIFF
--- a/mindsdb/integrations/handlers/snowflake_handler/snowflake_handler.py
+++ b/mindsdb/integrations/handlers/snowflake_handler/snowflake_handler.py
@@ -484,8 +484,9 @@ class SnowflakeHandler(MetaDatabaseHandler):
             query += f" AND TABLE_NAME IN ({table_names_str})"
 
         result = self.native_query(query)
-        result.data_frame["ROW_COUNT"] = result.data_frame["ROW_COUNT"].astype(int)
-
+        if result.data_frame is not None and "ROW_COUNT" in result.data_frame.columns:
+            # Snowflake can return NULL for ROW_COUNT (e.g., for views); preserve as <NA>.
+            result.data_frame["ROW_COUNT"] = result.data_frame["ROW_COUNT"].astype("Int64")
         return result
 
     def meta_get_columns(self, table_names: Optional[List[str]] = None) -> Response:


lucas-koontz:
I'm going to mark all [entelligence-ai-pr-reviews](https://github.com/apps/entelligence-ai-pr-reviews) comments that are incorrect with a  👎  emoji. Some comments does not apply or it's out of the scope for this PR.

mindsdb/utilities/ml_task_queue/consumer.py
```diff
@@ -172,8 +173,11 @@ def _listen(self) -> None:
             task_type = ML_TASK_TYPE(message_content[b"task_type"])
             model_id = int(message_content[b"model_id"])
             company_id = message_content[b"company_id"]
-            if len(company_id) == 0:
-                company_id = None
+            user_id = message_content[b"user_id"]
+
+            company_id = company_id if len(company_id) > 0 else DEFAULT_COMPANY_ID
+            user_id = user_id if len(user_id) > 0 else DEFAULT_USER_ID
```
StpMax: this is unnecessary, as company_id and user_id always set

mindsdb/utilities/ml_task_queue/producer.py
```diff
@@ -59,7 +60,8 @@ def apply_async(self, task_type: ML_TASK_TYPE, model_id: int, payload: dict, dat
             redis_key = RedisKey.new()
             message = {
                 "task_type": task_type.value,
-                "company_id": "" if ctx.company_id is None else ctx.company_id,  # None can not be dumped
+                "company_id": ctx.company_id if ctx.company_id else DEFAULT_COMPANY_ID,
```
StpMax: i guess that ctx.company_id and ctx.user_id always set, so if/else is unnecessary

mindsdb/api/common/middleware.py
```diff
@@ -99,6 +99,6 @@ def check_auth(username, password, scramble_func, salt, company_id, config):
             return {"success": False}
 
         logger.info(f"Check auth, user={username}: Ok")
-        return {"success": True, "username": username}
+        return {"success": True, "username": username, "company_id": company_id, "user_id": user_id}
```
StpMax, line 102: do we need to return company_id and user_id here?

mindsdb/api/mysql/mysql_proxy/mysql_proxy.py
```diff
@@ -465,8 +465,12 @@ def is_cloud_connection(self):
             client_capabilities = self.request.recv(8)
             client_capabilities = struct.unpack("L", client_capabilities)[0]
 
-            company_id = self.request.recv(4)
-            company_id = struct.unpack("I", company_id)[0]
+            size_str = "16"  # 16 bytes of null-terminated string
```
StpMax, line 468: Note: it will require changes in cloud code if/when mysql api will be used there.

mindsdb/api/mysql/mysql_proxy/mysql_proxy.py
```diff
@@ -465,8 +465,12 @@ def is_cloud_connection(self):
...
+            company_id = self.request.recv(size_str)
+            company_id = company_id.decode().strip("\x00") if len(company_id) > 0 else DEFAULT_COMPANY_ID
```
StpMax: if/else are unnecessary here, as len() always will be > 0

mindsdb/migrations/versions/2026-01-12_f64112749455_add_user_id_field.py
```diff
@@ -0,0 +1,171 @@
...
+        except Exception:
+            logger.exception(f"Failed to create constraint {constraint_name} for table {table_name}")
+            raise
+
+    # Remove user_id column and revert company_id to nullable for all tables
+    for table_name in TABLES_WITH_USER_ID:
+        with op.batch_alter_table(table_name, schema=None) as batch_op:
+            batch_op.drop_column("user_id")
+            # Revert company_id to nullable without default
+            batch_op.alter_column("company_id", existing_type=sa.String(), nullable=True, server_default=None)
```
StpMax, line 491: to fully downgrade, it may be valuable also set it to null/0

mindsdb/api/common/middleware.py
lucas-koontz: I'm not sure, I added because the company_id was an argument in this method.

mindsdb/api/mysql/mysql_proxy/mysql_proxy.py
lucas-koontz: I think this will have the same behavior for unpacking. I'm setting to 16 bytes now because we are using UUID as identifies.

PR #12156, opened 2026-01-27
Make KB Insert Batch by default

ea-rus:
## Description

This PR enables auto-batching for inserts into knowledge bases from selects. It finds id_column in KB settings and convert query to partitioned with track_column = id_column (and query is run async)
If track column doesn't exist in source table - it doesn't run query in async mode

Implementation:

**To enable batching by default for inserts into knowledge bases**

1. Update config
- Add parameters:
  - `knowledge_bases.disable_pgvector_autobatch` - default true. It disables auto batching for KBs with pgvector 
  - `knowledge_bases.disable_autobatch` - default false. It is used in unit tests and keeps insert in KBs in sync mode
2. Update executor:
- When planner is initialized
  - Collect information about all knowledge bases  in query and its verctor DBs
  - Pass info about KBs and its storages to planner
3. Update planner:
- If 
  - this is an insert to a knowledge base from select:
  - Knowledge base storage is not “pgvector” or disable_pgvector_autobatch
- Then
  - Convert query to “run in partitions”, parameters:
  - Get “track_column” from “id_column” parameter of knowledge base 
  - Use batch_size = 1000 (we can choose different constant)

**Failback if source table doesn’t have the `id` column**

1. Update planner.
- For the case when we automatically batch insert into KB from select, add two attributes to plan
  - ‘Failback_plan’ - when we convert plan to partitioned, save original plan to this column
  - ‘Probe_query’ - `select <track_column> from <source_table> limit 1`
2. Update executor:
- If plan has a `Probe_query`:
  - Run this query first (will be planned inside)
- If probe query fails:
  - Run ‘Failback_plan’
Else: run plan itself

Fixes https://linear.app/mindsdb/issue/FQE-1973/make-kb-insert-batch-by-default

## Type of change

## Verification Process

## Additional Media:

## Checklist:

### DIFF
--- a/mindsdb/integrations/handlers/pgvector_handler/pgvector_handler.py
+++ b/mindsdb/integrations/handlers/pgvector_handler/pgvector_handler.py
@@ -411,8 +411,8 @@ class PgVectorHandler(PostgresHandler, VectorStoreHandler, KeywordSearchBase):
 
             # PostgreSQL has a 63-character limit for identifiers.
             # Using full UUIDs (36 chars each) would exceed this limit.
-            # We use a hash of company_id + user_id to create a shorter, unique prefix.
-            namespace_hash = hashlib.md5(f"{company_id}_{user_id}".encode()).hexdigest()[:16]
+            # We use a SHA-256 hash of company_id + user_id to create a shorter, unique prefix.
+            namespace_hash = hashlib.sha256(f"{company_id}_{user_id}".encode()).hexdigest()[:16]
             new_table_name = f"t_{namespace_hash}_{table_name}"
 
             # Backwards compatibility: migrate old tables to the new hashed format


mindsdb/api/executor/sql_query/sql_query.py
```diff
@@ -237,6 +251,16 @@ def execute_query(self):
         except PlanningException as e:
             raise LogicError(e) from e
 
+        # -- a plan with failback --
+        if self.planner.plan.probe_query is not None and self.planner.plan.probe_query is not None:
```
StpMax: wrong condition

ea-rus: fixed

PR #12167, opened 2026-01-29
fix(agents): handle NoneType in skill params during agent deletion

Amogh-2404:
**Description**
This PR fixes a regression where `agents_controller.py` would throw an error during agent deletion commands.

**The Issue**
When running `drop agent`, the system encountered an `AttributeError`. The code previously attempted to access `.get("description")` on `rel.skill.params` without checking if `params` was `None`.

**The Fix**
Added a safe check: `description = rel.skill.params.get("description") if rel.skill.params else None`.

**Verification**
- Validated that `drop agent` commands now execute successfully without crashing.

**Type of Change**
- [x] Bug fix (non-breaking change which fixes an issue)

**Checklist**
- [x] My code follows the style guidelines of this project
- [x] I have performed a self-review of my own code


Amogh-2404:

Amogh-2404:
recheck

StpMax:
In the latest version, agents no longer use skills, so I'm closing the PR

PR #12187, opened 2026-02-02
[BUG] - Query charting fails

setohe0909:
## Description

- The chart agent’s LLM was generating SQL with ORDER BY 'Column Alias' (single-quoted string literal). DuckDB treats that as a literal and rejects it; column names must use double-quoted identifiers, e.g. ORDER BY "Column Alias".

- The agent was generating queries with subqueries (SELECT ... FROM (SELECT ... FROM table)). query_df() only allowed FROM table (single Identifier), so any FROM (subquery) was rejected.

Fixes [FQE-2046](https://linear.app/mindsdb/issue/FQE-2046/query-charting-fails)

## Type of change

## Verification Process

## Checklist:

### DIFF
--- a/mindsdb/api/executor/planner/plan_join.py
+++ b/mindsdb/api/executor/planner/plan_join.py
@@ -111,6 +111,7 @@ class PlanJoinTablesQuery:
         self.query_context = {}
 
         self.partition = None
+        self.has_ambiguous_columns = False
 
     def plan(self, query):
         self.tables_idx = {}
@@ -354,6 +355,8 @@ class PlanJoinTablesQuery:
                     col_parts = list(table_info.aliases[-1])
                     col_parts.append(node.parts[-1])
                     node.parts = col_parts
+                else:
+                    self.has_ambiguous_columns = True
 
         query_traversal(query, _check_identifiers)
 
@@ -509,6 +512,9 @@ class PlanJoinTablesQuery:
         # If this table is part of a join with a predictor: cannot prune
         # Predictors may need all columns from joined tables as input features
         if hasattr(self, "join_sequence") and self._join_has_predictor(self.join_sequence):
+            return False
+
+        if self.has_ambiguous_columns:
             return False
 
         # For subselects: can only prune if they have pure SELECT * (no other columns)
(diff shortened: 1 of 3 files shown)


mindsdb/interfaces/agents/modes/prompts.py
```diff
@@ -76,6 +76,7 @@
 - ALWAYS: When writing queries that involve time, use the time functions in MindsDB SQL, or duckdb functions.
 - ALWAYS:Include the name of the schema/database in query, for example, instead of `SELECT * FROM movies WHERE ...` write `SELECT * FROM somedb.movies WHERE..`;
 - ALWAYS: When columns contain spaces, special characters or are reserved words, use double quotes `"` to quote the column name, for example, "column name" instead of [column name].
+- ALWAYS: In ORDER BY clauses, reference column names or aliases with double quotes when they contain spaces or special characters (e.g. ORDER BY "Number of Reviews" DESC). Never use single quotes in ORDER BY—single quotes denote string literals and will cause a DuckDB error (order_by_non_integer_literal).
```
ea-rus: mindsdb uses mysql-like qouting, it should be backticks instead of double-quotes

mindsdb/api/executor/utilities/sql.py
```diff
@@ -157,8 +157,21 @@ def get_duckdb_functions_and_kw_list() -> list[str] | None:
     return _duckdb_functions_and_kw_list
 
 
+def _replace_table_refs_with_df(query_ast: ASTNode) -> None:
```
ea-rus: I would move all updates in sql.py to chart_agent.py: 
Don't make this replacement in the equery as general. Solve problem in the place when it appears: if the wrong query is generaged in the chart_agent.py it has to be fixed there

mindsdb/interfaces/agents/modes/prompts.py
setohe0909: Done

mindsdb/api/executor/utilities/sql.py
setohe0909: Done

PR #12194, opened 2026-02-04
Fix file upload vulnarability

StpMax:
## Description

Fixes #FQE-2068

## Type of change

## Verification Process

## Additional Media:

## Checklist:

### DIFF
--- a/mindsdb/api/http/namespaces/file.py
+++ b/mindsdb/api/http/namespaces/file.py
@@ -73,7 +73,8 @@ class File(Resource):
             file_name = file.file_name.decode()
             data["file"] = file_name
             if Path(file_name).name != file_name:
-                file_object.close()
+                if file_object is not None and not file_object.closed:
+                    file_object.close()
                 raise ValueError(f"Wrong file name: {file_name}")
             file_object = file.file_object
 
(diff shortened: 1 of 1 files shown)


mindsdb/api/http/namespaces/file.py
```diff
@@ -73,6 +73,7 @@ def on_file(file):
             file_name = file.file_name.decode()
             data["file"] = file_name
             if Path(file_name).name != file_name:
+                file_object.close()
```
ZoranPandovski: Do we need to check `if file_object is not None` here since it may be if filename is invalid

mindsdb/api/http/namespaces/file.py
```diff
@@ -100,17 +101,30 @@ def on_file(file):
...
                     file_object.close()
-                Path(file_object.name).rename(Path(file_object.name).parent / data["file"])
+                Path(file_object.name).rename(Path(file_object.name).parent / Path(data["file"]).name)
                 file_object = None
+            allowed_keys = {"source_type", "file", "original_file_name"}
         else:
             data = request.json
+            allowed_keys = {"source_type", "source", "original_file_name"}
+
+        if len(set(data.keys()) - allowed_keys) > 0:
```
ZoranPandovski, line 129: We may check if data is not None before getting keys?

mindsdb/api/http/namespaces/file.py
StpMax: `file_object` is opened by multipart, so it never should be None in this place. But added as additional check

PR #12209, opened 2026-02-09
Prevent memory overflow when duckdb is used

ea-rus:
## Description

If dataset generated by duckdb is too big to fit into memory - it might cause mindsdb server be killed by OS. 
Fixes:
- to fetch dataset from duckdb in batches and raise exception if forecast required RAM is exceed available RAM
- when query is executed, limit used memory to 80% from free ram (was 80% from total ram)

Also fixed error:  `cannot access free variable 'query_str' where it is not associated with a value in enclosing scope` if unknown duckdb function is used

Fixes https://linear.app/mindsdb/issue/FQE-2079/agents-agent-is-running-over-200s-and-error-shows-up-in-the-console
Fixes https://linear.app/mindsdb/issue/FQE-2075/agents-agent-streaming-failed-problem-with-final-query-cannot-access

## Type of change

## Verification Process

## Additional Media:

## Checklist:

### DIFF
--- a/mindsdb/__about__.py
+++ b/mindsdb/__about__.py
@@ -1,6 +1,6 @@
 __title__ = "MindsDB"
 __package_name__ = "mindsdb"
-__version__ = "25.14.1"
+__version__ = "26.0.0"
 __description__ = "MindsDB's AI SQL Server enables developers to build AI tools that need access to real-time data to perform their tasks"
 __email__ = "jorge@mindsdb.com"
 __author__ = "MindsDB Inc"
--- a/mindsdb/api/executor/planner/query_planner.py
+++ b/mindsdb/api/executor/planner/query_planner.py
@@ -476,8 +476,7 @@ class QueryPlanner:
 
         prev_step = self.plan_integration_select(query2)
 
-        # clear limit and where
-        query.limit = None
+        # clear where
         query.where = None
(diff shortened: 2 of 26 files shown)


ea-rus:
Yes the execution part is not covered, because memory_limit is automatically adjusted to 80% of RAM. 
I added the update in the latest commit to use 80% from free RAM.

On other side the memory_limit doesn't limit RAM usage when data is fetched. So it is kept as an addition check
> The original issue pointed to the error "Unable to allocate XXXMiB..." This error appears within DuckDB itself, chunk fetching cannot prevent it.


Agree, let me change comment to  `reserve twice memory than object is used because it might be required in subsequent usage of result (in next steps of planner)`
> The rule that pd.concat required up to twice ram, +1GB should be free after it, is not entirely correct.


looks more complex than pd.concat. did you test it:
- how less RAM it uses compared to pd.concat? 
- is it faster/slower?
- is it safe, can it bring new errors,  for example related to dtype casting? 
> there are more memory effective ways to join

StpMax:
@ea-rus when joining 10 x 300MB pieces build df from array vs pd.concat:
 - mem usage 3554MB vs 6933MB
 - avg time: 2.08s vs 0.72s
 - general types seems to be the same (int/float/datetime/object), but may be need to test more

PR #12224, opened 2026-02-13
Fix NULL handling in WHERE clause and INSERT operations

### LINKED ISSUE #12137: [Bug]: `NULL` handling in WHERE clause differs from standard SQL behavior

MindsDB handles NULL values differently than standard SQL when working with connected PostgreSQL databases. 
Two issues identified:

1. **`WHERE NOT val IS NULL`** returns inverted results compared to **`WHERE val IS NOT NULL`** (they should be equivalent)

2. **`INSERT INTO` with NULL values** behaves differently depending on where it's executed:
   - INSERT executed via **MindsDB** → NULL values are stored as empty strings (`''`)
   - INSERT executed via **PostgreSQL directly** → NULL values are stored correctly as NULL

### Video or screenshots

TABLE VALUES:
Image

GOOD RESULT:
Image

BUG:
Image

SAME QUERY IN POSTGRES:

Image

### Setup

```sql
CREATE TABLE test_db.public.test_nulls (
    id SERIAL PRIMARY KEY,
    val VARCHAR(10)
);
```

### Issue 1: `NOT val IS NULL` vs `val IS NOT NULL`

When data is inserted via PostgreSQL directly:
```sql
-- Run in PostgreSQL
INSERT INTO test_nulls (val) VALUES (NULL), ('A'), (NULL), ('B');
```

| Query | MindsDB Result | PostgreSQL Result | Match? |
|-------|----------------|-------------------|--------|
| `WHERE val IS NOT NULL` | id: 2, 4 | id: 2, 4 | ✅ |
| `WHERE NOT val IS NULL` | id: 1, 3 (NULL rows) | id: 2, 4 | ❌ |

### Issue 2: INSERT NULL via MindsDB stores empty strings

When the same INSERT is executed via MindsDB:
```sql
-- Run in MindsDB
INSERT INTO test_db.public.test_nulls (val) VALUES (NULL), ('A'), (NULL), ('B');
```

| Query | MindsDB Result | PostgreSQL Result | Expected |
|-------|----------------|-------------------|----------|
| `WHERE val IS NULL` | *(no rows)* | *(no rows)* | id: 1, 3 |
| `WHERE val = ''` | id: 1, 3 | id: 1, 3 | *(no rows)* |

This confirms that MindsDB converts NULL to empty string (`''`) during INSERT.

### Expected behavior

1. `WHERE NOT val IS NULL` and `WHERE val IS NOT NULL` should return identical results
2. `INSERT INTO ... VALUES (NULL)` should store actual NULL values, not empty strings

### Step 1: Connect a PostgreSQL database

```sql
CREATE DATABASE test_db
WITH ENGINE = 'postgres',
PARAMETERS = {
    "host": "your_host",
    "port": 5432,
    "database": "your_database",
    "user": "your_user",
    "password": "your_password"
};
```

### Step 2: Create test table

```sql
CREATE TABLE test_db.public.test_nulls (
    id SERIAL PRIMARY KEY,
    val VARCHAR(10)
);
```

### Step 3: Reproduce Issue 1 (NOT operator)

Insert data directly in PostgreSQL, then query via MindsDB:
```sql
-- In PostgreSQL
INSERT INTO test_nulls (val) VALUES (NULL), ('A'), (NULL), ('B');

-- In MindsDB - these two should return the same results but don't
SELECT * FROM test_db.public.test_nulls WHERE val IS NOT NULL ORDER BY id;
SELECT * FROM test_db.public.test_nulls WHERE NOT val IS NULL ORDER BY id;
```

### Step 4: Reproduce Issue 2 (INSERT NULL)

Clear table and insert via MindsDB:
```sql
-- In MindsDB
DELETE FROM test_db.public.test_nulls;
INSERT INTO test_db.public.test_nulls (val) VALUES (NULL), ('A'), (NULL), ('B');

-- Check in PostgreSQL directly
SELECT * FROM test_nulls WHERE val IS NULL;      -- Returns no rows (wrong)
SELECT * FROM test_nulls WHERE val = '';         -- Returns id 1, 3 (NULLs became '')
```

### Summary

| Issue | Description |
|-------|-------------|
| NOT operator | `WHERE NOT val IS NULL` returns inverted results vs `WHERE val IS NOT NULL` |
| INSERT NULL | NULL values inserted via MindsDB are converted to empty strings |

### Workaround
- Use `val IS NOT NULL` instead of `NOT val IS NULL`
- Insert data directly via PostgreSQL if NULL values are needed

### Environment
- MindsDB: Self-hosted Docker
- Source database: PostgreSQL

Mr-Neutr0n:
## Summary

Fixes #12137

Two SQL correctness issues where MindsDB deviated from standard SQL NULL semantics:

**1. `WHERE NOT val IS NULL` returning wrong results**

`NOT val IS NULL` and `val IS NOT NULL` are equivalent in standard SQL but returned opposite results in MindsDB. The root cause was in the SQLAlchemy renderer (`sqlalchemy_render.py`): NULL constants were rendered using `sa.literal(None).label("NULL")`, but SQLAlchemy's `__invert__()` method silently fails on labeled bind parameters. The NOT operator was completely swallowed, so `NOT (val IS NULL)` was rendered as just `val IS NULL`.

Fixed by using `sa.null()` for None-valued constants. `sa.null()` produces a proper SQL NULL element that correctly participates in IS/IS NOT inversion via `__invert__()`.

**2. `INSERT INTO ... VALUES (NULL)` storing empty strings**

NULL values inserted via MindsDB were stored as empty strings (`''`) in PostgreSQL. The cause was in `postgres_handler.py`: the insert method used `COPY ... FROM STDIN WITH CSV` with `df.to_csv()`, but pandas serializes None/NaN as empty quoted strings (`""`) in CSV output. PostgreSQL's COPY CSV interprets `""` as an empty string, not NULL.

Fixed by switching from CSV mode to text mode and using psycopg's `write_row()` method, which correctly transmits `None` values as SQL NULLs through the binary protocol.

## Test plan

- [x] All 13 existing `tests/unit/render/test_sqlalchemyrender.py` tests pass
- [x] Verified `NOT val IS NULL` renders as `val IS NOT NULL` (previously rendered as `val IS NULL`)
- [x] Verified `val IS NOT NULL` still renders correctly
- [x] Verified `val IS NULL` still renders correctly
- [x] Verified NULL in SELECT, CASE, COALESCE, INSERT, and comparison contexts all render correctly
- [x] Verified `write_row()` with `None` produces proper SQL NULLs (not empty strings)
- [ ] Integration test: INSERT NULL via MindsDB into PostgreSQL, verify `IS NULL` returns inserted rows
- [ ] Integration test: `WHERE NOT val IS NULL` and `WHERE val IS NOT NULL` return identical results

### DIFF
--- a/mindsdb/integrations/handlers/postgres_handler/postgres_handler.py
+++ b/mindsdb/integrations/handlers/postgres_handler/postgres_handler.py
@@ -383,10 +383,22 @@ class PostgresHandler(MetaDatabaseHandler):
         columns = [f'"{c}"' for c in columns]
         rowcount = None
 
+        def _normalize_value(v):
+            """Convert pandas NA/NaN/NaT to None for proper SQL NULL handling."""
+            if v is None:
+                return None
+            try:
+                if pd.isna(v) is True:
+                    return None
+            except (ValueError, TypeError):
+                pass
+            return v
+
         with connection.cursor() as cur:
             try:
-                with cur.copy(f'copy "{table_name}" ({",".join(columns)}) from STDIN WITH CSV') as copy:
-                    df.to_csv(copy, index=False, header=False)
+                with cur.copy(f'copy "{table_name}" ({",".join(columns)}) from STDIN') as copy:
+                    for _, row in df.iterrows():
+                        copy.write_row([_normalize_value(v) for v in row])
 
                 connection.commit()
             except Exception as e:
--- a/mindsdb/utilities/render/sqlalchemy_render.py
+++ b/mindsdb/utilities/render/sqlalchemy_render.py
@@ -187,15 +187,18 @@ class SqlalchemyRender:
         elif isinstance(t, ast.Last):
             col = self.to_column(ast.Identifier(parts=["last"]))
         elif isinstance(t, ast.Constant):
-            col = sa.literal(t.value)
-            if t.alias:
-                alias = self.get_alias(t.alias)
+            if t.value is None:
+                col = sa.null()
+                if t.alias:
+                    alias = self.get_alias(t.alias)
+                    col = col.label(alias)
             else:
-                if t.value is None:
-                    alias = "NULL"
+                col = sa.literal(t.value)
+                if t.alias:
+                    alias = self.get_alias(t.alias)
                 else:
                     alias = str(t.value)
-            col = col.label(alias)
+                col = col.label(alias)
         elif isinstance(t, ast.Identifier):
             # sql functions
             col = None


ea-rus:
@Mr-Neutr0n, please accept CLA

martyna-mindsdb:
Closing this PR due to inactivity.

PR #12246, opened 2026-02-25
FAISS: Support several vector tables

ea-rus:
### Added support of several tables for faiss (`duckdb_faiss`) handler

Tables are divided in file system: every table is located in its own folder. 

This folder contains

Files:
- `duckdb.db` - duckdb database to store metadata
  - also can temporary contain `duckdb.db.wal` file (create by duckdb)
- `faiss_index` - file for faiss index
  -  ivf file index contain also `faiss_index_merged` file
- also can contain other temporal `faiss_index*` during creation of index

Folder:
- `dump` - temporal folder dumped index vectors  (during creation of index)
- `recover` - temporal folder for for backup of index before creation of index. If something goes wrong index will be restored from this folder

**Important. Table name is not back compatible!**
Previously faiss handler ignored 'table_name' parameter and used 'data' as folder name. It means that after update all existing faiss db will have only `data` table (instead of name that was used in `create knowledge base` command)

### Changes in index locks

Was: 
When duckdb_faiss handler is used - the faiss index file is always locked inside. Lock is released when handler instance is removed from handler cache (after 1 minute)

Will be: 
`ivf_file` type index can be used from different processes (it applies changes in file immediately). Because of it all types of indexes are locked except of ivf_file. 
Locks means: if the same file is tried to be used from different thread / process - it will rise an exception

Also added a cache for FAISS tables. It is used for the heavy indexes, when the time to load/store index is notable.
Not all tables are cached, only if it has lock inside and contain more than 100K records. 

For the tables that are cached added a workaround to be able to insert into big faiss tables in background:
- before the query is send into background
  - look for KBs as table to insert
  - check if there is faiss vector db table 
     - unload faiss database from cache 
- after executing query in background 
  - do the same (unload faiss database from cache)

Fixes https://linear.app/mindsdb/issue/FQE-1994/faiss-support-several-vector-tables
Fixes https://linear.app/mindsdb/issue/FQE-2124/faiss-error-when-batching-is-used

## Type of change

## Verification Process

## Additional Media:

## Checklist:

### DIFF
--- a/mindsdb/integrations/handlers/duckdb_faiss_handler/faiss_index.py
+++ b/mindsdb/integrations/handlers/duckdb_faiss_handler/faiss_index.py
@@ -4,7 +4,10 @@ import numpy as np
 import psutil
 from pathlib import Path
 
-import portalocker
+try:
+    import fcntl
+except ImportError:
+    fcntl = None
 
 import faiss  # faiss or faiss-gpu
 
@@ -121,11 +124,11 @@ class FaissIndex:
     def _lock_index(self):
         if not self.lock_required:
             return
-        if os.name != "nt":
+        if os.name != "nt" and fcntl:
             self.index_fd = open(self.path, "rb")
             try:
-                portalocker.lock(self.index_fd, portalocker.LOCK_EX | portalocker.LOCK_NB)
-            except portalocker.exceptions.AlreadyLocked:
+                fcntl.flock(self.index_fd, fcntl.LOCK_EX | fcntl.LOCK_NB)
+            except OSError:
                 raise ValueError(f"Index is already used: {self.path}")
 
     def _load_index(self):
--- a/mindsdb/integrations/handlers/duckdb_faiss_handler/requirements.txt
+++ b/mindsdb/integrations/handlers/duckdb_faiss_handler/requirements.txt
@@ -1,2 +1,1 @@
 faiss-cpu==1.13.2
-portalocker


mindsdb/integrations/handlers/duckdb_faiss_handler/duckdb_faiss_handler.py
```diff
@@ -61,173 +64,207 @@ def __init__(self, name: str, **kwargs):
...
-        vector_filter = None
-        meta_filters = []
-        if conditions is None:
-            conditions = []
-        for condition in conditions:
-            if condition.column == "embeddings":
-                vector_filter = condition
+        try:
+            if entry:
+                entry.in_use_count += 1
```
StpMax: with very low chance python may switch threads before write `entry.in_use_count`, so better to wrap it `with self.tables_cache_lock:`

mindsdb/api/executor/sql_query/sql_query.py
```diff
@@ -276,6 +278,9 @@ def execute_query(self):
                 )
 
             if self.planner.plan.is_async and ctx.task_id is None:
+                # release KB locks before inserting in background
+                self.release_kb_lock(steps)
```
StpMax, line 282: This is long chain of calls (`release_kb_lock` -> `iterate over identifiers` -> `kb_controller` -> `delete from cache` ) that would be good to simplify. May be do the following: 
- add to handler custom property like `need_to_release`, set it if needed
- add to handlers_cache method `delete_if_need_to_release`
- add here call `self.session.integration_controller.handlers_cache.delete_if_need_to_release()`

mindsdb/integrations/handlers/duckdb_faiss_handler/duckdb_faiss_handler.py
```diff
@@ -61,173 +64,207 @@ def __init__(self, name: str, **kwargs):
...
+        self.tables_cache_lock = threading.Lock()
 
-            logger.info("Connected to DuckDB database")
-            return self.connection
+    def connect(self):
+        """
+        Handler readiness check.
+        Must not open long-lived DuckDB/FAISS resources; tables are opened per operation.
+        """
+        if not Path(self.persist_directory).exists():
```
StpMax: is it possible? in line 70 the dis is created

ea-rus: yes, maybe it is not probable that directory is removed after init. I can remove this condition

mindsdb/api/executor/sql_query/sql_query.py
ea-rus: will it unload all faiss databases even if they are not used in insert?

mindsdb/integrations/handlers/duckdb_faiss_handler/duckdb_faiss_handler.py
ea-rus: do you mean to wrap only this line, right? 
```python
with self.tables_cache_lock:
    entry.in_use_count += 1

yield table
```

ea-rus:
@StpMax added couple of updates https://github.com/mindsdb/mindsdb/compare/607cfec6e0e0ac6f2bdb48e2c60fd21bb832ae1e...f5b8bddebc21b15de0957d7b30b66a850c74fa7e
- [Prevents leaving orphan memory mapped shard files](https://github.com/mindsdb/mindsdb/commit/11b39e8971c1167e9c34739954138de7ae6f6162)
- check required disk space

mindsdb/integrations/handlers/duckdb_faiss_handler/duckdb_faiss_handler.py
ea-rus: removed condition

mindsdb/api/executor/sql_query/sql_query.py
StpMax: it will depend on what you will add to the `delete_if_need_to_release`

mindsdb/integrations/handlers/duckdb_faiss_handler/duckdb_faiss_table.py
```diff
@@ -0,0 +1,360 @@
...
+
+    def close(self) -> None:
+        self.faiss_index.close()
+        self.connection.close()
+
+    def _create_kw_index(self):
+        with self.connection.cursor() as cur:
+            cur.execute("PRAGMA create_fts_index('meta_data', 'id', 'content')")
+
+    def _drop_kw_index(self):
```
StpMax, line 71: should it set `self.is_kw_index_enabled=False` ? the same for `_create_kw_index`

mindsdb/integrations/handlers/duckdb_faiss_handler/duckdb_faiss_handler.py
StpMax: yes
```
if entry:
    with self.tables_cache_lock:
            entry.in_use_count += 1
yield table
```

mindsdb/api/executor/sql_query/sql_query.py
ea-rus: I mean we need to decide which database has to be unloaded. in `release_kb_lock` I look for a database that is used in specific query and unload only it.
if we keep this logic and mark handler `need_to_release` only for a specific query - it will be work almost in the same way (maybe even more complex):
- find KBs in query, find storages, if it is faiss - mark as 'need_to_release' and call 'delete_if_need_to_release'

mindsdb/integrations/handlers/duckdb_faiss_handler/duckdb_faiss_handler.py
ea-rus: added

mindsdb/integrations/handlers/duckdb_faiss_handler/duckdb_faiss_table.py
ea-rus: yes, missed it
added

mindsdb/api/executor/sql_query/sql_query.py
```diff
@@ -340,5 +345,14 @@ def execute_step(self, step, steps_data=None):
 
         return handler(self, steps_data=steps_data).call(step)
 
+    def release_kb_lock(self, steps):
+        # find knowledge bases that used as tables to insert.
+        #  then and release locks of vector for these knowledge bases
```
StpMax: Broken comment

mindsdb/integrations/handlers/duckdb_faiss_handler/faiss_index.py
```diff
@@ -65,6 +119,8 @@ def __init__(self, path: str, config: dict):
             self._load_index()
 
     def _lock_index(self):
+        if not self.lock_required:
+            return
         if os.name != "nt":
```
StpMax: If it is not for windows, then better use lockf/flock to avoid additional dependency

PR #12251, opened 2026-02-27
Fix scheduler worker status for failed jobs

ianu82:
## Summary
- stop reporting scheduler worker completion as success when `execute_task_local` raises
- emit `False` on the worker output queue for failures/interruption and `True` only on successful task execution
- add a regression unit test for `execute_async` failure signaling

## Why
`execute_async` previously pushed success to `q_out` even after exceptions, which masked task execution failures from the scheduler loop.

## Validation
- `python3 -m py_compile mindsdb/interfaces/jobs/scheduler.py tests/unit/interfaces/jobs/test_scheduler.py`
- Full pytest was not run locally because optional test dependencies are missing in this environment (`mindsdb_sql_parser`, `duckdb`).

### DIFF
--- a/mindsdb/interfaces/jobs/scheduler.py
+++ b/mindsdb/interfaces/jobs/scheduler.py
@@ -27,13 +27,15 @@ def execute_async(q_in, q_out):
         try:
             executor.execute_task_local(record_id, history_id)
         except (KeyboardInterrupt, SystemExit):
-            q_out.put(True)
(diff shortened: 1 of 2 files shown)


ianu82:

ianu82:
recheck

StpMax:
`q_out` is just indicator that job is finished, no matter that is send to it

PR #12252, opened 2026-02-27
Fix malformed BigQuery PK metadata SQL

ianu82:
## Summary
- remove the trailing comma from the `meta_get_primary_keys` SELECT list in the BigQuery handler
- extend the unit test to assert the generated query does not include a dangling comma before `FROM`

## Why
The trailing comma produces invalid SQL and breaks metadata introspection for primary keys.

## Validation
- `python3 -m py_compile mindsdb/integrations/handlers/bigquery_handler/bigquery_handler.py tests/unit/handlers/test_bigquery.py`
- `pytest -q tests/unit/handlers/test_bigquery.py -k meta_get_primary_keys_filters` *(fails in this environment due to missing optional dependency `google.api_core`)*


ianu82:
recheck

PR #12253, opened 2026-02-27
Fix Postgres column stats UnboundLocalError on errors

ianu82:
## Summary
- return early from `meta_get_column_statistics` when `native_query` does not return a table response
- only run dataframe post-processing when a table dataframe is present
- add a regression unit test that verifies non-table responses are returned unchanged

## Why
The previous implementation always referenced `df` after the conditional transformation block, causing an `UnboundLocalError` when `native_query` returned an error/non-table response.

## Validation
- `python3 -m py_compile mindsdb/integrations/handlers/postgres_handler/postgres_handler.py tests/unit/handlers/test_postgres.py`
- `pytest -q tests/unit/handlers/test_postgres.py -k meta_get_column_statistics_returns_non_table_response` *(fails in this environment due to missing optional dependency `psycopg`)*

### DIFF
--- a/mindsdb/integrations/handlers/postgres_handler/postgres_handler.py
+++ b/mindsdb/integrations/handlers/postgres_handler/postgres_handler.py
@@ -687,31 +687,33 @@ class PostgresHandler(MetaDatabaseHandler):
 
         result = self.native_query(query)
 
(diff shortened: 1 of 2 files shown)


ianu82:
recheck

PR #12254, opened 2026-02-27
Fix wrong JSON exception type in Postgres subscribe path

ianu82:
## Summary
- replace `json.JSONDecoder` with `json.JSONDecodeError` in Postgres notification payload parsing

## Why
`json.JSONDecoder` is a decoder class, not an exception type. Catching it does not intercept malformed JSON payloads and can surface unintended errors during subscription callbacks.

## Validation
- `python3 -m py_compile mindsdb/integrations/handlers/postgres_handler/postgres_handler.py`


ianu82:
recheck

PR #12256, opened 2026-02-27
Fix SQLQuery on_error step number handling

ianu82:
## Summary
- initialize a dedicated `failed_step_num` before step execution begins
- update it per-step and use it in `run_query.on_error(...)`
- avoid dereferencing `step.step_num` in exception paths where `step` may never have been assigned

## Why
When failures occur before entering the loop body (for example during pre-step setup), the old error path could raise `UnboundLocalError` while trying to report the original failure, masking root cause diagnostics.

## Validation
- `python3 -m py_compile mindsdb/api/executor/sql_query/sql_query.py`

### DIFF
--- a/mindsdb/api/executor/sql_query/sql_query.py
+++ b/mindsdb/api/executor/sql_query/sql_query.py
@@ -289,6 +289,7 @@ class SQLQuery:
             ctx.run_query_id = self.run_query.record.id
 
         step_result = None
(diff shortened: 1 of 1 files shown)


ianu82:
recheck

StpMax:
Adding additional variable for this check is redundant. The check can be done in one line. I'll prepare new PR for this.

PR #12257, opened 2026-02-27
Fix Databricks empty SELECT response type

ianu82:
## Summary
- update Databricks `native_query` to return `RESPONSE_TYPE.TABLE` whenever result-set metadata (`cursor.description`) exists, even when the rowset is empty
- keep `RESPONSE_TYPE.OK` for statements that do not return a result set
- add a regression test for empty SELECT behavior

## Why
Empty SELECT queries were incorrectly classified as `OK`, which hides schema information and causes downstream metadata/table handling inconsistencies.

## Validation
- `python3 -m py_compile mindsdb/integrations/handlers/databricks_handler/databricks_handler.py tests/unit/handlers/test_databricks.py`
- `pytest -q tests/unit/handlers/test_databricks.py -k empty_select_returns_table` *(fails in this environment due to missing optional dependency `appdirs`)*

### DIFF
--- a/mindsdb/integrations/handlers/databricks_handler/databricks_handler.py
+++ b/mindsdb/integrations/handlers/databricks_handler/databricks_handler.py
@@ -404,7 +404,7 @@ class DatabricksHandler(MetaDatabaseHandler):
             try:
                 cursor.execute(query)
                 result = cursor.fetchall()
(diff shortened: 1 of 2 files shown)


ianu82:
recheck

PR #12258, opened 2026-02-27
Fix BigQuery empty SELECT response typing

ianu82:
## Summary
- update BigQuery `native_query` to treat any dataframe result as `RESPONSE_TYPE.TABLE`, including empty rowsets
- add a regression unit test covering empty SELECT behavior

## Why
Empty SELECT queries produce dataframes with schema but no rows. Returning `OK` for these responses drops the table contract and breaks metadata/column-aware downstream logic.

## Validation
- `python3 -m py_compile mindsdb/integrations/handlers/bigquery_handler/bigquery_handler.py tests/unit/handlers/test_bigquery.py`
- `pytest -q tests/unit/handlers/test_bigquery.py -k empty_select_returns_table` *(fails in this environment due to missing optional dependency `google.api_core`)*

### DIFF
--- a/mindsdb/integrations/handlers/bigquery_handler/bigquery_handler.py
+++ b/mindsdb/integrations/handlers/bigquery_handler/bigquery_handler.py
@@ -139,7 +139,8 @@ class BigQueryHandler(MetaDatabaseHandler):
             )
             query = connection.query(query, job_config=job_config)
             result = query.to_dataframe()
-            if isinstance(result, pd.DataFrame):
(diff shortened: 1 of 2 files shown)


ianu82:
Addressed. I refined `native_query` result typing so non-result operations still return `RESPONSE_TYPE.OK`, while empty `SELECT` queries with schema still return `RESPONSE_TYPE.TABLE`.

Implemented condition:
- `isinstance(result, pd.DataFrame) and (not result.empty or len(result.columns) > 0)`

Also added a regression test for the no-schema empty dataframe case (`pd.DataFrame()` -> `OK`) alongside the existing empty-select test (columns present -> `TABLE`).

PR #12259, opened 2026-02-27
Fix Databricks get_tables non-table response handling

ianu82:
## Summary
- add a guard in Databricks `get_tables()` to return early for non-table responses or missing dataframes
- only uppercase/rename columns when a table dataframe is present
- add a regression unit test ensuring error responses pass through unchanged

## Why
`get_tables()` previously assumed `result.data_frame` always existed. On error/non-table responses this triggered secondary exceptions and masked the original failure.

## Validation
- `python3 -m py_compile mindsdb/integrations/handlers/databricks_handler/databricks_handler.py tests/unit/handlers/test_databricks.py`
- `pytest -q tests/unit/handlers/test_databricks.py -k get_tables_returns_non_table_response_without_transform` *(fails in this environment due to missing optional dependency `appdirs`)*

### DIFF
--- a/mindsdb/integrations/handlers/databricks_handler/databricks_handler.py
+++ b/mindsdb/integrations/handlers/databricks_handler/databricks_handler.py
@@ -465,6 +465,8 @@ class DatabricksHandler(MetaDatabaseHandler):
                 {all_filter}
         """
         result = self.native_query(query)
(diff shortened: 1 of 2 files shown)


ianu82:
review

PR #12260, opened 2026-02-27
Fix MySQL get_columns schema scoping

ianu82:
## Summary
- add `table_schema = DATABASE()` filter to MySQL `get_columns` query
- update MySQL unit test expectations to include schema scoping

## Why
Without schema filtering, `get_columns` can return metadata for same-named tables from other databases, causing inconsistent or incorrect column introspection.

## Validation
- `python3 -m py_compile mindsdb/integrations/handlers/mysql_handler/mysql_handler.py tests/unit/handlers/test_mysql.py`


ianu82:
recheck

StpMax:
there is issue in tests, i'll fix it after merge

PR #12261, opened 2026-02-27
Fix MSSQL check_connection exception handling

ianu82:
## Summary
- broaden `SqlServerHandler.check_connection()` exception handling from `OperationalError` to `Exception`
- add a regression assertion covering non-`OperationalError` failures (e.g., invalid configuration)

## Why
`check_connection()` can fail with exceptions other than `OperationalError` (especially in ODBC/validation paths). Previously those escaped and bypassed `StatusResponse` error reporting.

## Validation
- `python3 -m py_compile mindsdb/integrations/handlers/mssql_handler/mssql_handler.py tests/unit/handlers/test_mssql.py`
- `pytest -q tests/unit/handlers/test_mssql.py -k test_check_connection` *(fails in this environment due to missing optional dependency `appdirs`)*


ianu82:
recheck

PR #12262, opened 2026-02-27
Fix TaskThread crash on missing task record

ianu82:
## Summary
- add a null-check in `TaskThread.run()` for missing task records
- log the missing record and return early instead of dereferencing `None`
- add a unit test for the early-return path

## Why
If a task is deleted or not found when a thread starts, `task_record` is `None` and the previous code raises an `AttributeError` while setting context fields.

## Validation
- `python3 -m py_compile mindsdb/interfaces/tasks/task_thread.py tests/unit/interfaces/tasks/test_task_thread.py`
- `pytest -q tests/unit/interfaces/tasks/test_task_thread.py` *(fails in this environment due to missing optional dependency `mind_castle`)*


ianu82:
recheck

PR #12263, opened 2026-02-27
FQE fixes for agent 

ea-rus:
Fixes:

1. Fixes for "Optimizations of combining view with the outer query" #11437

- find the proper column name (with table alias) and replace name in injected conditions
- don't inject if column isn't found and query has a star
- copy alias from view to subselect (column of the query might relate to it)

Fixed query example:
```sql
CREATE VIEW enterprise_sales AS (
  SELECT *  FROM postgres_demo.websales_sales AS sales
   JOIN mongodb_demo.customers AS customers ON sales.customer_id = customers.customer_id
  WHERE   customers.segment = "Enterprise"  AND sales.sales > 1000
);

SELECT DISTINCT enterprise_sales.customer_id
FROM enterprise_sales AS enterprise_sales
WHERE enterprise_sales.segment = 'Enterprise';
```

2. Fixes for "Optimize inner join" #11904

- Added timeout (5 seconds), if it passed: don't try to fetch sample of data and fetch all records 
- When query has CTE, it is converted to wrong optimized query.
Fixed query example:
```sql
WITH best_selling_product AS (
    SELECT ...    FROM mindsdb.enterprise_sales AS enterprise_sales
    ...   LIMIT 1
),
enterprise_customers AS (
    SELECT DISTINCT enterprise_sales.customer_id
    FROM mindsdb.enterprise_sales
    ...   
)
SELECT   reviews_kb.chunk_content AS answer
 FROM mindsdb.reviews_kb AS reviews_kb
 JOIN best_selling_product    ON reviews_kb.product_id = best_selling_product.product_id
 WHERE reviews_kb.customer_id IN (SELECT customer_id FROM enterprise_customers)
LIMIT 10000;
```

3. Error  "cannot access local variable 'column_quoted' where it is not associated with a value"

Fixes https://linear.app/mindsdb/issue/FQE-2142/agents-error-when-querying

## Type of change

## Verification Process

## Additional Media:

## Checklist:

### DIFF
--- a/mindsdb/api/executor/sql_query/steps/fetch_dataframe_partition.py
+++ b/mindsdb/api/executor/sql_query/steps/fetch_dataframe_partition.py
@@ -129,7 +129,7 @@ class FetchDataframePartitionCall(BaseStepCall):
                 break
 
             # break if process is too long or to many tries
-            if try_num > 3 or started_at - time.time() > 5:
+            if try_num > 3 or time.time() - started_at > 5:
                 # the last try without the limit
                 first_table_limit = None
                 continue
--- a/mindsdb/interfaces/database/projects.py
+++ b/mindsdb/interfaces/database/projects.py
@@ -203,17 +203,17 @@ class Project:
 
                     if isinstance(arg1, Identifier):
                         name = arg1.parts[-1].lower()
-                        if name in white_list:
-                            arg1 = white_list[name]
                         # don't move condition for join with Star
-                        elif name in black_list or not (has_star and not is_join):
-                            continue
-                    if isinstance(arg2, Identifier):
-                        name = arg2.parts[-1].lower()
-                        if name in white_list:
-                            arg2 = white_list[name]
                         if name in black_list or not (has_star and not is_join):
                             continue
+                        elif name in white_list:
+                            arg1 = white_list[name]
+                    if isinstance(arg2, Identifier):
+                        name = arg2.parts[-1].lower()
+                        if name in black_list or not (has_star and not is_join):
+                            continue
+                        elif name in white_list:
+                            arg2 = white_list[name]
 
                     # condition can be moved into view
                     condition2 = BinaryOperation(condition.op, [arg1, arg2])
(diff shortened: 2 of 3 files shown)


mindsdb/api/executor/sql_query/steps/fetch_dataframe_partition.py
```diff
@@ -126,7 +128,8 @@ def repeat_till_reach_limit(self, step, limit):
                 result = result[:limit]
                 break
 
-            if try_num > 3:
+            # break if process is too long or to many tries
+            if try_num > 3 or started_at - time.time() > 5:
```
StpMax: it should be `time.time() - started_at`

mindsdb/interfaces/database/projects.py
```diff
@@ -185,28 +185,34 @@ def get_conditions_to_move(node):
...
+                            arg1 = white_list[name]
+                        # don't move condition for join with Star
+                        elif name in black_list or not (has_star and not is_join):
                             continue
                     if isinstance(arg2, Identifier):
                         name = arg2.parts[-1].lower()
-                        if name in black_list or not (has_star or name in white_list):
+                        if name in white_list:
+                            arg2 = white_list[name]
+                        if name in black_list or not (has_star and not is_join):
```
StpMax, line 213: Should there be `elif`?

mindsdb/interfaces/query_context/last_query.py
```diff
@@ -179,29 +189,46 @@ def index_query(node, is_table, parent_query, **kwargs):
...
+                query.group_by is None
+                and query.order_by is None
+                and query.having is None
+                and query.distinct is False
+                and query.where is None
+                and query.limit is None
+                and query.offset is None
+                and query.cte is None
+            ):
+                query = query.from_table
```
StpMax: in this case will be better to do deepcopy
`to_string` should not modify object silently

ea-rus: thanks, fixed

mindsdb/interfaces/database/projects.py
ea-rus: yes, but black list should should be greater priority 
I moved it up

mindsdb/api/executor/sql_query/steps/fetch_dataframe_partition.py
ea-rus: thanks, fixed

StpMax (approved):

PR #12279, opened 2026-03-08
fix: correct isinstance check for redirect_url in database endpoints

jnMetaCode:
## Summary

Fixes a bug in  where  is used instead of .

 is always a  instance, so  always evaluates to . This makes the redirect URL handling completely unreachable in both the  and  endpoints.

The fix changes  to  in both locations, which matches the original intent of checking whether the  attribute contains a string value before returning the redirect response.

## Test plan
- Verify that  with a non-None  now correctly triggers the redirect response path
- Existing tests should continue to pass since this code path was previously unreachable


StpMax:
@jnMetaCode please, leave a comment as described above

jnMetaCode:

PR #12283, opened 2026-03-11
FQE-2165 - Add an endpoint for extracting default vector stores

setohe0909:
## Type of change

## Checklist:

### DIFF
--- a/mindsdb/integrations/handlers/pgvector_handler/pgvector_handler.py
+++ b/mindsdb/integrations/handlers/pgvector_handler/pgvector_handler.py
@@ -134,7 +134,7 @@ class PgVectorHandler(PostgresHandler, VectorStoreHandler, KeywordSearchBase):
             return self.native_query(query_str, params, no_restrict=True, stream=False)
         return super().query(query)
 
-    def native_query(self, query, params=None, no_restrict=False, stream: bool = True, **kwargs) -> Response:
+    def native_query(self, query, params=None, no_restrict=False) -> Response:
         """
         Altered `native_query` method of postgres handler.
         Restrict usage of native query from executor with shared pg vector connection
@@ -143,7 +143,7 @@ class PgVectorHandler(PostgresHandler, VectorStoreHandler, KeywordSearchBase):
         # Prevent execute native queries
         if self._is_shared_db and not no_restrict:
             return Response(RESPONSE_TYPE.OK)
-        return super().native_query(query, params=params, stream=stream, **kwargs)
+        return super().native_query(query, params=params)
 
     def raw_query(self, query, params=None) -> Response:
         resp = super().native_query(query, params, stream=False)


mindsdb/interfaces/knowledge_base/controller.py
```diff
@@ -1467,14 +1467,40 @@ def _test_reranking(self, params):
...
+            )
             return vector_store_name
 
-        self.session.integration_controller.add(vector_store_name, "pgvector", params or {})
+        self.session.integration_controller.add(vector_store_name, "pgvector", requested_params)
         return vector_store_name
 
+    @staticmethod
+    def _validate_pgvector_store_compatibility(vector_store_name: str, existing_params: dict, requested_params: dict):
+        existing_is_sparse = bool(existing_params.get("is_sparse", False))
```
ea-rus: is it required by the scope of this PR? (to compare two parameters from previous and new vector store)
if not, we can make it simpler (keep `_create_persistent_pgvector` as is)

mindsdb/interfaces/knowledge_base/default_storage_resolver.py
```diff
@@ -0,0 +1,62 @@
...
+
+def get_knowledge_base_storage_config(config_obj=None) -> list[str]:
+    config_obj = config_obj or config
+    storage = config_obj.get("knowledge_bases", {}).get("storage", [])
+
+    if storage is None:
+        return []
+    if isinstance(storage, str):
+        storage = [storage]
+    if not isinstance(storage, list):
```
ea-rus: as I understand storage engine is the single value (not the list of storage engines). or not?

mindsdb/interfaces/knowledge_base/controller.py
```diff
@@ -1500,6 +1526,35 @@ def _create_persistent_chroma(self, kb_name, engine="chromadb"):
         self.session.integration_controller.add(vector_store_name, engine, connection_args)
         return vector_store_name
 
+    def _resolve_default_vector_storage(self, kb_name: str, is_sparse: bool = False, vector_size: int = None):
```
ea-rus, line 1447: this function is not used
but I think we can remove it, and made in [this place](https://github.com/mindsdb/mindsdb/blob/0758a11c5a03138ba7dafacc7a3688a01439656c/mindsdb/interfaces/knowledge_base/controller.py#L1273) update like:
```python
vector_engine = get_from_config()
if vector_engine is None:
   if os.environ.get("KB_PGVECTOR_URL"):
      vector_engine = 'pgvector'
   else:
      vector_engine = 'faiss'

if vector_engine = 'pgvector':
     self._create_persistent_pgvector()
elif  vector_engine = 'faiss':
   self._create_persistent_faiss()
else:
    exception 
```

mindsdb/interfaces/knowledge_base/default_storage_resolver.py
setohe0909: I was thinking of a list because in the designs we have an input list to select more than one, and also in this requirement, we have :

```
mindsdb running via docker extension or docker compose --> pgvector, faiss
```

so probably in this kind of env there is more than one option there

mindsdb/interfaces/knowledge_base/default_storage_resolver.py
```diff
@@ -0,0 +1,85 @@
...
+        if len(storage) == 0:
+            return None
+        storage = storage[0]
+
+    if not isinstance(storage, str):
+        raise ValueError("knowledge_bases.storage must be a string value")
+
+    return _normalize_engine_name(storage)
+
+
```
ea-rus, line 46: integration_controller is not used

mindsdb/api/http/namespaces/config.py
```diff
@@ -34,6 +35,10 @@ def get(self):
             if value is not None:
                 resp[key] = value
 
+        knowledge_bases_config = copy.deepcopy(config["knowledge_bases"])
+        knowledge_bases_config.update(resolve_default_storage_engines(ca.integration_controller, config))
```
ea-rus: we can call list of engines directly:
```python
 knowledge_bases_config['engines'] = default_storage_resolver._get_env_available_engines()
```

mindsdb/interfaces/knowledge_base/default_storage_resolver.py
```diff
@@ -0,0 +1,85 @@
...
+def resolve_default_storage_engines(integration_controller, config_obj=None) -> dict[str, Any]:
```
ea-rus: do I understand right the logic that should happens here:

```python
# if set in config - return it
configured_storage = get_knowledge_base_storage_config(config_obj)
if configured_storage is not None:
    return configured_storage

# default value, if nothing set
return 'duckdb_faiss' 
```

is it right? if yes, maybe simplify this function?

mindsdb/interfaces/knowledge_base/default_storage_resolver.py
```diff
@@ -0,0 +1,85 @@
...
+                "source": "config" if configured_storage == engine else "fallback",
+            }
+        )
+
+    return {
+        "storage": configured_storage,
+        "resolved_storage": resolved_storage,
+        "default_storage": default_engine,
+        "available_vector_engines": available_vector_engines,
+        "pgvector_enabled": pgvector_enabled,
```
ea-rus, line 89: I didn't found how these fields are used:
- pgvector_enabled
- available_vector_engines
- resolved_storage
- storage
are they required?

mindsdb/interfaces/knowledge_base/default_storage_resolver.py
```diff
@@ -0,0 +1,85 @@
...
+    normalized = engine.strip().lower()
+    if normalized in ("duckdb_faiss", "faiss"):
+        return "faiss"
+    if normalized == "pgvector":
+        return "pgvector"
+    return normalized or None
+
+
+def _get_env_available_engines() -> list[str]:
+    env_value = os.environ.get("KNOWLEDGE_BASES_STORAGE", "")
```
ea-rus: As I understand it is used to show available in ui, right? 
We already have KB_PGVECTOR_URL env var, why don't use it? 
```python
engines = ['duckdb_faiss']

if  os.environ.get("KB_PGVECTOR_URL"):
   engines.append('pgvector')
return engines
```

mindsdb/api/http/namespaces/config.py
setohe0909: Done

mindsdb/interfaces/knowledge_base/default_storage_resolver.py
setohe0909: Done. We now derive available engines from `KB_PGVECTOR_URL` when `KNOWLEDGE_BASES_STORAGE` is not set (faiss + optional pgvector)

mindsdb/interfaces/knowledge_base/default_storage_resolver.py
setohe0909: Those are required and used in the UI to detect some configs and handle the UX

mindsdb/integrations/handlers/pgvector_handler/pgvector_handler.py
```diff
@@ -134,7 +134,7 @@ def query(self, query: ASTNode) -> Response:
             return self.native_query(query_str, params, no_restrict=True, stream=False)
         return super().query(query)
 
-    def native_query(self, query, params=None, no_restrict=False) -> Response:
+    def native_query(self, query, params=None, no_restrict=False, stream: bool = True, **kwargs) -> Response:
```
ea-rus: update of this file might be not required

docker-compose.yml
```diff
@@ -36,6 +36,7 @@ services:
     environment:
       MINDSDB_DB_CON: "postgresql://postgres:postgres@postgres/mindsdb"
       KB_PGVECTOR_URL: "postgresql://postgres:postgres@postgres/kb"  # Use pgvector as the default store for knowledge bases
+      KNOWLEDGE_BASES_STORAGE: "pgvector,faiss"  # Available engines exposed to API/UI
```
ea-rus: KB_PGVECTOR_URL is a flag for us that pgvector is available. we can remove KNOWLEDGE_BASES_STORAGE

requirements/requirements.txt
```diff
@@ -23,7 +23,7 @@ dateparser==1.2.0
 dill == 0.3.6
 numpy ~= 2.0
 pytz
-botocore
+botocore >= 1.34.0  # need 1.34+ for urllib3 2.x (ssl_ import)
```
ea-rus: is it required?

docker-compose.yml
setohe0909: Done

requirements/requirements.txt
setohe0909: Done

mindsdb/integrations/handlers/pgvector_handler/pgvector_handler.py
setohe0909: Done

ea-rus (approved):

PR #12294, opened 2026-03-14
fix: normalize backtick-wrapped qualified KB names before SQL parsing

Krishnav1237:
The LLM sometimes generates qualified KB table references wrapped in a single set of backticks, e.g. `mindsdb.my_kb`. mindsdb_sql_parser correctly treats this as one identifier with a literal dot, not as a schema.table pair, causing:

  "Can't select from <mindsdb.my_kb> in project: unknown"

The existing sql.strip("`") call only strips backticks from the edges of the full SQL string and does not address identifiers inside the query.

This PR adds normalize_kb_table_ref() which rewrites `schema.table` patterns to their correctly split form before parse_sql() is called. Single identifiers that genuinely need quoting (e.g. `my-kb`) are preserved correctly.

Fixes: #11156

Tested: tests/unit/interfaces/agents/test_sql_toolkit.py

### DIFF
--- a/mindsdb/interfaces/agents/utils/sql_toolkit.py
+++ b/mindsdb/interfaces/agents/utils/sql_toolkit.py
@@ -111,7 +111,7 @@ def normalize_kb_table_ref(sql: str) -> str:
             return f"{prefix}{quoted}"
             
         # Qualified name — split and quote each part only if necessary
-        parts = inner.split('.', 1)
+        parts = inner.split('.')
         quoted = [
             p if re.match(r'^[A-Za-z_][A-Za-z0-9_]*$', p) else f'`{p}`'
             for p in parts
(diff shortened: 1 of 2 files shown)


Krishnav1237:

ea-rus:
The issue https://github.com/mindsdb/mindsdb/issues/11156 was closed in the last July. 
This problem should be handled in [this block](https://github.com/mindsdb/mindsdb/blob/9dd046c5e8e4db3446f877c54b6ad56b91e1ab3a/mindsdb/interfaces/agents/utils/sql_toolkit.py#L253):
When table has dots in name, it tries to unquote it.
Originally done in https://github.com/mindsdb/mindsdb/pull/11191/changes

Do you have this problem now?

ea-rus:
@Krishnav1237 closing it. You can reopen if you have this problem

PR #12429, opened 2026-05-06
fix(mssql): expose real schemas in information_schema.TABLES and fix preview query (#12242)

### LINKED ISSUE #12242: [Bug]: [MSSQL] information_schema.TABLES does not expose real schemas + wrong preview query in UI

When using a MSSQL Server datasource in MindsDB, there is an inconsistency between:
- the Explorer UI (left panel), which correctly shows schemas and tables
- the SQL layer (information_schema.TABLES), which does not expose real schemas

Additionally, the table preview tooltip in the UI generates an invalid query, because it ignores the schema name.

This causes confusion and breaks SQL queries that rely on schemas.

### Video or screenshots

Image

### Expected behavior

_No response_

Add a MSSQL Server datasource (example: f_prod)

Open the Explorer UI

Schemas like dbo, app, usr, etc. are shown correctly

Run the following query in the MindsDB SQL editor:

`SELECT
  TABLE_SCHEMA,
  TABLE_NAME
FROM INFORMATION_SCHEMA.TABLES
WHERE TABLE_SCHEMA = 'f_prod'
ORDER BY TABLE_SCHEMA, TABLE_NAME;`

### Anything else?

_No response_

Beandon13:
## Description

Fixes #12242

When a MSSQL Server datasource is connected to MindsDB, two related bugs
affected how table schemas were exposed:

**Bug 1 — `INFORMATION_SCHEMA.TABLES` hides real SQL schemas**

MindsDB's `system_tables` layer unconditionally overwrites `TABLE_SCHEMA`
with the datasource name (e.g. `f_prod`).  For single-schema integrations
(MySQL, SQLite) this is fine, but MSSQL databases routinely contain multiple
user schemas (`dbo`, `app`, `usr`, …).  After the overwrite, all real schema
information was lost, so `SELECT TABLE_SCHEMA, TABLE_NAME FROM
INFORMATION_SCHEMA.TABLES` returned `f_prod` in `TABLE_SCHEMA` and a bare
table name instead of the fully-qualified `dbo.MyTable`.

**Bug 2 — UI table preview generates an invalid two-part query**

Because the schema was not embedded in `TABLE_NAME`, the table-preview tooltip
generated `SELECT * FROM f_prod.MyTable` instead of the correct three-part
form `SELECT * FROM f_prod.dbo.MyTable`, which fails at execution time.

## Type of change

## Fix

`get_tables()` in the MSSQL handler now accepts an `all: bool = False`
parameter, matching the pattern already used by the Postgres and Databricks
handlers:

| Caller | `all` | Behaviour |
|--------|-------|-----------|
| `INFORMATION_SCHEMA.TABLES` (system_tables layer) | `False` (default) | `table_name` returned as `<schema>.<table>` so the full three-part reference is preserved after the datasource-level `TABLE_SCHEMA` override |
| Explorer / tree API (`tree.py`) | `True` | Raw `table_schema` + plain `table_name` returned so the UI can group tables under schema nodes |
| `self.schema` configured | either | Schema filter applied; plain unqualified `table_name` returned (original behaviour) |

`get_columns()` was extended with an optional `schema_name` parameter
(consistent with the Postgres handler) and now automatically extracts the
schema from a qualified `<schema>.<table>` name produced by `get_tables()`,
so column lookups remain precise in all modes.

Non-user system schemas (`sys`, `guest`, fixed database role schemas) are
filtered out in both `all=True` and `all=False` modes.

## Verification Process

- [ ] Test Location: `tests/unit/handlers/test_mssql.py`
- [ ] Verification Steps:
  1. Connect a MSSQL datasource with multiple schemas (e.g. `dbo`, `app`).
  2. Run `SELECT TABLE_SCHEMA, TABLE_NAME FROM INFORMATION_SCHEMA.TABLES WHERE TABLE_SCHEMA = '<datasource>'` — `TABLE_NAME` now includes the schema prefix (`dbo.Customers`, `app.Orders`, …).
  3. Run `SELECT * FROM <datasource>.dbo.Customers LIMIT 10` — executes successfully.
  4. Hover a table in the Explorer UI — the preview query reads `SELECT * FROM <datasource>.<schema>.<table> LIMIT 100`.

## Checklist

- [x] My code follows the style guidelines (PEP 8) of MindsDB.
- [x] Relevant unit tests are updated or added (3 new `test_get_tables_*` tests, 2 new `test_get_columns_*` tests).

### DIFF
--- a/mindsdb/integrations/handlers/mssql_handler/mssql_handler.py
+++ b/mindsdb/integrations/handlers/mssql_handler/mssql_handler.py
@@ -435,39 +435,107 @@ class SqlServerHandler(MetaDatabaseHandler):
             resp.error_message += f"\nThe problem with render: {render_error}"
         return resp
 
-    def get_tables(self) -> Response:
+    def get_tables(self, all: bool = False) -> Response:
         """
-        Retrieves a list of all non-system tables and views in the current schema of the Microsoft SQL Server database.
+        Retrieves a list of all non-system tables and views in the Microsoft SQL Server database.
+
+        When a specific schema is configured (``self.schema``), only tables belonging to that
+        schema are returned regardless of the ``all`` flag.  When no schema is configured:
+
+        * ``all=False`` (default, used by ``INFORMATION_SCHEMA.TABLES``) – returns tables from
+          every non-system schema and exposes the real SQL schema name in *table_schema*.
+          ``table_name`` is qualified as ``<schema>.<table>`` so that the three-part name
+          ``<datasource>.<schema>.<table>`` works transparently in MindsDB SQL.
+        * ``all=True`` (used by the Explorer UI / tree endpoint) – same query but
+          ``table_name`` is **not** prefixed; the raw ``table_schema`` column is kept so the
+          UI can group tables under their schema nodes.
+
+        Args:
+            all (bool): When *True* the unqualified ``table_name`` is returned (Explorer mode).
+                        When *False* (default) ``table_name`` is prefixed with the schema so
+                        that ``INFORMATION_SCHEMA.TABLES`` exposes the full three-part table
+                        reference.
 
         Returns:
-            Response: A response object containing the list of tables and views, formatted as per the `Response` class.
+            Response: A response object containing the list of tables and views, formatted as
+                      per the ``Response`` class.
         """
 
+        if self.schema:
+            # Single-schema mode: schema filter is always applied; names are unambiguous.
+            query = f"""
+                SELECT
+                    table_schema,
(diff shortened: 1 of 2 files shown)


Beandon13:
Good catch on the injection risk. Fixed in the latest push: both `table_name` and `effective_schema` are now passed through `str.replace("'", "''")` before interpolation — standard SQL string-literal escaping. These values come from the DB's own `INFORMATION_SCHEMA` or from the handler's configured `self.schema`, but agreed the defensive escaping is the right call regardless of the source.

PR #12440, opened 2026-05-17
Avoid mutating LLM prompt dataframes

pragnyanramtha:
## Summary
- build LLM prompt template completions without writing an internal `__mdb_prompt` column into the caller DataFrame
- return a typed empty integer array for no-placeholder non-strict prompt templates
- add a regression test that preserves a user-provided `__mdb_prompt` column

## Why
`get_completed_prompts` is used by LLM/model handlers to render prompt templates. The helper previously mutated its input DataFrame by adding `__mdb_prompt`, which could leak internal state into downstream code or overwrite a real user column with the same name.

## Validation
- `.test-venv/bin/python -m pytest tests/unit/various/test_llm_utils.py -q` -> `2 passed in 13.09s`
- `/tmp/mindsdb-ruff-venv/bin/ruff check mindsdb/integrations/libs/llm/utils.py tests/unit/various/test_llm_utils.py` -> `All checks passed!`

### DIFF
--- a/mindsdb/integrations/libs/llm/utils.py
+++ b/mindsdb/integrations/libs/llm/utils.py
@@ -62,7 +62,7 @@ def get_completed_prompts(base_template: str, df: pd.DataFrame, strict=True) -> 
         if strict:
             raise AssertionError("No placeholders found in the prompt, please provide a valid prompt template.")
         prompts = [base_template] * len(df)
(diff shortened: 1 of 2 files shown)


pragnyanramtha:

pragnyanramtha:
recheck

PR #12493, opened 2026-09-02
docs: promote Tasks API branding, drop Use Cases, add hosted base URL (ENG-2207)

torrmal:
## Summary

Rebrands the Cowork docs API reference as the **Agents API**, promoted around a Task-centric model, with a fully reworked ERD, a Local/Hosted base-URL toggle, and complete curl+response coverage for every endpoint.

### Branding
- Nav bar: `MindsHub Cowork` → **MindsHub Agents**; `Setup` → `Cowork-App`; `API` → **Agents API**; `Hosted API` → **Inference API**.
- API page: title/H1/meta tags → **Agents API**; new lede ("Launch agents to perform all kinds of tasks programmatically...").
- Homepage: hero → "...with MindsHub Agents"; "Get Started" card → **Agent Apps**, reworded description.
- Top banner rebranded to signpost the separate MindsHub **Inference API** (Model-Router), replacing the old "these pages document Cowork..." framing.
- Removed the Use Cases page and all its nav/homepage links.

### Tasks API content
- `/conversations` → `/tasks` throughout (paths, headings, prose); "Tasks and Conversations are the same concept" called out; literal field names/enum values left as-is since that's what the server actually returns.
- **Audited the Tasks section against the real `cowork-server` code** and fixed real drift: removed fictional `goal`/`status` fields, corrected the create/list/update body & query params, fixed the non-nullable `project_id` claim, documented the real response shape, and added two previously-undocumented endpoints (`POST /tasks/{id}/move`, `DELETE /tasks/{id}/turns/{turn_index}`).
- Streaming section ("Streaming a task") documents the real `/responses` lifecycle including the previously-undocumented `GET /responses/tail` (resume via `from_seq`) and `POST /responses/cancel`.

### ERD diagram
- Task now sits beside Project (not below it) as the diagram's main glowing block; Project uses a plain border.
- Schedule box removed; Artifacts and Scratchpad leveled; Task→Artifacts strengthened to a solid "owns" line.
- Connectors/Memories/Skills/Files grouped in a dashed "Agent Task Resources" box, pluralized, with reworked subtext (apps and data / self learnings storage / apps, presentations, documents / agent task outputs).
- Removed the legend and all now-dead relationship lines/markers.

### Base URL toggle
- Local/Hosted toggle, sticky under the nav, defaults to **MindsHub Cloud** (`https://cowork.mindshub.ai/api/v1`) with **Your Server** as the alternate.
- Every endpoint badge and curl example (including two-invocation blocks) shows the full URL and switches live; a **Get API Key** button appears next to the toggle when Hosted is selected, linking to `console.mindshub.ai/apiKeys`.
- Curl examples include a conditional `Authorization: Bearer` header, correctly ordered after the method+URL line and hidden when Your Server is selected.

### Page layout & completeness
- Each endpoint with real params and an example now splits into two CSS-grid columns (params/explanation left, curl+response right) — no JS, just layout.
- Params tables: type and required/optional now stack on two lines per cell instead of two cramped columns.
- Copy buttons on every code block (delegated click handler).
- **All 52 endpoints now show both a curl example and a response card** — added whatever was missing, reusing each section's documented object shape for consistency; verified via a scripted inventory pass, not by eyeballing it.

Linear: https://linear.app/mindsdb/issue/ENG-2207

## Test plan
- [x] Tag-balance verified (`<div>`/`</div>`, `<span>`/`</span>`, `<table>`/`</table>`, `<p>`/`</p>`, `<pre>`/`</pre>`) after every structural change.
- [x] `node --check` on the extracted page JS after every script change.
- [x] Scripted inventory confirms 0/52 endpoints missing a curl or response card.
- [x] Opened the page locally throughout to sanity-check rendering.
- [ ] Manual click-through of the Local/Hosted toggle, the copy buttons, and the two-column layout on a live preview (recommended before merge) — this PR was authored without a browser-screenshot tool, so visual QA has not been done end-to-end.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


Dav1dF (changes requested):
Requesting changes. Two blockers.

What holds up: CI is green, both test-plan claims verify (tag balance under a real parser, `node --check` on all three scripts), and the `/tasks` and `/responses` sections are accurate against the server.

## 1. Response cards: 20 of the 26 I checked misdescribe the server

The base file had no per-endpoint response cards, so this PR added all 52. Five of the six section shapes it reused were already wrong, and each was copied into several cards.

| Section | Card says | Server |
|---|---|---|
| Schedules | `"cron": "0 9 * * 1"`, `"status": "active"` | `cadence` enum (`once/hourly/daily/weekdays/weekly`), `enabled: bool`. No cron field exists; the section blurb is wrong too |
| Files | `{name, mime_type, size, project_id, conversation_id}` | OpenAI Files shape: `{id, object, bytes, created_at (int epoch), filename, purpose}` |
| DELETE x4 (projects, schedules, files, skills) | `{"ok": true}` | `204 NO_CONTENT`, empty body |
| Skills | `"instructions"`, `"tools": [...]` | serializes as `declarative`; no `tools` field on the model |
| Projects | `description`, `connector_ids`, `skill_ids`, `updated_at` | none exist; it is `modified_at`. Missing required `path`, `is_active`, `attribution`, `capabilities` |
| Connector specs | `"name"`, `"auth_type"`, `"fields[].key"` | `label`; no `auth_type` on any connector model; `ConnectorField` key is `name` |
| Connections | `"id"`, `"spec_id"`, `"status": "connected"` | keyed on `(engine, name)`; `engine`; `status` is null when healthy, `"needs_reconnect"` when not |
| Artifacts | `type: app/report/doc/dashboard`, `created_at`, `updated_at` | `html-app/document/dataset/image/mixed/fullstack-*`; no creation timestamp; `updated` is a human string. `type` is also a query filter, so the filter values fail too |
| Search | `total`, `excerpt`, `score: 0.93`, `type: conversation` | `{results: [...]}` only; `subtitle`; `_score` is an unbounded int; emits `task/project/artifact/schedule/pin` |

Several of these sections serialize camelCase (`nextRunAt`, `createdAt`, `projectId`). The cards show snake_case throughout.

Separately, 18 of the 52 documented endpoints do not exist — detail in my earlier comment. Nine `/scratchpads` endpoints now carry worked curl examples and response cards for routes that have never existed.

`GET /artifacts/preview`, `POST /artifacts/open` and `/artifacts/serve/...` all carry `require_local_tenancy`, so they cannot succeed against the hosted base URL the page defaults to.

## 2. Deleting Use Cases drops an indexed URL with no redirect

`/cowork/use-cases.html` returns 200 today with `<meta name="robots" content="index, follow">` and no canonical. `/use-cases.html` 301s to it. Unmatched paths 404 with no fallback, so both go dead on merge.

Nothing in this repo still links to it and it is in no sitemap, so the fix is not here: it is a `LEGACY_COWORK_PAGES` entry in `mindsdb/mindshub_inference` `docs/router/router.js`, plus a manual `wrangler deploy`. `docs/router/verify.mjs:64` also asserts that URL returns 200 and will begin failing.

## Suggested path

Land the `/tasks` section, the `Inference API` rename and the base-URL toggle. Delete the scratchpads and memories sections rather than repathing them. For the rest, generating the reference from the FastAPI schema would cost less than another manual pass and would stop this recurring; I can raise that separately.

